PR #1, opened 2022-08-24
Develop

BudnikOleksii:
- [DEMO LINK](https://BudnikOleksii.github.io/react_todo-app-with-api/)


mgrinko (changes requested):
I tried to toggle `JS` option but for some reason the loader appeared at the bottom

Also the input became focused which is wrong I think

https://prnt.sc/Hxs7-FcfMtFX

BudnikOleksii:
> I tried to toggle `JS` option but for some reason the loader appeared at the bottom
> 
> Also the input became focused which is wrong I think
> 
> https://prnt.sc/Hxs7-FcfMtFX

When input changes status active/completed I change it on the server and update it accordingly

BudnikOleksii:
> I tried to toggle `JS` option but for some reason the loader appeared at the bottom
> 
> https://prnt.sc/Hxs7-FcfMtFX
> 
> * Also the input became focused which is wrong I think
> * Filters should not jump when Clear Completed button disappears (see the demo)

Thank you, fixed

mgrinko:
- Loaders should be disabled in case of error
- In case of update error title should not be changed

BudnikOleksii:
> * Loaders should be disabled in case of error
> * In case of update error title should not be changed

My bad, thank you. Also, fixed a bug when the title is empty and a delete error occurred

src/api/todos.ts
```diff
@@ -1,8 +1,21 @@
...
 export const getTodos = (userId: number) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const addTodo = (todo: NewTodo): Promise<Todo> => {
+  return client.post<Todo>('/todos', todo);
+};
+
+export const removeTodoByTodoId = (todoId: number) => {
```
VETALEK, line 12:
```suggestion
export const removeTodoById = (todoId: number) => {
```

src/api/todos.ts
```diff
@@ -1,8 +1,21 @@
...
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const updateTodoByTodoId = (
```
VETALEK, line 16:
```suggestion
export const updateTodoById = (
```

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,88 @@
...
+    todos.filter(todo => todo.completed)
+  ), [todos]);
+
+  const handleRemoveCompletedTodos = () => {
+    todos.forEach(todo => {
+      if (todo.completed) {
+        handleRemoveTodo(todo.id);
+      }
+    });
+  };
```
VETALEK, line 31: let's wrap it with `useCallback`

VETALEK: let's do it with `Promise.all` (i believe removing todo is async function)

image

VETALEK (changes requested):
GJ 🔥 
commented

src/components/TodoFooter/TodoFooter.tsx
BudnikOleksii: > let's do it with `Promise.all` (i believe removing todo is async function)
> 
> image

I had done with Promise.all, but then found that it is possible to handle with other methods😁 I Refactored one more time.
Also, I think it is unnecessary to wrap with useCallback if the function I use only in the same component where it is declared, but ok.

VETALEK (approved):
🚀🚀🚀

PR #4, opened 2022-08-26
Develop

fledith:
[DEMO LINK](https://fledith.github.io/react_todo-app-with-api/)


oleksandr-kotliarov (changes requested):
image

---------------------------------------------------------------------------------------------------

image

---------------------------------------------------------------------------------------------------

- Hide everything except the NewTodoField if there are no todos yet;
- Don't hide everything if todos are just filtered out.
image

---------------------------------------------------------------------------------------------------

- if new title is the same as the old one just cancel editing;
- if after editing title is empty todo should be deleted (check example)

------------------------------------------------------------------------------------


When i try to edit again i see the same text that was before error
1. image
2.  image
3.  image

---------------------------------------------------------------------------------------------------

- And i dont know if its important but when i start to click fast enter after todo add, i receive a lot of todos, you can just disable form

src/App.tsx
```diff
@@ -1,223 +1,313 @@
...
+
+    todos.forEach(todo => {
+      if (todo.completed) {
+        changedTodoId.push(todo.id);
+      }
+    });
+
+    setChangedTodosId(changedTodoId);
+
+    todos.forEach(todo => todo.completed && removeTodo(todo.id));
```
VETALEK, line 185: let's try `Promise.all` here
image

src/App.tsx
```diff
@@ -1,223 +1,313 @@
...
+
+    updateTodoById(todoId, data)
+      .then(() => setShouldUpdate(true))
+      .catch(() => onError('Unable to update a todo'))
+      .finally(() => {
+        setIsLoading(false);
+      });
+  };
+
+  const toggleAllTodo = () => {
```
VETALEK, line 153: let's wrap your methods with `useCallback`

VETALEK (changes requested):
GJ 👍 
commented

P.S.
when I click `ENTER` with empty name field it shows error only for a brief moment, should be 3 seconds

P.P.S.
i can add a todos with the space as a title

src/api/todos.ts
```diff
@@ -5,4 +5,22 @@ export const getTodos = (userId: number) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const createTodo = (
+  title: string,
+  userId: number,
+  completed: boolean,
```
solaryasha, line 11: It's better to pass object, so you will never make mistake with wrong arguments order

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+interface Props {
+  todo: Todo;
+  selectedTodoId: number | null;
+  onDeleteTodo: (todoId: number) => void;
+  onUpdateTodo: (todoId: number, data: {}) => void;
+  isLoading: boolean;
+  changedTodosId: number[];
+}
+
+export const TodoItem = (props: Props) => {
```
solaryasha, line 17: use `React.memo`

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface Props {
+  todos: Todo[];
+  selectedTodoId: number | null;
+  onDeleteTodo: (todoId: number) => void;
+  onUpdateTodo: (todoId: number, data: {}) => void;
+  isLoading: boolean;
+  changedTodosId: number[];
+}
+
+export const TodoList = (props: Props) => {
```
solaryasha, line 13: use React.memo

src/App.tsx
```diff
@@ -1,223 +1,356 @@
...
+  const [selectedTodoId, setSelectedTodoId] = useState<number | null>(null);
+  const [shouldUpdate, setShouldUpdate] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+  const [changedTodosId, setChangedTodosId] = useState<number[]>([]);
+  const [filteredBy, setFilteredBy] = useState<string | null>(null);
+
+  const onError = useCallback((errorTitle: string) => {
+    setErrorMessage(errorTitle);
+
+    setTimeout(() => setErrorMessage(''), 3000);
```
solaryasha, line 30: You should clear previous timer before starting new one

solaryasha (changes requested):

src/api/todos.ts
fledith: mb, but i dont wanna create new interface for object without id.

PR #7, opened 2022-08-28
Develop

silchyn:
[DEMO](https://silchyn.github.io/react_todo-app-with-api/)


fledith (approved):
I like your variant with adding field loading to the Todo object!

PR #9, opened 2022-08-28
Solution

Sergii2705:
- [DEMO LINK](https://Sergii2705.github.io/react_todo-app-with-api/)


oleksandr-kotliarov (approved):
good

src/App.tsx
```diff
@@ -1,223 +1,216 @@
...
+    }
+  }, [todos]);
+
+  const clearCompleted = useCallback(() => {
+    todos.forEach(({ completed, id }) => {
+      if (completed) {
+        handleRemoveTodo(id);
+      }
+    });
+  }, [todos]);
```
VETALEK, line 144: let's use `Promise.all` here, since todos should be removed simultaneously

image

VETALEK (changes requested):
GJ 👍 
commented

src/App.tsx
Sergii2705: My deletions work like several separate deletions and run at the same time, their deletion ends at different times only. I don't agree with you about Promise.all, in this case it's wrong to use it.

Sergii2705: I can prove why

VETALEK: I think we need only compute requests simultaneously, but not end deletion at the same time

VETALEK: I only doubt about `Promise.all`
Now i think `Promise.allSettled` is more appropriate here

VETALEK: >in this case it's wrong to use it. I can prove why

but I'll be glad to hear why it is wrong and will use it later if so

VETALEK (approved):
GJ 🚀 🚀 🚀

PR #75, opened 2022-10-03
Draft

VolodymyrChuyko:
[DEMO LINK](https://VolodymyrChuyko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,0 +1,150 @@
+import classNames from 'classnames';
+import { useEffect, useRef, useState } from 'react';
+import {
+  CSSTransition,
+  TransitionGroup,
+} from 'react-transition-group';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[];
+  isProcessing: number[];
+  onStatusChange: (todoId: number) => void;
+  onTitleChange: (todoId: number, newTitle: string) => void;
+  onDeleteTodo: (todoId: number) => void;
+};
+
+export const TodoList: React.FC<Props> = (props) => {
+  const {
+    todos,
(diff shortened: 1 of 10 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,154 @@
...
+    <TransitionGroup>
+      {todos.map(todo => {
+        const {
+          id,
+          title,
+          completed,
+        } = todo;
+
+        return (
+          <CSSTransition
```
NataliyaAn, line 75: :fire: TransitionGroup- great implementation

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,154 @@
...
+    }
+  };
+
+  const handleCancelChanges = (
+    event: React.KeyboardEvent<HTMLInputElement>,
+  ) => {
+    if (event.key === 'Escape') {
+      setSelectedTodo(null);
+    }
+  };
```
NataliyaAn, line 53: You could add a check - which key was pressed to save. Now the edited todo is saved if you pressed alt or clicked mouse on another field.

NataliyaAn (approved):
:fire: Great job

src/components/TodoList/TodoList.tsx
VolodymyrChuyko: But this is exactly what we need: save changes when the field loses focus (onBlur)

src/components/TodoList/TodoList.tsx
VolodymyrChuyko: Thanks, I got fun messing around with it =)

VETALEK (approved):
GJ 👍

PR #77, opened 2022-10-04
Develop

Githi54:
[DEMO LINK](https://githi54.github.io/react_todo-app-with-api/)


oksana-logos-frontend (changes requested):
Good work, but in my opinion it's not finish yet.You should add spiner on toggle, clear completed, and when you delete completed todos.Also you can try to add transition to your App.

PR #95, opened 2022-10-06
Develop

ibondrk:
[DEMO LINK](https://ibondrk.github.io/react_todo-app-with-api/)

This is not finall version. Need to add loader


Andrii-Stiebalov (approved):
Good job! Some mistakes are the same as in previous solutions. I don't want to repeat that))) 👏😎

VETALEK (changes requested):
- You don't need so huge padding on edit
image
- When I click `clear completed` all the todos should start loading till the server responds. But they disappear immediately
- When I rename todo with an empty name - the empty todo appears at the bottom and then disappear

PR #108, opened 2022-10-06
solution

lizapolokhina:
[DEMO LINK](https://lizapolokhina.github.io/react_todo-app-with-api/)


aleksandra-prognimak (changes requested):
Link does not work😢

PR #133, opened 2022-11-01
solution

maryna-mazur:
[DEMO LINK](https://maryna-mazur.github.io/react_todo-app-with-api/)


vlad-demchuk (changes requested):
it's not working, fix this🙂
image

maryna-mazur:
Try again. It's working)

https://user-images.githubusercontent.com/107570670/199487233-b9b60968-3788-4028-9b33-4b2cd9407fb3.mp4

src/App.tsx
```diff
@@ -19,205 +198,53 @@ export const App: React.FC = () => {
...
+            completedTodosIds={completedTodosIds}
+            handlerClearCompletedButton={handlerClearCompletedButton}
+          />
+        )}
       </div>
+
+      <TodoErrorNotification
+        error={error}
+        setError={setError}
+      />
```
vlad-demchuk, line 247: you could use conditional rendering instead of hiding it using styles

vlad-demchuk (changes requested):
there is always should be active todos
image

src/App.tsx
```diff
@@ -1,16 +1,195 @@
...
+import { User } from './types/User';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
-  const newTodoField = useRef<HTMLInputElement>(null);
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodoId, setSelectedTodoId] = useState<number | null>(null);
+  const [error, setError] = useState<string | null>(null);
```
vlad-demchuk, line 26: use `enum` for errors

src/App.tsx
```diff
@@ -1,16 +1,195 @@
...
+  const [sortType, setSortType] = useState(SortType.ALL);
+  const [isAdding, setIsAdding] = useState(false);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [title, setTitle] = useState('');
+  const [isLoader, setIsLoader] = useState(false);
+  const [isLoaderCompletedTodo, setIsLoaderCompletedTodo] = useState(false);
+  const [isLoaderUnCompletedTodo, setIsLoaderUnCompletedTodo] = useState(false);
+  const [isDoubleClick, setIsDoubleClick] = useState(false);
```
vlad-demchuk, line 34: why do you need this `isDoubleClick `? and you use it many times. it's not necessary to check always. make it simpler. you have `onDoubleClick` react event, so make your logic simpler.

src/App.tsx
```diff
@@ -1,16 +1,195 @@
...
+    }
+
+    return filteredTodos;
+  };
+
+  const handlerSubmitForm = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (!newTodoTitle) {
+      setError('Title can\'t be empty');
+    } else {
```
vlad-demchuk, line 65:
```suggestion
    if (!newTodoTitle) {
      setError('Title can\'t be empty');
      
      return
     }
```
it's better for readability

vlad-demchuk (changes requested):
Looks good! Just try to make your logic simpler

PR #134, opened 2022-11-03
Solution

ElinaBahirova:
https://elinabahirova.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,144 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import {
+  useContext, useEffect, useMemo, useRef, useState,
+} from 'react';
+import classNames from 'classnames';
 import { AuthContext } from './components/Auth/AuthContext';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/Todo';
+import {
+  getTodos, addTodo, deleteTodo, updateTodoStatus, updateTodoTitle,
+} from './api/todos';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,18 +1,144 @@
...
-import React, { useContext, useEffect, useRef } from 'react';
+import {
+  useContext, useEffect, useMemo, useRef, useState,
+} from 'react';
+import classNames from 'classnames';
 import { AuthContext } from './components/Auth/AuthContext';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/Todo';
+import {
+  getTodos, addTodo, deleteTodo, updateTodoStatus, updateTodoTitle,
```
loralevitska, line 10: this is not a mistake, you can also write this way, just a recommendation

in large projects it is convenient when each variable is on a new line, it is convenient to delete or copy through shortcuts, and also in the git there are not so many changes (when editing a method only one line)
```suggestion
  getTodos,
  addTodo, 
  deleteTodo, 
  updateTodoStatus, 
  updateTodoTitle,
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+    renameTodo(id, newTodoTitle).finally(() => setIsChanging(false));
+  };
+
+  const saveTitleChanges = () => {
+    setTitleIsChanging(false);
+    if (title === newTitle) {
+      return;
+    }
+
+    if (newTitle === '') {
```
loralevitska, line 52:
```suggestion
    if (!newTitle.trim()) {
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim

loralevitska (changes requested):
almost done! need little improvements

PR #135, opened 2022-11-04
solution

AntonVaida:
[DEMO LINK](https://AntonVaida.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,136 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import classNames from 'classnames';
 import { AuthContext } from './components/Auth/AuthContext';
+import {
+  getTodos,
+  createTodos,
+  deleteTodos,
+  UpdateTodo,
+  // EditTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+// eslint-disable-next-line import/extensions
+import { ToDo } from './components/Auth/Todo';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todoList, setTodoList] = useState<Todo[]>([]);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [isAdding, setIsAdding] = useState<Todo | null>(null);
+  const [errorAdd, setErrorAdd] = useState(false);
+  const [errorRemove, setErrorRemove] = useState(false);
+  const [errorUpdate, setErrorUpdate] = useState(false);
+  const [errorEmptyTitile, setErrorEmptyTitile] = useState(false);
+  const [typeFilter, setTypeFilter] = useState('All');
+  const [togglerSelectAll, setTogglerSelectAll] = useState(true);
+  const [hidden, setHidden] = useState(true);
+  const [edditing, setEdditing] = useState<Todo | null>(null);
+  const editTitle = useRef<HTMLInputElement>(null);
+
+  const selectTypeFiltred = (type:string) => {
+    setTypeFilter(type);
+  };
+
+  const foundTodoList = (u:User) => {
+    getTodos(u.id).then(response => {
+      setTodoList(response);
+      setIsAdding(null);
+      setEdditing(null);
+    });
+  };
+
+  const createTodo = async (event:any) => {
+    event.preventDefault();
+    if (!newTodoTitle) {
+      setErrorEmptyTitile(true);
+      setHidden(false);
+
+      return;
+    }
+
+    if (user) {
+      const newTodo = {
+        userId: user.id,
+        title: newTodoTitle,
+        completed: false,
+      };
+
+      setIsAdding(newTodo);
+
+      try {
+        await createTodos(newTodo);
+      } catch {
+        setErrorAdd(true);
+        setHidden(false);
+        setIsAdding(null);
+      }
+
+      foundTodoList(user);
+      setNewTodoTitle('');
+    }
+  };
+
+  const selectComplited = async (todo:Todo) => {
+    try {
+      if (todo.completed) {
+        await UpdateTodo(todo, false);
+      } else {
+        await UpdateTodo(todo, true);
+      }
+    } catch {
+      setErrorUpdate(true);
+      setHidden(false);
+    }
+
+    if (user) {
+      foundTodoList(user);
+    }
+  };
+
+  const selectAllTodos = () => {
+    todoList.map(async (todo) => {
+      if (!todo.completed && togglerSelectAll) {
+        selectComplited(todo);
+      } else if (todo.completed && !togglerSelectAll) {
+        selectComplited(todo);
+      }
+    });
+
+    setTogglerSelectAll(prefer => !prefer);
+  };
+
+  const clearCompletedTodo = () => {
+    todoList.map(async (todo) => {
+      if (todo.completed) {
+        deleteTodos(todo);
+        try {
+          await deleteTodos(todo);
+        } catch {
+          setErrorRemove(true);
+          setHidden(false);
+        }
+
+        if (user) {
+          foundTodoList(user);
+        }
+      }
+    });
+  };
+
+  useEffect(() => {
+    if (user) {
+      foundTodoList(user);
+    }
+  }, [user]);
 
   useEffect(() => {
     // focus the element with `ref={newTodoField}`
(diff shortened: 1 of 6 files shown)


src/api/todos.ts
```diff
@@ -1,8 +1,37 @@
...
+  return client.delete(`/todos/${todo.id}`);
+};
+
+export const UpdateTodo = async (todo:Todo, completed:boolean) => {
+  return client.patch(`/todos/${todo.id}`, { completed });
+};
+
+export const EditTodo = async (todo:Todo, title:string) => {
+  return client.patch(`/todos/${todo.id}`, { title });
+};
```
Andr1yK, line 36: as for me, u can create once update method something looks like this:
```
export const updateTodo = (
  todoId: number,
  values: Partial<Omit<Todo, 'id'>>,
) => {
  return client.patch<Todo>(`/todos/${todoId}`, values);
};

File: src/components/Auth/Todo.tsx [hunk 2]
```diff
@@ -0,0 +1,207 @@
...
+      setErrorRemove(true);
+      setHidden(false);
+    }
+
+    if (user) {
+      foundTodoList(user);
+    }
+  };
+
+  const editingTodo = async (toDo:Todo) => {
```
Andr1yK, line 58: use useCallback :)

src/components/Auth/Todo.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const onKeyPressHandler = (event: { key: string }) => {
+    if (event.key === 'Escape') {
+      setToggleDoubleClick(null);
+    }
+  };
+
+  const activeTodoLoader = (isLoading && isLoading.id === todo.id)
+  || (!todo.id)
+  || (loadingAllTodos)
+  || (clearLoader && todo.completed);
```
Andr1yK, line 132:
```suggestion
  const activeTodoLoader = (isLoading && isLoading.id === todo.id)
	  || (!todo.id)
	  || (loadingAllTodos)
	  || (clearLoader && todo.completed);
```

src/components/Auth/Todo.tsx
```diff
@@ -0,0 +1,207 @@
...
+        <div className="loader" />
+      </div>
+
+      <button
+        type="button"
+        className="todo__remove"
+        data-cy="TodoDeleteButton"
+        onClick={() => removeTodo(todo)}
+      >
+        { toggleDoubleClick !== todo && (
```
Andr1yK, line 201:
```suggestion
        {toggleDoubleClick !== todo && (
```

src/App.tsx
```diff
@@ -1,223 +1,294 @@
...
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todoList, setTodoList] = useState<Todo[]>([]);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [errorAdd, setErrorAdd] = useState(false);
+  const [errorRemove, setErrorRemove] = useState(false);
+  const [errorUpdate, setErrorUpdate] = useState(false);
+  const [errorEmptyTitile, setErrorEmptyTitile] = useState(false);
```
Anna-Tkachenko, line 31: let's create one variable in state for setting errors, and will use enum with available options to define which error was occured? in this way for handling one more error you'll need only to add and process one more option in enum, instead of creating new variable in state and add code for it's processing

src/App.tsx
```diff
@@ -1,223 +1,294 @@
...
+  const [typeFilter, setTypeFilter] = useState('All');
+  const [hidden, setHidden] = useState(true);
```
Anna-Tkachenko, line 33: it's not clear, what is hidden when this variable has true value

src/App.tsx
```diff
@@ -1,223 +1,294 @@
...
+  const [isAdding, setIsAdding] = useState<Todo | null>(null);
+  const [clearLoader, setClearLoader] = useState(false);
+  const [loadingAllTodos, setLoadingAllTodos] = useState(false);
+  const [isLoading, setIsLoading] = useState<Todo | null>(null);
```
Anna-Tkachenko, line 37: isLoading variable should store some boolean value, not Todo

src/App.tsx
```diff
@@ -1,223 +1,294 @@
...
+
+  const handlerRemoveError = useCallback((value: boolean) => {
+    setErrorRemove(value);
+  }, []);
+
+  const handlerNewTitle = useCallback((value: string) => {
+    setNewTodoTitle(value);
+  }, []);
+
+  const foundTodoList = useCallback((u:User) => {
```
Anna-Tkachenko, line 65: probably you don't want to found smth here, just want to get updated data from server?)
let's think about naming for this function

src/App.tsx
```diff
@@ -1,223 +1,294 @@
...
+  );
+
+  const selectComplited = useCallback(async (todo:Todo) => {
+    try {
+      setIsLoading(todo);
+      if (todo.completed) {
+        await UpdateTodo(todo, false);
+      } else {
+        await UpdateTodo(todo, true);
+      }
```
Anna-Tkachenko, line 118: will be better to create separate variable with needed value of completed property, and call function for updating todo only once, to avoid code repeating

src/App.tsx
```diff
@@ -1,223 +1,294 @@
...
+    }
+
+    if (user) {
+      foundTodoList(user);
+    }
+  }, [todoList]);
+
+  const selectAllTodos = useCallback(() => {
+    todoList.map(async (todo) => {
+      setLoadingAllTodos(true);
```
Anna-Tkachenko, line 132: should we set this value to the true on each callback call?

src/App.tsx
```diff
@@ -1,223 +1,294 @@
...
+  }, [user]);
 
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
   }, []);
 
+  const listFiltring = (list:Todo[]) => (
```
Anna-Tkachenko, line 174: check the function name, it should contain an action

src/api/todos.ts
```diff
@@ -1,8 +1,36 @@
...
+    completed,
+  });
 };
 
-// Add more methods here
+export const deleteTodos = async (todo: Todo) => {
+  return client.delete(`/todos/${todo.id}`);
+};
+
+export const UpdateTodo = async (todo:Todo, value: boolean | string) => {
```
Anna-Tkachenko, line 30:
```suggestion
export const updateTodo = async (todo:Todo, value: boolean | string) => {
```

Anna-Tkachenko: will be better to use separate arguments for completed and title values instead of using one general argument with any possible type of todo properties

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,223 +1,286 @@
...
+    setLoadingAllTodos(true);
+    todoList.map(async (todo) => {
+      if (!checkedAllCompletedTodo(todoList)) {
+        if (!todo.completed) {
+          selectComplited(todo);
+        }
+      } else {
+        selectComplited(todo);
+      }
+    });
```
Anna-Tkachenko, line 146: use forEach method instead of the map one, if you don't need the result of map method

src/App.tsx
```diff
@@ -1,223 +1,286 @@
...
+  }, [todoList]);
+
+  const clearCompletedTodo = useCallback(() => {
+    todoList.map(async (todo) => {
```
Anna-Tkachenko, line 150: same here

src/api/todos.ts
```diff
@@ -1,8 +1,35 @@
 import { Todo } from '../types/Todo';
 import { client } from '../utils/fetchClient';
 
-export const getTodos = (userId: number) => {
-  return client.get<Todo[]>(`/todos?userId=${userId}`);
+export const getTodos = async (userId: number) => {
+  const todo = await client.get<Todo[]>(`/todos?userId=${userId}`);
```
Anna-Tkachenko, line 5: it's not one todo, check variable name

src/api/todos.ts
```diff
@@ -1,8 +1,35 @@
...
+) => {
+  return client.post<Todo[]>('/todos', {
+    userId,
+    title,
+    completed,
+  });
 };
 
-// Add more methods here
+export const deleteTodos = async (todo: Todo) => {
```
Anna-Tkachenko, line 26: will be better to accept just id here

src/api/todos.ts
```diff
@@ -1,8 +1,35 @@
...
+  return client.delete(`/todos/${todo.id}`);
+};
+
+export const updateTodo = async (
+  todo:Todo,
```
Anna-Tkachenko, line 31: same here, will be better to accept just todo id

src/components/Footer.tsx
```diff
@@ -0,0 +1,83 @@
...
+        </a>
+        <a
+          data-cy="FilterLinkCompleted"
+          href="#/completed"
+          className={
+            classNames('filter__link', {
+              'filter__link selected': typeFilter === 'Completed',
+            })
+          }
+          onClick={() => selectTypeFiltred('Completed')}
```
Anna-Tkachenko, line 63: will be better to use enum for type filtered

src/components/Todo.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const user = useContext(AuthContext);
+  const [editTodoTitile, setEditTodoTitle] = useState('');
+  const [toggleDoubleClick, setToggleDoubleClick] = useState<Todo | null>(null);
+  const editTitle = useRef<HTMLInputElement>(null);
+
+  const removeTodo = async (toDo:Todo) => {
+    deleteTodos(toDo);
+    try {
+      onLoadTodo(toDo);
+      await deleteTodos(toDo);
```
Anna-Tkachenko, line 49: why did you call deleteTodos function 2 times one by one?

Anna-Tkachenko (approved):

PR #139, opened 2022-11-07
third part implemented

illia-sosnytskyi:
[DEMO LINK](https://illia-sosnytskyi.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,291 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useState,
+  useContext,
+  useEffect,
+  useRef,
+  useMemo,
+  useCallback,
+} from 'react';
+
 import { AuthContext } from './components/Auth/AuthContext';
+import { ErrorMessages } from './components/ErrorMessages';
+import { TodoContent } from './components/TodoContent/TodoContent';
+
+import { Todo } from './types/Todo';
+import { FilterStatus } from './types/FilterStatus';
+import { ErrorType } from './types/ErrorType';
+
+import {
+  getTodos,
+  createTodo,
+  removeTodo,
+  editCompleteTodoStatus,
+  changeTitle,
+} from './api/todos';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
 
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [proccessedTodoId, setProccessedTodoId] = useState<number[]>([]);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  const isEveryTodosComplited = todos.every(todo => todo.completed);
+
+  const [filterStatus, setFilterStatus]
+    = useState<FilterStatus>(FilterStatus.All);
+
+  const manageErrors = useCallback((errorType: ErrorType) => {
+    setErrorMessage(currentMessage => {
+      let newMessage = currentMessage;
+
+      switch (errorType) {
+        case ErrorType.Endpoint:
+          newMessage = 'Fetch error';
+          break;
+
+        case ErrorType.Title:
+          newMessage = 'Title can`t be empty';
+          break;
+
+        case ErrorType.Add:
+          newMessage = 'Unable to add a todo';
+          break;
+
+        case ErrorType.Delete:
+          newMessage = 'Unable to delete a todo';
+          break;
+
+        case ErrorType.Update:
+          newMessage = 'Unable to update a todo';
+          break;
+
+        case ErrorType.None:
+        default:
+          newMessage = '';
+      }
+
+      return newMessage;
+    });
+  }, []);
+
+  const showError = useCallback((errorType: ErrorType) => {
+    manageErrors(errorType);
+    setTimeout(() => {
+      manageErrors(ErrorType.None);
+    }, 3000);
+  }, []);
+
+  const onCompleteTodo = useCallback((todoId: number) => {
+    setTodos(currentTodos => (
+      currentTodos.map(todo => {
+        if (todo.id === todoId) {
+          return {
+            ...todo,
+            completed: !todo.completed,
+          };
+        }
+
+        return todo;
+      })
+    ));
+  }, [todos]);
+
+  const onChangeTodoTitle = useCallback((todoId: number, newTitle: string) => {
+    setTodos(currentTodos => (
+      currentTodos.map(todo => {
+        if (todo.id === todoId) {
+          return {
+            ...todo,
+            title: newTitle,
+          };
+        }
+
+        return todo;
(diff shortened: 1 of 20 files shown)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  const addFocusOnForm = () => {
+    setFocusOnForm(true);
+  };
+
+  const removeFocusFromForm = () => {
+    setFocusOnForm(false);
+  };
+
+  const handleForm = (event: React.FormEvent) => {
```
OleksandrGlulhovskiy, line 39: i think it's will be better with useCallback

OleksandrGlulhovskiy (commented):
GJ

src/components/TodoBody/TodoBody.tsx
```diff
@@ -0,0 +1,48 @@
...
+  todoTitle,
+  deleteTodo,
+  proccessedTodoId,
+  changeCompleteStatus,
+  changeTodoTitle,
+}) => (
+  <section className="todoapp__main" data-cy="TodoList">
+
+    {
+      visibleTodos.map(todo => (
```
Andr1yK, line 28:
```suggestion
    {visibleTodos.map(todo => (
```

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "@mate-academy/students-ts-config",
   "include": [
-    "src"
+    "src",
+    "public/src"
   ]
-}
+}
```
Andr1yK, line 7:
```suggestion
}

```

Andr1yK (changes requested):

src/App.tsx
```diff
@@ -1,223 +1,259 @@
...
+  updateTodo,
+} from './api/todos';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
 
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [proccessedTodoId, setProccessedTodoId] = useState<number[]>([]);
```
Anna-Tkachenko, line 30: if it's an array of ids, let's refactor the name: Ids instead of id

src/App.tsx
```diff
@@ -1,223 +1,259 @@
...
+  const [filterStatus, setFilterStatus] = useState(FilterStatus.All);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  const isEveryTodosComplited = useMemo(() => (
```
Anna-Tkachenko, line 36:
```suggestion
  const isEveryTodoComplited = useMemo(() => (
```

src/App.tsx
```diff
@@ -1,223 +1,259 @@
...
+    todos.every(todo => todo.completed)
+  ), [todos]);
+
+  const countOfTodos = useMemo(() => todos.length, [todos]);
```
Anna-Tkachenko, line 40: it's redundant to cover primitive value with useMemo
in case with objects or functions we need it, because for each rerendering we receive the new link for some object
but in case with primitive values, the value doesn't change (there is no any link changes), it's still the same primitive value, so you don't need to use useMemo for it

src/App.tsx
```diff
@@ -1,223 +1,259 @@
...
+        createNewTodo={createNewTodo}
+        isLoading={isLoading}
+        deleteTodo={deleteTodo}
+        proccessedTodoId={proccessedTodoId}
+        changeCompleteStatus={changeCompleteStatus}
+        deleteAllCompletedTodos={deleteAllCompletedTodos}
+        copleteAllTodos={copleteAllTodos}
+        isEveryTodosComplited={isEveryTodosComplited}
+        changeTodoTitle={changeTodoTitle}
+      />
```
Anna-Tkachenko, line 251: there is so many props for TodoContent component
can we optimize it? let's check this props, probably some of them can be moved and be calculated inside the TodoContent component and be passed as an arguments for the needed functions

src/api/todos.ts
```diff
@@ -5,4 +5,21 @@ export const getTodos = (userId: number) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const createTodo = (title: string, userId: number) => {
```
Anna-Tkachenko, line 8: add types for returned values for the api functions

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,223 +1,181 @@
...
+    try {
+      setProccessedTodoIds(currentIds => ([
+        ...currentIds,
+        todoId,
+      ]));
+
+      await updateTodo(todoId, { title: data });
+      await getTodosFromServer();
+
+      if (data === '') {
```
OleksiiBlezniuk, line 132:
```suggestion
      if (!data) {
```

src/components/TodoContent/TodoContent.tsx
```diff
@@ -0,0 +1,121 @@
...
+          default:
+            return todo;
+        }
+      });
+    };
+
+    const filteredTodos = useMemo(
+      () => filterTodos(filterStatus),
+      [filterStatus, todos],
+    );
```
OleksiiBlezniuk, line 84: Just filter array and save result to constant using `useMemo`, why to create function for this?

src/components/AddingTodoItem/AddingTodoItem.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+type Props = {
+  title: string
+};
+
+export const AddingTodoItem: React.FC<Props> = React.memo(({ title }) => {
```
OleksiiBlezniuk, line 7: But this component is almost same as TodoItem, why do you need separate constant for it? We can just reuse TodoItem component and pass needed data

src/components/TodoBody/TodoBody.tsx
```diff
@@ -0,0 +1,44 @@
...
+  filteredTodos: Todo[],
+  isLoading: boolean,
+  todoTitle: string,
+  deleteTodo: (todoId: number) => void;
+  proccessedTodoIds: number[],
+  changeCompleteStatus: (todoId: number, isComplited: boolean) => void;
+  changeTodoTitle: (todoId: number, newTitle: string) => void;
+};
+
+export const TodoBody: React.FC<Props> = React.memo(({
```
OleksiiBlezniuk, line 16:
```suggestion
export const TodoList: React.FC<Props> = React.memo(({
```

OleksiiBlezniuk: It's not clear what this component is used for from this name

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+    event.preventDefault();
+
+    if (title !== newTitle) {
+      changeTodoTitle(id, newTitle);
+    }
+
+    if (newTitle === '') {
```
OleksiiBlezniuk, line 46:
```suggestion
    if (!newTitle) {
```

OleksiiBlezniuk (changes requested):

PR #144, opened 2022-11-08
Develop

grafit-off:
[DEMO LINK](https://grafit-off.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -98,8 +98,8 @@ export const App: React.FC = () => {
           completed: false,
         });
 
-        setIsAdding(false);
         setTodos(currTodos => [...currTodos, addedTodo]);
+        setIsAdding(false);
       } catch (error) {
         generateError('Unable to add todo');
       }
@@ -194,6 +194,10 @@ export const App: React.FC = () => {
     }
   };
 
+  const isSelectionVisible = useMemo(() => (
+    filtredTodos.length > 0 || isAdding
+  ), [filtredTodos, isAdding]);
+
   useEffect(() => {
     setTimeout(() => setHasError(false), 3000);
   }, [hasError]);
@@ -214,6 +218,7 @@ export const App: React.FC = () => {
 
       <div className="todoapp__content">
         <TodoForm
+          isToggleVisible={todos.length !== 0}
           newTodoField={newTodoField}
           toggleAllTodosServerStatus={toggleAllTodosServerStatus}
           addTodoToServer={addTodoToServer}
@@ -223,26 +228,24 @@ export const App: React.FC = () => {
           isAllTodosCompleted={completedTodos.length === todos.length}
         />
 
-        {todos.length > 0 && (
-          <>
-            <TodoList
-              filtredTodos={filtredTodos}
-              isAdding={isAdding}
-              tempTodo={tempTodo}
-              todoIdsLoading={todoIdsLoading}
-              removeTodoFromServer={removeTodoFromServer}
-              toggleTodoServerStatus={toggleTodoServerStatus}
-              sendNewTodoTitleToServer={sendNewTodoTitleToServer}
-            />
+        <TodoList
+          filtredTodos={filtredTodos}
+          isAdding={isAdding}
+          tempTodo={tempTodo}
+          todoIdsLoading={todoIdsLoading}
+          removeTodoFromServer={removeTodoFromServer}
+          toggleTodoServerStatus={toggleTodoServerStatus}
+          sendNewTodoTitleToServer={sendNewTodoTitleToServer}
+        />
 
-            <TodosSelection
-              completedTodosLength={completedTodos.length}
-              todosLength={todos.length}
-              statusToFilter={statusToFilter}
-              setStatusToFilter={setStatusToFilter}
-              removeAllCompletedTodos={removeAllCompletedTodos}
-            />
-          </>
+        {isSelectionVisible && (
+          <TodosSelection
+            completedTodosLength={completedTodos.length}
+            todosLength={todos.length}
+            statusToFilter={statusToFilter}
+            setStatusToFilter={setStatusToFilter}
+            removeAllCompletedTodos={removeAllCompletedTodos}
+          />
         )}
       </div>
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,257 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
+
 import { AuthContext } from './components/Auth/AuthContext';
+import {
+  ErrorNotification,
+} from './components/ErrorNotification/ErrorNotification';
+import { TodosSelection } from './components/TodosSelection/TodosSelection';
+import { TodoForm } from './components/TodoForm/TodoForm';
+import { TodoList } from './components/TodoList/TodoList';
+
+import { Todo } from './types/Todo';
+import { TodosFilter } from './types/TodosFilter';
+import {
+  addTodo,
+  getTodos,
+  removeTodo,
+  sendNewTodoTitle,
+  toggleTodoStatus,
+} from './api/todos';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isAdding, setIsAdding] = useState(false);
+  const [hasError, setHasError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [todoIdsLoading, setTodoIdsLoading] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo>({
+    id: 0,
+    userId: 0,
+    title: '',
+    completed: false,
+  });
+  const [
+    statusToFilter,
+    setStatusToFilter,
+  ] = useState<TodosFilter>(TodosFilter.All);
+
+  const closeNotification = useCallback(() => setHasError(false), []);
+
+  const filtredTodos = useMemo(() => (
+    todos.filter(({ completed }) => {
+      switch (statusToFilter) {
+        case TodosFilter.Active:
+          return !completed;
+
+        case TodosFilter.Completed:
+          return completed;
+
+        default:
+          return true;
+      }
+    })
+  ), [todos, statusToFilter]);
+
+  const completedTodos = useMemo(() => (
+    todos.filter((todo) => todo.completed)
+  ), [todos]);
+
+  const generateError = (message: string) => {
+    setErrorMessage(message);
+    setHasError(true);
+  };
+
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,223 +1,260 @@
...
+        const addedTodo = await addTodo({
+          title: todoTitle,
+          userId: user.id,
+          completed: false,
+        });
+
+        setTodos(currTodos => [...currTodos, addedTodo]);
+        setIsAdding(false);
+      } catch (error) {
+        generateError('Unable to add todo');
```
Yevheniya-Deryka, line 104: Maybe it's better to set isAdding to false in case of error also

src/api/todos.ts
```diff
@@ -5,4 +5,18 @@ export const getTodos = (userId: number) => {
...
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const toggleTodoStatus = (todoId: number, completed: boolean) => {
+  return client.patch<Todo>(`/todos/${todoId}`, { completed });
+};
+
+export const sendNewTodoTitle = (todoId: number, newTitle: string) => {
+  return client.patch<Todo>(`/todos/${todoId}`, { title: newTitle });
+};
```
Yevheniya-Deryka, line 22: Maybe it's better to have 1 function to edit todo which will receive data object with fields that must be changed. But it's just my thought)

Yevheniya-Deryka (approved):
GJ! 🔥🔥🔥
Like your namings!) 
Left a couple of comments, just to add something) 
And maybe some functions can be wrapped into useCallback

src/App.tsx
```diff
@@ -1,223 +1,260 @@
...
+  const [isAdding, setIsAdding] = useState(false);
+  const [hasError, setHasError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [todoIdsLoading, setTodoIdsLoading] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo>({
+    id: 0,
+    userId: 0,
+    title: '',
+    completed: false,
+  });
```
OleksiiBlezniuk, line 41: Why do you need to save this temporary todo in state?

src/api/todos.ts
OleksiiBlezniuk, line 22: why do you need this two function? you can use one function for both cases, just send what data you need to update

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+  setErrorMessage,
+  toggleAllTodosServerStatus,
+  isAllTodosCompleted,
+  isToggleVisible,
+}) => {
+  const [todoText, setTodoText] = useState('');
+
+  const handleInput = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setTodoText(e.target.value);
+  };
```
OleksiiBlezniuk, line 29: useCallback

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
```
OleksiiBlezniuk, line 31: useCallback

src/App.tsx
```diff
@@ -1,223 +1,260 @@
...
+        const todosFromServer = await getTodos(user.id);
+
+        setTodos(todosFromServer);
+      }
+    } catch (error) {
+      generateError('Unable to show todos');
+    }
+  };
+
+  const addTodoToServer = async (todoTitle: string) => {
```
OleksiiBlezniuk, line 85: useCallback

src/App.tsx
```diff
@@ -1,223 +1,260 @@
...
+  ), [todos, statusToFilter]);
+
+  const completedTodos = useMemo(() => (
+    todos.filter((todo) => todo.completed)
+  ), [todos]);
+
+  const generateError = (message: string) => {
+    setErrorMessage(message);
+    setHasError(true);
+  };
```
OleksiiBlezniuk, line 71: useCallback

src/App.tsx
```diff
@@ -1,223 +1,260 @@
...
+
+  const getTodosFromServer = async () => {
```
OleksiiBlezniuk, line 73: useCallback

src/App.tsx
```diff
@@ -1,223 +1,260 @@
...
+      setTodoIdsLoading(currIds => (
+        currIds.filter((id) => id !== todoId)
+      ));
+    } catch (error) {
+      generateError(`Unable to change status of todo #${todoId}, todo not exist!`);
+      await getTodosFromServer();
+    }
+  };
+
+  const toggleAllTodosServerStatus = async () => {
```
OleksiiBlezniuk, line 160: useCallback

src/App.tsx
```diff
@@ -1,223 +1,260 @@
...
+
+      await Promise.all(todosToToggleStatus.map(({ id, completed }) => (
+        toggleTodoServerStatus(id, !completed)
+      )));
+    } catch (error) {
+      generateError('Unable to toggle all todos status!');
+    }
+  };
+
+  const sendNewTodoTitleToServer = async (todoId: number, newTitle: string) => {
```
OleksiiBlezniuk, line 174: useCallback

src/App.tsx
```diff
@@ -1,223 +1,260 @@
...
+            : todo
+        ))
+      ));
+
+      setTodoIdsLoading(currIds => (
+        currIds.filter((id) => id !== todoId)
+      ));
+    } catch (error) {
+      generateError(`Unable to change title of todo #${todoId}, todo not exist!`);
+      await getTodosFromServer();
```
OleksiiBlezniuk, line 193: why do you need to send this request again? just update it locally in state

src/App.tsx
```diff
@@ -1,223 +1,260 @@
...
+
+        setTodos(currTodos => [...currTodos, addedTodo]);
+        setIsAdding(false);
+      } catch (error) {
+        generateError('Unable to add todo');
+      }
+    }
+  };
+
+  const removeTodoFromServer = async (todoId: number) => {
```
OleksiiBlezniuk, line 109: useCallback

src/App.tsx
```diff
@@ -1,223 +1,260 @@
...
+
+      setTodoIdsLoading(currIds => (
+        currIds.filter((id) => id !== todoId)
+      ));
+    } catch (error) {
+      generateError(`Unable to remove todo ${todoId}`);
+    }
+  };
+
+  const removeAllCompletedTodos = async () => {
```
OleksiiBlezniuk, line 127: useCallback

src/App.tsx
```diff
@@ -1,223 +1,260 @@
...
+    try {
+      await Promise.all(completedTodos.map(({ id }) => (
+        removeTodoFromServer(id)
+      )));
+    } catch (error) {
+      generateError('Unable to remove all completed todo');
+    }
+  };
+
+  const toggleTodoServerStatus = async (todoId: number, status: boolean) => {
```
OleksiiBlezniuk, line 137: useCallback

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,153 @@
...
+    title,
+    completed,
+    id,
+  } = todo;
+
+  const [isFormVisible, setIsFormVisible] = useState(false);
+  const [newTitle, setNewTitle] = useState(title);
+  const titleInput = useRef<HTMLInputElement>(null);
+
+  const handleRemove = () => {
```
OleksiiBlezniuk, line 33: wrap your functions to usecallback

OleksiiBlezniuk (changes requested):

src/App.tsx
grafit-off: To be honest I need only tempTodoTitle in state

src/App.tsx
```diff
@@ -1,223 +1,268 @@
...
+
+      await patchTodo(todoId, { completed: status });
+
+      setTodos(currTodos => (
+        currTodos.map(todo => (
+          todo.id === todoId
+            ? ({ ...todo, completed: status })
+            : todo
+        ))
+      ));
```
OleksiiBlezniuk, line 150: in that case better to use updated todo from server or load new todos than setting this value

I mean in case of validation or any other errors on api you'll use data from server (which should be only source of truth)

src/App.tsx
```diff
@@ -1,223 +1,268 @@
...
+      setTodoIdsLoading(currIds => [...currIds, todoId]);
+
+      await patchTodo(todoId, { title: newTitle });
+
+      setTodos(currTodos => (
+        currTodos.map(todo => (
+          todo.id === todoId
+            ? ({ ...todo, title: newTitle })
+            : todo
+        ))
```
OleksiiBlezniuk, line 191: same here about updating

OleksiiBlezniuk (changes requested):

src/App.tsx
TheVovchik: you need handle only title change

src/App.tsx
grafit-off: Okay, will save response from fetch to todos

OleksiiBlezniuk: better make a request to server, that is better in case of this application, 'cause we need to have one source of truth, and it's an database on server, additional request is not critical in this case (I was wrong about it on Q&A)

grafit-off: I'm a little bit confused, so I can just make one extra request for getting todo after pathTodo?

OleksiiBlezniuk: yes, you can do same with other operations with todo (and better to do it)

Sorry for confusing, I was wrong when I tell about avoiding additional requests, in case of this application this solution is better, 'cause we won't face any performance issues if we'll make them, but we'll have an actual information from database

grafit-off: It's okay, will try to fix all tomorrow ⭐

src/App.tsx
```diff
@@ -1,223 +1,247 @@
...
+
+      await patchTodo(todoId, { completed: status });
+      await getTodosFromServer();
+
+      setTodoIdsLoading(currIds => (
+        currIds.filter((id) => id !== todoId)
+      ));
+    } catch (error) {
+      generateError(`Unable to change status of todo #${todoId}, todo not exist!`);
+      await getTodosFromServer();
```
OleksiiBlezniuk, line 147: do we need to load todos from server in case of error?

OleksiiBlezniuk: I mean, in this case nothing is changed and there is no need to update current todo list

src/App.tsx
```diff
@@ -1,223 +1,247 @@
...
+
+      await patchTodo(todoId, { title: newTitle });
+      await getTodosFromServer();
+
+      setTodoIdsLoading(currIds => (
+        currIds.filter((id) => id !== todoId)
+      ));
+    } catch (error) {
+      generateError(`Unable to change title of todo #${todoId}, todo not exist!`);
+      await getTodosFromServer();
```
OleksiiBlezniuk, line 180: same here

src/App.tsx
OleksiiBlezniuk: Let's do this only when actions with todos was successful

OleksiiBlezniuk (changes requested):
GJ, almost done, let's remove unnecessary requests to database

OleksiiBlezniuk (approved):
Well done!

PR #146, opened 2022-11-09
Ability to rename and change todo status was added

OleksandrGlulhovskiy:
[DEMO LINK](https://OleksandrGlulhovskiy.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,143 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useCallback,
+  useEffect,
+  useState,
+} from 'react';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+import { Error } from './types/Error';
 import { AuthContext } from './components/Auth/AuthContext';
+import {
+  getTodos,
+  addTodo,
+  deleteTodo,
+  changeTodo,
+} from './api/todos';
+
+import { TodoList } from './components/TodoList/TodoList';
+import { NewTodo } from './components/NewTodo/NewTodo';
+import { TodoError } from './components/TodoError/TodoError';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
-  const newTodoField = useRef<HTMLInputElement>(null);
+  const [userTodos, setUserTodos] = useState<Todo[]>([]);
+  const [hasTodos, setHasTodos] = useState(false);
+  const [hasError, setHasError] = useState<Error>({ status: false });
+  const [isAdding, setIsAdding] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [changingIds, setChangingIds] = useState<number[]>([0]);
+
+  const handleSetError = useCallback((error: Error) => {
+    setHasError(error);
+    setTimeout(() => {
+      setHasError({ status: false });
+    }, 3000);
+  }, []);
+
+  const handleLoadTodos = useCallback(async () => {
+    try {
+      setHasError({ status: false });
+
+      const todosFromServer = await getTodos((user as User).id);
+
+      setUserTodos(todosFromServer);
+    } catch {
+      handleSetError({ status: true });
+    }
+  }, []);
 
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
-    if (newTodoField.current) {
-      newTodoField.current.focus();
+    handleLoadTodos();
+  }, []);
+
+  useEffect(() => {
+    if (userTodos.length !== 0) {
+      setHasTodos(true);
+    } else {
+      setHasTodos(false);
+    }
+  }, [userTodos]);
+
+  const handleCloseError = useCallback(() => {
+    handleSetError({ status: false });
+  }, []);
+
+  const handleAddTodo = useCallback(async (title: string) => {
+    handleSetError({ status: false });
+
+    try {
+      const preparedData = {
+        title,
+        userId: (user as User).id,
+        completed: false,
+      };
+
+      setTempTodo({ ...preparedData, id: 0 });
+      setIsAdding(true);
+
+      await addTodo(preparedData);
+      await handleLoadTodos();
+
+      setIsAdding(false);
+    } catch {
+      handleSetError({ status: true, message: 'Unable to add a todo' });
+      setIsAdding(false);
+    }
+  }, []);
+
+  const handleDeleteTodo = useCallback(async (id: number) => {
+    handleSetError({ status: false });
+
+    try {
+      setChangingIds(currValue => [...currValue, id]);
+
+      await deleteTodo(id);
+      await handleLoadTodos();
+
+      setChangingIds(currValue => currValue.filter(currId => currId !== id));
+    } catch {
+      handleSetError({ status: true, message: 'Unable to delete a todo' });
+      setChangingIds(currValue => currValue.filter(currId => currId !== id));
+    }
+  }, []);
+
+  const handleChangeTodoStatus = useCallback(async (todo: Todo) => {
+    const { id, completed } = todo;
+
+    handleSetError({ status: false });
+
+    try {
+      setChangingIds(currValue => [...currValue, id]);
+
+      await changeTodo(id, { completed: !completed });
+      await handleLoadTodos();
+
+      setChangingIds(currValue => currValue.filter(currId => currId !== id));
+    } catch {
+      handleSetError({ status: true, message: 'Unable to update a todo' });
+      setChangingIds(currValue => currValue.filter(currId => currId !== id));
+    }
+  }, []);
+
+  const handleChangeTodoTitle = useCallback(async (
+    id: number,
+    newTitle: string,
+  ) => {
+    handleSetError({ status: false });
+
+    try {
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,16 +1,143 @@
...
+    try {
+      setChangingIds(currValue => [...currValue, id]);
+
+      await deleteTodo(id);
+      await handleLoadTodos();
+
+      setChangingIds(currValue => currValue.filter(currId => currId !== id));
+    } catch {
+      handleSetError({ status: true, message: 'Unable to delete a todo' });
+      setChangingIds(currValue => currValue.filter(currId => currId !== id));
```
illia-sosnytskyi, line 103: maybe for avoiding duplicate this code you should to use `finally` block
```suggestion 
    } catch {
      handleSetError({ status: true, message: 'Unable to delete a todo' });
    } finally {
      setChangingIds(currValue => currValue.filter(currId => currId !== id));
```

illia-sosnytskyi (approved):
nice work

src/App.tsx
OleksandrGlulhovskiy: i think it's not possible

src/App.tsx
```diff
@@ -1,16 +1,143 @@
...
+import { NewTodo } from './components/NewTodo/NewTodo';
+import { TodoError } from './components/TodoError/TodoError';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
-  const newTodoField = useRef<HTMLInputElement>(null);
+  const [userTodos, setUserTodos] = useState<Todo[]>([]);
+  const [hasTodos, setHasTodos] = useState(false);
+  const [hasError, setHasError] = useState<Error>({ status: false });
```
OleksiiBlezniuk, line 28: from `hasError` name I expect that this state will have boolean value, but instead there is an Error object

use more clear names, in this case errorState will be more clear

src/App.tsx
```diff
@@ -1,16 +1,143 @@
...
+  const [isAdding, setIsAdding] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [changingIds, setChangingIds] = useState<number[]>([0]);
+
+  const handleSetError = useCallback((error: Error) => {
+    setHasError(error);
+    setTimeout(() => {
```
OleksiiBlezniuk, line 35: not critical, but should we pass full error object here? I mean if we use this method while errors are appeared, so we'll always have `{ status: true }` part here, maybe let's pass only message here?

src/App.tsx
```diff
@@ -1,16 +1,143 @@
...
+      setHasError({ status: false });
+    }, 3000);
+  }, []);
+
+  const handleLoadTodos = useCallback(async () => {
+    try {
+      setHasError({ status: false });
+
+      const todosFromServer = await getTodos((user as User).id);
```
OleksiiBlezniuk, line 44: avoid typecasting

src/App.tsx
```diff
@@ -1,16 +1,143 @@
...
+    handleLoadTodos();
+  }, []);
+
+  useEffect(() => {
+    if (userTodos.length !== 0) {
+      setHasTodos(true);
+    } else {
+      setHasTodos(false);
+    }
+  }, [userTodos]);
```
OleksiiBlezniuk, line 62: Why do you need this state? You can just compare todos length to 0 and save this result to constant

src/App.tsx
```diff
@@ -1,16 +1,143 @@
...
+    handleSetError({ status: false });
+  }, []);
+
+  const handleAddTodo = useCallback(async (title: string) => {
+    handleSetError({ status: false });
+
+    try {
+      const preparedData = {
+        title,
+        userId: (user as User).id,
```
OleksiiBlezniuk, line 74: avoid typecasting

src/App.tsx
```diff
@@ -1,16 +1,143 @@
...
+        completed: false,
+      };
+
+      setTempTodo({ ...preparedData, id: 0 });
+      setIsAdding(true);
```
OleksiiBlezniuk, line 79: not critical, but more logical to change this state at the beginning of this function

src/App.tsx
```diff
@@ -1,16 +1,143 @@
...
+
+      await addTodo(preparedData);
+      await handleLoadTodos();
+
+      setIsAdding(false);
+    } catch {
+      handleSetError({ status: true, message: 'Unable to add a todo' });
```
OleksiiBlezniuk, line 86: Not critical, but would be nice to move error messages to constants in `constants.ts` file

src/App.tsx
```diff
@@ -1,16 +1,143 @@
...
+      setIsAdding(false);
+    }
+  }, []);
```
OleksiiBlezniuk, line 89: should this array be empty?

src/App.tsx
```diff
@@ -20,203 +147,33 @@ export const App: React.FC = () => {
...
-          </button>
-        </footer>
-      </div>
+        {hasTodos && (
+          <TodoList
+            todos={userTodos}
+            tempTodo={tempTodo}
+            isAdding={isAdding}
+            onDelete={handleDeleteTodo}
+            deletingIds={changingIds}
```
OleksiiBlezniuk, line 165: To understand where this constant is used I had to look on every component props, and search it using browser page search

Of course it's not critical, but would be nice to use same names for prop and value passed here 🙂

src/App.tsx
```diff
@@ -1,16 +1,143 @@
...
+  const handleDeleteTodo = useCallback(async (id: number) => {
+    handleSetError({ status: false });
+
+    try {
+      setChangingIds(currValue => [...currValue, id]);
+
+      await deleteTodo(id);
+      await handleLoadTodos();
+
+      setChangingIds(currValue => currValue.filter(currId => currId !== id));
```
OleksiiBlezniuk, line 100: this array used for ids of todos that are loading now, isn't it? probably better to rename state

src/App.tsx
OleksiiBlezniuk: but why?
and even without finally method you still can move logic out of try/catch block

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,52 @@
...
+  status,
+}) => {
+  return (
+    <nav className="filter" data-cy="Filter">
+      <a
+        data-cy="FilterLinkAll"
+        href="#/"
+        className={classNames(
+          'filter__link',
+          { selected: status === TodoStatus.ALL },
```
OleksiiBlezniuk, line 21: Or you can create array from statuses enum and map it here 🙂 
But it's not critica;

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+    if (query === title) {
+      setTitleIsInput(false);
+      setQuery(title);
+    } else if (query === '') {
+      onDelete(id);
+    } else {
+      onChangeTitle(id, query);
+      setTitleIsInput(false);
+    }
```
OleksiiBlezniuk, line 66: Avoid such big if else constructions

OleksiiBlezniuk, line 66:
```suggestion
    if (query === title) {
      setTitleIsInput(false);
      setQuery(title);
      
      return;
    }
    
    if (query === '') {
      onDelete(id);
      
      return;
    }
    
    onChangeTitle(id, query);
    setTitleIsInput(false);
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,116 @@
...
+        ));
+        break;
+      default:
+        filtered = todos;
+    }
+
+    return filtered;
+  }, [todos, status]);
+
+  const filteredTodos = useMemo(() => filterTodos(), [todos, status]);
```
OleksiiBlezniuk, line 48: overkill
just use `useMemo` for filtering result 🙂

Andr1yK: agree

src/App.tsx
```diff
@@ -1,16 +1,139 @@
...
+import { TodoList } from './components/TodoList/TodoList';
+import { NewTodo } from './components/NewTodo/NewTodo';
+import { TodoError } from './components/TodoError/TodoError';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
-  const newTodoField = useRef<HTMLInputElement>(null);
+  const [userTodos, setUserTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<Error>({ status: false });
```
Andr1yK, line 26: u can just save the message (and check length)🤔

src/App.tsx
```diff
@@ -1,16 +1,139 @@
...
+  }, []);
 
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
-    if (newTodoField.current) {
-      newTodoField.current.focus();
+    handleLoadTodos();
+  }, []);
+
+  const hasTodos = (userTodos.length !== 0);
```
Andr1yK, line 56:
```suggestion
  const hasTodos = userTodos.length !== 0;
```

src/App.tsx
Andr1yK: I think we need to add user here

Andr1yK (changes requested):
Fix or reply to all comments of @OleksiiBlezniuk

PR #147, opened 2022-11-09
solution

myroslav-diiak:
[DEMO LINK](https://myroslav-diiak.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,32 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext, useEffect, useRef, useState,
+} from 'react';
+import cn from 'classnames';
+import {
+  createTodos, getTodos, removeTodo, updateTodo,
+} from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Todo } from './types/Todo';
+import { TodoComponent } from './components/Todo';
+import { Footer } from './components/Footer';
+import { ErrorMessage } from './components/ErrorMessage';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [inputTitle, setInputTitle] = useState('');
+  const [isAdding, setIsAdding] = useState<Todo | null>(null);
+  const [addError, setAddError] = useState(false);
+  const [removeError, setRemoveError] = useState(false);
+  const [updateError, setUpdateError] = useState(false);
+  const [inputError, setInputError] = useState(false);
+  const [filterType, setFilterType] = useState('All');
+  const [isAllSelected, setIsAllSelected] = useState(true);
+  const [isHidden, setIsHidden] = useState(true);
+  const [isEditing, setIsEditing] = useState<Todo | null>(null);
+
+  const editTitle = useRef<HTMLInputElement>(null);
+
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
@@ -14,6 +38,124 @@ export const App: React.FC = () => {
     }
   }, []);
 
+  const getTodosFromApi = async (userId: number) => {
+    try {
+      const response = await getTodos(userId);
+
+      setTodos(response);
+      setIsAdding(null);
+      setIsEditing(null);
+    } catch (error) {
+      throw new Error('Error on loading todos');
+    }
+  };
+
+  const createTodo = async (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!inputTitle) {
+      setInputError(true);
+      setIsHidden(false);
+
+      return;
+    }
+
+    if (user) {
+      const newTodo = {
+        userId: user.id,
+        title: inputTitle,
+        completed: false,
+      };
+
+      setIsAdding(newTodo);
+
+      try {
+        await createTodos(newTodo);
+      } catch {
+        setAddError(true);
+        setIsHidden(false);
+        setIsAdding(null);
+      }
+
+      getTodosFromApi(user.id);
+      setInputTitle('');
+    }
+  };
+
+  const selectCompleted = async (todo: Todo) => {
+    try {
+      await updateTodo(todo, !todo.completed);
+    } catch {
+      setUpdateError(true);
+      setIsHidden(false);
+    }
+
+    if (user) {
+      getTodosFromApi(user.id);
+    }
+  };
+
+  const selectAll = () => {
+    todos.map(async (todo) => {
+      if (!todo.completed && isAllSelected) {
+        await selectCompleted(todo);
+      } else if (todo.completed && !isAllSelected) {
+        await selectCompleted(todo);
+      }
+    });
+
+    setIsAllSelected(prev => !prev);
+  };
+
+  const clearCompleted = () => {
+    todos.map(async todo => {
+      if (todo.completed) {
+        removeTodo(todo);
+        try {
+          await removeTodo(todo);
+        } catch {
+          setRemoveError(true);
+          setIsHidden(false);
+        }
+
+        if (user) {
+          getTodosFromApi(user.id);
+        }
+      }
+    });
+  };
+
+  useEffect(() => {
+    if (user) {
+      getTodosFromApi(user.id);
+    }
+  }, [user]);
+
+  const filteredTodos = todos.filter(todo => {
+    switch (filterType) {
+      case 'Active':
+        return !todo.completed;
+      case 'Completed':
+        return todo.completed;
+      default:
+        return todo;
+    }
+  });
+
+  const clearErrors = () => {
+    setIsHidden(true);
+    setAddError(false);
+    setUpdateError(false);
+    setRemoveError(false);
+    setInputError(false);
+  };
+
+  useEffect(() => {
+    setTimeout(clearErrors, 3000);
+  }, [isHidden]);
+
+  const isAllCompleted = todos.every(todo => todo.completed);
+
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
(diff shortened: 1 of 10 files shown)


jojnajbude (approved):
not bad, but not implemented loader when todo to have updating(not cool) and where the soft transition for list(not cool too).

src/App.tsx
```diff
@@ -1,8 +1,32 @@
...
+import { ErrorMessage } from './components/ErrorMessage';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [inputTitle, setInputTitle] = useState('');
+  const [isAdding, setIsAdding] = useState<Todo | null>(null);
+  const [addError, setAddError] = useState(false);
+  const [removeError, setRemoveError] = useState(false);
+  const [updateError, setUpdateError] = useState(false);
+  const [inputError, setInputError] = useState(false);
```
OleksiiBlezniuk, line 22: do we really need 4 states for error?

OleksiiBlezniuk: probably simpler to set just error state and error message

src/App.tsx
```diff
@@ -1,8 +1,32 @@
...
 import { AuthContext } from './components/Auth/AuthContext';
+import { Todo } from './types/Todo';
+import { TodoComponent } from './components/Todo';
+import { Footer } from './components/Footer';
+import { ErrorMessage } from './components/ErrorMessage';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [inputTitle, setInputTitle] = useState('');
+  const [isAdding, setIsAdding] = useState<Todo | null>(null);
```
OleksiiBlezniuk, line 18: from name `isAdding` I expect to see boolean value here, but instead of this I see Todo or null

it's strange

src/App.tsx
```diff
@@ -14,6 +38,124 @@ export const App: React.FC = () => {
     }
   }, []);
 
+  const getTodosFromApi = async (userId: number) => {
```
OleksiiBlezniuk, line 41: useCallback

src/App.tsx
```diff
@@ -14,6 +38,124 @@ export const App: React.FC = () => {
...
+    try {
+      const response = await getTodos(userId);
+
+      setTodos(response);
+      setIsAdding(null);
+      setIsEditing(null);
+    } catch (error) {
+      throw new Error('Error on loading todos');
+    }
```
OleksiiBlezniuk, line 50: Why to throw error if it's thrown before? Show error message instead

src/App.tsx
```diff
@@ -14,6 +38,124 @@ export const App: React.FC = () => {
...
+
+    if (user) {
+      const newTodo = {
+        userId: user.id,
+        title: inputTitle,
+        completed: false,
+      };
+
+      setIsAdding(newTodo);
```
OleksiiBlezniuk, line 70: in that case probably better to rename this state 🙂  tempTodo or smth like that

src/App.tsx
```diff
@@ -14,6 +38,124 @@ export const App: React.FC = () => {
...
+
+      try {
+        await createTodos(newTodo);
+      } catch {
+        setAddError(true);
+        setIsHidden(false);
+        setIsAdding(null);
+      }
+
+      getTodosFromApi(user.id);
```
OleksiiBlezniuk, line 80: should we load new todos in case of error? and what if there will be an error?

src/App.tsx
```diff
@@ -23,201 +165,70 @@ export const App: React.FC = () => {
...
             <input
               data-cy="NewTodoField"
               type="text"
               ref={newTodoField}
               className="todoapp__new-todo"
               placeholder="What needs to be done?"
+              value={inputTitle}
+              onChange={event => setInputTitle(event.target.value)}
             />
           </form>
```
OleksiiBlezniuk, line 182: Let's move this form to separate component

src/App.tsx
```diff
@@ -14,6 +38,124 @@ export const App: React.FC = () => {
...
+      await updateTodo(todo, !todo.completed);
+    } catch {
+      setUpdateError(true);
+      setIsHidden(false);
+    }
+
+    if (user) {
+      getTodosFromApi(user.id);
+    }
+  };
```
OleksiiBlezniuk, line 96: same here, load todos in try block

src/App.tsx
```diff
@@ -14,6 +38,124 @@ export const App: React.FC = () => {
...
+        setIsHidden(false);
+        setIsAdding(null);
+      }
+
+      getTodosFromApi(user.id);
+      setInputTitle('');
+    }
+  };
+
+  const selectCompleted = async (todo: Todo) => {
```
OleksiiBlezniuk, line 85: useCallback

OleksiiBlezniuk: and again, this name is confusing

from method that called `selectCompleted` I expect that it will select completed todos for something, but in reality this method updating todos on server 

let's use more clear names

src/App.tsx
```diff
@@ -14,6 +38,124 @@ export const App: React.FC = () => {
...
+
+    if (user) {
+      getTodosFromApi(user.id);
+    }
+  };
+
+  const selectAll = () => {
+    todos.map(async (todo) => {
+      if (!todo.completed && isAllSelected) {
+        await selectCompleted(todo);
```
OleksiiBlezniuk, line 101: you've triggered loading todos on every using of this method, this action increases amount of requests to server twice

we don't need to do it every time, let's optimise this method

src/App.tsx
```diff
@@ -14,6 +38,124 @@ export const App: React.FC = () => {
...
+      } else if (todo.completed && !isAllSelected) {
+        await selectCompleted(todo);
+      }
+    });
+
+    setIsAllSelected(prev => !prev);
```
OleksiiBlezniuk, line 107: should we change this value before our functions will update data on server? 

let's use Promise.all to await until our requests will be completed

src/types/Todo.ts
```diff
@@ -1,5 +1,5 @@
 export interface Todo {
-  id: number;
+  id?: number;
```
OleksiiBlezniuk, line 2: how can this field be optional? I mean in all cases even when you show temporary todo we'll have an id

OleksiiBlezniuk (changes requested):
Don't forget to use optimising methods such as useMemo, useCallback, memo

src/App.tsx
```diff
@@ -1,8 +1,30 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [inputTitle, setInputTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterType, setFilterType] = useState('All');
+  const [isAllSelected, setIsAllSelected] = useState(true);
+  const [isHidden, setIsHidden] = useState(true);
+  const [isEditing, setIsEditing] = useState<Todo | null>(null);
```
sergii-nosachenko, line 24: `isEditing` refers to boolean, use proper naming

src/App.tsx
```diff
@@ -14,6 +36,132 @@ export const App: React.FC = () => {
...
+    event.preventDefault();
+
+    if (!inputTitle) {
+      setErrorMessage('Title can\'t be empty');
+      setIsHidden(false);
+
+      return;
+    }
+
+    if (user) {
```
sergii-nosachenko, line 63: Start your function with `if (!user) return` - it's better way to avoid unnecessary execution. Also no need to wrap a lot of code in block.

src/App.tsx
```diff
@@ -14,6 +36,132 @@ export const App: React.FC = () => {
...
+
+      setInputTitle('');
+    }
+  };
+
+  const setTodoSelected = useCallback(async (todo: Todo) => {
+    try {
+      await updateTodo(todo, !todo.completed);
+
+      if (user) {
```
sergii-nosachenko, line 90: What if no user? Better start with checking this condition.

src/App.tsx
```diff
@@ -14,6 +36,132 @@ export const App: React.FC = () => {
...
+          await updateTodo(todo, !todo.completed);
+        } catch {
+          setErrorMessage('Unable to update a todo');
+          setIsHidden(false);
+        }
+      }
+    });
+
+    Promise.all(todoFunctions).then(() => {
+      if (user) {
```
sergii-nosachenko, line 113: What if no user? Better start with checking this condition.

src/App.tsx
```diff
@@ -14,6 +36,132 @@ export const App: React.FC = () => {
...
+      setIsHidden(false);
+    }
+  }, []);
+
+  const selectAll = () => {
+    const todoFunctions = todos.map(async (todo) => {
+      if ((!todo.completed && isAllSelected)
+        || (todo.completed && !isAllSelected)) {
+        try {
+          await updateTodo(todo, !todo.completed);
```
sergii-nosachenko, line 104: Don't need to wait for update if you then call Promise.all(). And why do you use array map if your callback doesn't return anything?
First, fill array with promises (return it here, not await). Then your Promise.all() will work correctly.

src/App.tsx
```diff
@@ -14,6 +36,132 @@ export const App: React.FC = () => {
...
+
+    Promise.all(todoFunctions).then(() => {
+      if (user) {
+        getTodosFromApi(user.id);
+      }
+    });
+  };
+
+  const clearCompleted = () => {
+    todos.map(async todo => {
```
sergii-nosachenko, line 120: Improper usage of array map

src/App.tsx
```diff
@@ -14,6 +36,132 @@ export const App: React.FC = () => {
...
+      if (todo.completed) {
+        removeTodo(todo);
+        try {
+          await removeTodo(todo);
+        } catch {
+          setErrorMessage('Unable to delete a todo');
+          setIsHidden(false);
+        }
+
+        if (user) {
```
sergii-nosachenko, line 130: Fix it

src/App.tsx
```diff
@@ -14,6 +36,132 @@ export const App: React.FC = () => {
...
+
+  useEffect(() => {
+    if (user) {
+      getTodosFromApi(user.id);
+    }
+  }, [user]);
+
+  const filteredTodos = todos.filter(todo => {
+    switch (filterType) {
+      case 'Active':
```
sergii-nosachenko, line 145: Use enum

src/App.tsx
```diff
@@ -23,201 +171,59 @@ export const App: React.FC = () => {
...
-        />
-
-        Unable to add a todo
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
-      </div>
+      <ErrorMessage
+        isHidden={isHidden}
```
sergii-nosachenko, line 223:
```suggestion
      {errorMessage && <ErrorMessage
```
You don't need `isHidden` at all. It's the same as check `errorMessage === ''`.
Also don't render this component at all if no message.

src/App.tsx
```diff
@@ -23,201 +171,59 @@ export const App: React.FC = () => {
           <button
             data-cy="ToggleAllButton"
             type="button"
-            className="todoapp__toggle-all active"
+            className={cn('todoapp__toggle-all', { active: isAllCompleted })}
+            onClick={selectAll}
+          />
```
sergii-nosachenko, line 176: Hidden when no todos

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,189 @@
...
+  editTitle,
+  setIsHidden,
+  isEditting,
+  setIsEditting,
+  setTodoSelected,
+  setErrorMessage,
+}) => {
+  const user = useContext(AuthContext);
+  const [inputTitle, setInputTitle] = useState('');
+  const [isRemoving, setIsRemoving] = useState<Todo | null>(null);
```
sergii-nosachenko, line 30: `isRemoving` refers to boolean type, not Todo

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const [
+    tooggleDoubleClick,
+    setToggleDoubleClick,
+  ] = useState<Todo | null>(null);
```
sergii-nosachenko, line 34: Improper naming

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  const removeTodos = async (toDo: Todo) => {
+    removeTodo(toDo);
+    try {
+      setIsRemoving(toDo);
+      await removeTodo(toDo);
```
sergii-nosachenko, line 40: Repeated code

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,189 @@
...
+    } catch {
+      setErrorMessage('Unable to delete a todo');
+      setIsHidden(false);
+      setIsRemoving(null);
+    }
+
+    if (user) {
```
sergii-nosachenko, line 47: Move up and (!user)

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,189 @@
...
+      }
+
+      setToggleDoubleClick(null);
+      await editTodo(toDo, trimTitle);
+    } catch {
+      setErrorMessage('Unable to update a todo');
+      setIsHidden(false);
+    }
+
+    if (user) {
```
sergii-nosachenko, line 80: Move up and (!user)

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,189 @@
...
+      getTodo(user.id);
+    }
+
+    setToggleDoubleClick(null);
+  };
+
+  useEffect(() => {
```
sergii-nosachenko, line 87: Useless

sergii-nosachenko (changes requested):
A lot of problems from previous stage of TodoApp - look at my comments from there too.

PR #154, opened 2022-11-10
react todo app with api was done

iPanda177:
[SUPER DEMO LINK](https://iPanda177.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,212 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  FormEvent,
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+  useMemo,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
 
+import {
+  getTodos,
+  createTodo,
+  deleteTodos,
+  updateTodo,
+} from './api/todos';
+
+import { SortType } from './types/SortType';
+import { Todo } from './types/Todo';
+import { ErrorType } from './types/ErrorType';
+
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorMessage } from './components/ErrorMessage';
+import { TodoForm } from './components/TodoForm';
+
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
 
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortType, setSortType] = useState<SortType>(SortType.ALL);
+  const [error, setError] = useState('');
+  const [title, setTitle] = useState('');
+  const [selectedId, setSelectedId] = useState<number[]>([]);
+  const [isAdding, setIsAdding] = useState(false);
+
+  const filteredTodo = useMemo(() => todos.filter(todo => {
+    switch (sortType) {
+      case SortType.ALL:
+        return todo;
+      case SortType.ACTIVE:
+        return !todo.completed;
+      case SortType.COMPLETED:
+        return todo.completed;
+      default:
+        return null;
+    }
+  }), [todos, sortType]);
+
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
   }, []);
 
+  useEffect(() => {
+    if (!user) {
+      return;
+    }
+
+    const getTodosFromServer = async (userId: number) => {
+      try {
+        const receivedTodos = await getTodos(userId);
+
+        setTodos(receivedTodos);
+      } catch (errorMessage) {
+        setError(`${errorMessage}`);
+      }
+    };
+
+    getTodosFromServer(user.id);
+  }, []);
+
+  const getValue = (element: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(element.target.value);
+  };
+
+  const addNewTodo = useCallback(async (event: FormEvent) => {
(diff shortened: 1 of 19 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+  setSortType,
+  clearCompletedTodos,
+}) => {
+  const todosLeft = useMemo(() => (
+    todos.filter(({ completed }) => !completed).length
+  ), [todos]);
+
+  const completedTodos = useMemo(() => (
+    todos.filter(todo => todo.completed).length
+  ), [todos]);
```
Andr1yK, line 28: use `reduce` (this will reduce the algorithmic complexity)

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+          data-cy="ClearCompletedButton"
+          type="button"
+          className="todoapp__clear-completed"
+          onClick={clearCompletedTodos}
+          disabled={!completedTodos}
+        >
+          Clear completed
+        </button>
+      ) : (
+        <span>{'             '}</span>
```
Andr1yK, line 64: It's a super weird solution.. mb u can use CSS :)

Andr1yK (changes requested):

src/App.tsx
```diff
@@ -1,223 +1,212 @@
...
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
   }, []);
 
+  useEffect(() => {
+    if (!user) {
+      return;
+    }
+
```
Anna-Tkachenko, line 63: let's combine these useEffects, both of them works as componentDidMount method
why did you decide to create them in separate way?

src/App.tsx
```diff
@@ -1,223 +1,212 @@
...
+        setError(`${errorMessage}`);
+      }
+    };
+
+    getTodosFromServer(user.id);
+  }, []);
+
+  const getValue = (element: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(element.target.value);
+  };
```
Anna-Tkachenko, line 79: should it be covered with useCallback?

src/App.tsx
```diff
@@ -1,223 +1,212 @@
...
+    try {
+      await deleteTodos(TodoId);
+
+      setTodos((prevTodos) => prevTodos.filter(({ id }) => id !== TodoId));
+    } catch {
+      setError(ErrorType.NotDelete);
+    }
+  }, [todos, error]);
+
+  const completedTodos = todos.filter(({ completed }) => completed);
```
Anna-Tkachenko, line 116: should it be covered with useMemo?

src/App.tsx
```diff
@@ -1,223 +1,212 @@
...
+
+  const clearCompletedTodos = useCallback(
+    async () => {
+      try {
+        setSelectedId([...completedTodos].map(({ id }) => id));
+
+        await Promise.all(completedTodos.map(({ id }) => removeTodo(id)));
+        await (() => setTodos((prevTodos) => prevTodos
+          .filter(({ completed }) => !completed)));
```
Anna-Tkachenko, line 125: what's the reason of using await here?

src/App.tsx
```diff
@@ -1,223 +1,212 @@
...
+      }
+    }, [todos, selectedId, error],
+  );
+
+  const handleChange = async (todoId: number, data: Partial<Todo>) => {
+    setSelectedId([todoId]);
+
+    function updateStatus(value: Todo) {
+      return value;
+    }
```
Anna-Tkachenko, line 138: for what purpose did you create this function?

src/App.tsx
```diff
@@ -1,223 +1,212 @@
...
+          : todo
+      )));
+    } catch {
+      setError(ErrorType.NotUpdate);
+    }
+
+    setSelectedId([]);
+  };
+
+  const handleAllCompleted = () => {
```
Anna-Tkachenko, line 155: should it be covered with useCallback?

src/api/todos.ts
```diff
@@ -5,4 +5,18 @@ export const getTodos = (userId: number) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const createTodo = (title: string, userId: number) => {
```
Anna-Tkachenko, line 8: add types for the returned value

Anna-Tkachenko (changes requested):

src/App.tsx
iPanda177: i have linter error if won't add it

src/App.tsx
```diff
@@ -1,223 +1,210 @@
...
+    setTitle('');
+  }, [title, user]);
+
+  const removeTodo = useCallback(async (TodoId: number) => {
+    setSelectedIds([TodoId]);
+
+    try {
+      await deleteTodos(TodoId);
+
+      setTodos((prevTodos) => prevTodos.filter(({ id }) => id !== TodoId));
```
Anna-Tkachenko, line 110: same here, let's call method for getting todos from server instead of changing it manually

src/App.tsx
Anna-Tkachenko: let's also call here the function for getting actual todos from server and update them in state instead of doing it manually

Denchepornyuk (changes requested):
test reject

src/api/todos.ts
```diff
@@ -5,4 +5,18 @@ export const getTodos = (userId: number) => {
...
+  return client.post<Todo>('/todos', {
+    userId,
+    completed: false,
+    title,
+  });
+};
+
+export const deleteTodos = (todoId: number) => {
```
Anna-Tkachenko, line 16:
```suggestion
export const deleteTodo = (todoId: number) => {
```

Anna-Tkachenko (approved):
Good job! 🔥

PR #157, opened 2022-11-10
solution was added

Maksym845:
- [DEMO LINK](https://Maksym845.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,52 +28,42 @@ export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
   const [selectedFilter, setSelectedFilter] = useState<FilterBy>(FilterBy.ALL);
-  const [counter, setCounter] = useState(0);
 
   const [newTodoTitle, setNewTodoTitle] = useState('');
-  const [emptyTitleError, setEmptyTitleError] = useState(false);
   const [isAdding, setIsAdding] = useState(false);
   const [deletingCompletedTodos, setDeletingCompletedTodos] = useState(false);
-
-  const [addingError, setAddingError] = useState(false);
-  const [deletingError, setDeletingError] = useState(false);
   const [addingNewTodo, setAddingNewTodo] = useState(false);
-  const [updatingError, setUpdatingError] = useState(false);
+
+  const [errorMessage, setErrorMessage] = useState('');
 
   const [toggleAll, setToggleAll] = useState(false);
 
-  const increaseCounter = () => {
-    setCounter(prevState => prevState + 1);
-  };
-
   const filterTodos = (filterBy: FilterBy) => {
-    if (todos) {
-      setVisibleTodos(todos.filter((todo: Todo) => {
-        switch (filterBy) {
-          case FilterBy.ACTIVE:
-            return !todo.completed;
-          case FilterBy.COMPLETED:
-            return todo.completed;
-          default:
-            return true;
-        }
-      }));
-    }
+    setVisibleTodos(() => todos.filter((todo: Todo) => {
+      switch (filterBy) {
+        case FilterBy.ACTIVE:
+          return !todo.completed;
+        case FilterBy.COMPLETED:
+          return todo.completed;
+        default:
+          return true;
+      }
+    }));
   };
 
   const todosLength = () => {
-    return todos?.filter(todo => !todo.completed).length || '0';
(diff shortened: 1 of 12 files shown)


Veremchukk (approved):
GJ. But try to fix this problem

Maksym845:
> GJ. But try to fix this problem

Yes, I forgot about this problem, thank you)

src/App.tsx
```diff
@@ -1,18 +1,206 @@
...
+  const [updatingError, setUpdatingError] = useState(false);
+
+  const [toggleAll, setToggleAll] = useState(false);
+
+  const increaseCounter = () => {
+    setCounter(prevState => prevState + 1);
+  };
+
+  const filterTodos = (filterBy: FilterBy) => {
+    if (todos) {
```
Anna-Tkachenko, line 50: this check is redundant, you always have the link to array in todos variable

src/App.tsx
```diff
@@ -1,18 +1,206 @@
...
+      setVisibleTodos(todos.filter((todo: Todo) => {
```
Anna-Tkachenko, line 51: use functional setState if you change it according to the current value

src/App.tsx
```diff
@@ -1,18 +1,206 @@
...
+          default:
+            return true;
+        }
+      }));
+    }
+  };
+
+  const todosLength = () => {
+    return todos?.filter(todo => !todo.completed).length || '0';
+  };
```
Anna-Tkachenko, line 66: todos variable always store an array

src/App.tsx
```diff
@@ -1,18 +1,206 @@
...
+
+  const todosCompletedLength = () => {
+    return todos?.filter(todo => todo.completed).length;
+  };
+
+  const getTodosFromApi = useCallback(async () => {
+    try {
+      if (user) {
+        setIsAdding(true);
+        const apiTodos = await getTodos(user.id).then(gotTodos => gotTodos);
```
Anna-Tkachenko, line 76: then after await function is redundant here

src/App.tsx
```diff
@@ -1,18 +1,206 @@
...
+      setCounter((prevState) => prevState + 1);
+    }
+
+    setNewTodoTitle('');
+    setAddingNewTodo(false);
+  };
+
+  const onDeleteTodo = async (id: number) => {
+    await deleteTodo(id);
+    increaseCounter();
```
Anna-Tkachenko, line 158: for what purpose did you add counter value? can we calculate it according to the total length of todos?

src/App.tsx
```diff
@@ -1,18 +1,206 @@
...
+  };
+
+  const onToggleAll = (newStatus: boolean) => {
+    setToggleAll(prevState => !prevState);
+    try {
+      todos.forEach(async (todo) => {
+        await updateTodo({
+          ...todo,
+          completed: newStatus,
+        }, todo.id);
```
Anna-Tkachenko, line 183: let's not change todos that are already completed / uncompleted

src/App.tsx
```diff
@@ -23,201 +211,108 @@ export const App: React.FC = () => {
...
+              </button>
+            </footer>
+          </>
+        )}
       </div>
+
+      {(emptyTitleError
+        || addingError
+        || deletingError
+        || updatingError) && (
```
Anna-Tkachenko, line 306: let's create separate variable for storing the final result of this condition and use this variable in jsx

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,18 +1,201 @@
...
+  };
+
+  const onDeleteTodo = async (id: number) => {
+    await deleteTodo(id);
+    await getTodosFromApi();
+  };
+
+  const handleClearCompleted = () => {
+    setDeletingCompletedTodos(true);
+    todos?.forEach(async (todo) => {
```
Anna-Tkachenko, line 151: todos variable can't be undefined or null, so ? sign is redundant here

src/App.tsx
```diff
@@ -23,201 +206,102 @@ export const App: React.FC = () => {
...
+            <TodosList
+              todos={visibleTodos}
+              onDeleteTodo={onDeleteTodo}
+              addingNewTodo={addingNewTodo}
+              tempTodo={{
+                id: 0,
+                userId: user.id,
+                title: newTodoTitle,
+                completed: false,
+              }}
```
Anna-Tkachenko, line 240: will be better to create separate variable for storing this object

Anna-Tkachenko (approved):
Good job! 🔥

PR #158, opened 2022-11-10
react_todo-app-with-api

xiimiik:
[DEMO LINK](https://xiimiik.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -147,6 +147,25 @@ export const App: React.FC = () => {
     }
   }, [todos]);
 
+  const setNewTodoTitleToServer = useCallback(async (
+    todoId: number,
+    newTitle: string,
+  ) => {
+    try {
+      setTodoIdsToRemove(currIds => [...currIds, todoId]);
+
+      await changeTodo(todoId, { title: newTitle });
+      await getTodosFromServer();
+
+      setTodoIdsToRemove(currIds => (
+        currIds.filter((id) => id !== todoId)
+      ));
+    } catch (error) {
+      setHasError(true);
+      setErrorMessage('Unable to update a todo');
+    }
+  }, []);
+
   const closeNotification = useCallback(() => setHasError(false), []);
 
   useEffect(() => {
@@ -188,6 +207,7 @@ export const App: React.FC = () => {
               tempTodo={tempTodo}
               todoIdsToRemove={todoIdsToRemove}
               toggleTodoStatus={toggleTodoStatus}
+              changeTitle={setNewTodoTitleToServer}
             />
 
             <Footer
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -44,7 +44,7 @@ export const App: React.FC = () => {
       const todosFromServer = await getTodos(user.id);
 
       setTodos(todosFromServer);
-    } catch (error) {
+    } catch {
       setErrorMessage('Can\'t download ToDos from server');
     }
   }, []);
@@ -68,7 +68,7 @@ export const App: React.FC = () => {
       await addTodo(currTodo);
       await getTodosFromServer();
       setIsAdding(false);
-    } catch (error) {
+    } catch {
       setErrorMessage('Unable to add a todo');
     }
   }, []);
@@ -83,7 +83,7 @@ export const App: React.FC = () => {
       setTodoIdsLoading(currIds => (
         currIds.filter((id) => id !== todoId)
       ));
-    } catch (error) {
+    } catch {
       setErrorMessage('Unable to remove ToDo');
     }
   }, []);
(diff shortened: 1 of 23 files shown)


vladyslav-harkusha (approved):
There are no loaders when todo status is changing

src/App.tsx
```diff
@@ -1,223 +1,232 @@
...
+        default:
+          return true;
+      }
+    })
+  ), [todos, filterType]);
+
+  const removeAllCompletedTodos = useCallback(async () => {
+    try {
+      await Promise.all(completedTodos.map(async ({ id }) => (
+        removeTodoFromServer(id)
```
OleksiiBlezniuk, line 113: Now inside this method you trigger loading of new todos after every deleted todo, let's separate it from deleting method and load new todos after deleting of all todos will be done

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,74 @@
...
+  onRemove,
+  todosLength,
+}) => {
+  const uncompletedCount = todosLength - completedTodos;
+
+  const filterHrefByType = useMemo(() => ({
+    [FilterType.All]: '#/',
+    [FilterType.Active]: '#/active',
+    [FilterType.Completed]: '#/completed',
+  }), []);
```
OleksiiBlezniuk, line 26: As for me this object should be in constants file, 'cause there its not depends on props or any component data to store it here

OleksiiBlezniuk (changes requested):
GJ! There is few small issues to fix and task is completed

Can you add loader on todo status loader? 🙂 

https://user-images.githubusercontent.com/84275496/201307454-522c3d1d-10e8-4301-8852-50865d2d52e2.mov

src/App.tsx
```diff
@@ -1,223 +1,255 @@
...
+      }
+    } catch (error) {
+      setHasError(true);
+      setErrorMessage('Can`t download ToDos from server');
+    }
+  }, []);
+
+  const addNewTodoToServer = useCallback(async (title: string) => {
+    try {
+      if (user) {
```
sergii-nosachenko, line 56: Better use negative logic - if (!user) return;

src/App.tsx
```diff
@@ -1,223 +1,255 @@
...
+      }
+    } catch (error) {
+      setHasError(true);
+      setErrorMessage('Unable to add a todo');
+    }
+  }, []);
+
+  const removeTodoFromServer = useCallback(async (todoId: number) => {
+    try {
+      if (user) {
```
sergii-nosachenko, line 79: Better use negative logic - if (!user) return;

src/App.tsx
```diff
@@ -1,223 +1,255 @@
...
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState<FilterType>(FilterType.All);
+  const [hasError, setHasError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isAdding, setIsAdding] = useState(false);
+  const [todoIdsToRemove, setTodoIdsToRemove] = useState<number[]>([]);
```
sergii-nosachenko, line 33: As I understand your logic - this state holds ids of todos that have loading overlay. But why it's called `todosIdsToRemove`?

src/utils/fetchClient.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
sergii-nosachenko, line 1: Instead of disabling linter use `unknown` type

sergii-nosachenko (changes requested):
Good solid work, just fix some minor issues.

src/App.tsx
```diff
@@ -1,223 +1,255 @@
...
+import { FilterType } from './types/FilterType';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState<FilterType>(FilterType.All);
+  const [hasError, setHasError] = useState(false);
```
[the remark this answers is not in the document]
xiimiik: Maybe it will be better, but as for me code readability becomes higher and memory loss is not so significant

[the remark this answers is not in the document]
sergii-nosachenko: @xiimiik I don't understand your argument - you use two state values for the same reason. x2 hook calls every time you need to change error. Hmm, it's not about memory loss, it's about managing your code.

sergii-nosachenko (approved):
Awesome job 🚀🚀🚀
Read comments.

PR #172, opened 2022-11-13
Develop

SerhiiRomaniukWD:
[DEMO LINK](https://serhiiromaniukwd.github.io/react_todo-app-with-api/)


Anna-Borodaienko (commented):
1. After adding new Todo there isn't focus on the NewTodoField.
2. There isn't button 'Clear all completed', if not all todos are completed

src/components/MainContent/MainContent.tsx
```diff
@@ -0,0 +1,266 @@
...
+      // eslint-disable-next-line
+    } catch (error: any) {
+      // eslint-disable-next-line no-console
+      console.log('error', error.message);
+      setError('Unable to update a todo');
+    }
+  };
+
+  const changeCompletedForGroup = async (group: Todo[], value: boolean) => {
+    group.map(todo => addOnLoadingTodo(todo.id));
```
Anna-Tkachenko, line 154: if you don't need the result of the map method use forEach, that doesn't return nothing

src/components/MainContent/MainContent.tsx
```diff
@@ -0,0 +1,266 @@
...
+    await Promise.all(
+      todos.map(todo => editTodo(todo.id, { completed: value })),
```
Anna-Tkachenko, line 156: same here

src/components/MainContent/MainContent.tsx
```diff
@@ -0,0 +1,266 @@
...
+    );
+    await loadTodosFromServer();
+    todos.map(todo => deleteFromLoadingTodo(todo.id));
```
Anna-Tkachenko, line 159: and here

Anna-Tkachenko: can we make needed changes in one iteration instead of 3?

src/components/MainContent/MainContent.tsx
```diff
@@ -0,0 +1,266 @@
...
+      // eslint-disable-next-line no-console
+      console.log('error', error.message);
+      setError('Can\'t change todos');
+    }
+  };
+
+  const deleteAllCompletedTodos = async () => {
+    try {
+      await Promise.all(
+        completedTodos.map(todo => deleteTodoOnServer(todo.id)),
```
Anna-Tkachenko, line 180: use forEach

src/components/MainContent/MainContent.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+      {todos.length > 0 && (
+        <footer className="todoapp__footer" data-cy="Footer">
+          <span className="todo-count" data-cy="todosCounter">
+            {`${activeTodos.length} items left`}
+          </span>
+
+          <nav className="filter" data-cy="Filter">
+            {filterList.map((filter, index) => {
+              const newKey = filter + index;
```
Anna-Tkachenko, line 238: you shouldn't use index in process of generating key for the elements
in this way the key will be different if one of elements will be deleted, and it won't help React to optimize rendering for such elements

Anna-Tkachenko (approved):
Good job!

src/components/MainContent/MainContent.tsx
SerhiiRomaniukWD: I can't use forEach in Promise.all cause I need new array

PR #176, opened 2022-11-13
react todo app with api

Arsen-Shpak:
[DEMO LINK](https://Arsen-Shpak.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, {
-  useContext, useEffect, useRef, useState,
+  useContext, useEffect, useRef, useState, useCallback, useMemo,
 } from 'react';
 import cn from 'classnames';
 import { AuthContext } from './components/Auth/AuthContext';
@@ -14,7 +14,7 @@ import { TodoList } from './components/Auth/TodoList';
 import { Footer } from './components/Auth/Footer';
 import { Error } from './components/Auth/Error';
 import { FilterTodos } from './utils/FilterTodos';
-// import { User } from './types/User';
+
 const defaultTodo = {
   id: 0,
   userId: 0,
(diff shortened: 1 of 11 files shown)


src/components/Auth/AddTodo.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+type Props = {
+  newTodoField: React.RefObject<HTMLInputElement>;
```
Anna-Borodaienko, line 4: But you don't use this ```ref``` here.
And after adding you don't focus on the input field.
Add the same UseEffect with this ```ref``` as at App.tsx

src/App.tsx
```diff
@@ -1,223 +1,252 @@
...
+
+      return todo;
+    });
+  };
+
+  // const handleEditTodo = async (id: number, comleted: boolean) => {
+  //   setIsCompleted(comleted);
+  //   editTodo(id, comleted);
+  //   getTodosFromServer();
+  // };
```
Anna-Borodaienko, line 111: Delete commented parts

src/App.tsx
```diff
@@ -1,223 +1,252 @@
...
+        {todos.length > 0 && (
+          <TodoList
+            todos={filteredTodos}
+            getTodosFromServer={getTodosFromServer}
+            isAdding={isAdding}
+            tempTodo={tempTodo}
+            setHasError={setHasError}
+            setMessageError={setMessageError}
+            setIsLoading={setIsLoading}
+            isLoading={isLoading}
```
Anna-Borodaienko, line 230: It's not a good practice to use state-changing methods directly, maybe you can add handle-functions for them

Anna-Borodaienko (commented):
1. "if new title is the same as the old one just cancel editing" - but I see the loader
2. If I edit one todo and leave empty title, it deletes. But then all other attemps to edit todos work incorrect.
3. If all todos are completed, toggleAll button doesn't work.
4. When I change status of some todos using toggleAll button, there are not loaders on these todos.

Arsen-Shpak:
> 1. "if new title is the same as the old one just cancel editing" - but I see the loader
> 2. If I edit one todo and leave empty title, it deletes. But then all other attemps to edit todos work incorrect.
> 3. If all todos are completed, toggleAll button doesn't work.
> 4. When I change status of some todos using toggleAll button, there are not loaders on these todos.

PR #177, opened 2022-11-13
toggle and update

korniienkokarina:
[DEMO LINK](https://korniienkokarina.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Errors/Errors.tsx
+++ b/src/components/Errors/Errors.tsx
@@ -1,0 +1,40 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import classNames from 'classnames';
+import React, { useEffect } from 'react';
+import { Error } from '../../types/Error';
+
+type Props = {
+  hasError: boolean;
+  onCloseError: () => void;
+  error: Error,
+};
+
+export const Errors: React.FC<Props> = ({ hasError, onCloseError, error }) => {
+  useEffect(() => {
+    setTimeout(() => {
+      onCloseError();
+    }, 3000);
+  }, [hasError]);
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={classNames(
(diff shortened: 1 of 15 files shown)


src/components/UpdateTitleForm/UpdateTitleForm.tsx
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <form
+      onSubmit={(event) => {
+        event.preventDefault();
+        handleSubmit(newTitle, id);
+      }}
+      onBlur={(event) => {
+        event.preventDefault();
+        handleSubmit(newTitle, id);
+      }}
```
ArturNagaycev, line 40: functions look similar. it's better to move them to separate variable

src/components/Errors/Errors.tsx
```diff
@@ -0,0 +1,41 @@
...
+        type="button"
+        className="delete"
+        onClick={onCloseError}
+      />
+      {error === Error.OnAdding && 'Unable to add a todo'}
+      {error === Error.OnTitle && 'Unable to add a todo'}
+      {error === Error.OnDeleting && 'Unable to delete a todo'}
+      {error === Error.OnLoading && 'Loading failed'}
+      {error === Error.OnToggleAll && 'Toggle failed'}
+      {error === Error.OnUpdating && 'Title updating failed'}
```
ArturNagaycev, line 38: would be better to save error text in state and just use it here

ArturNagaycev (approved):
GJ ⭐ 

left only 2 comments for you

silchyn (changes requested):
> `NewTodo` form should trim redundant spaces.

silchyn (changes requested):
Are you sure app works correctly now?

stukankarina:
> 

i don't see any issues here 🙈

PR #182, opened 2022-11-14
implemented APP with API

VovaDzhuryk:
[DEMO LINK](https://VovaDzhuryk.github.io/react_todo-app-with-api/)


dsydorenko534 (approved):
I see you had a problem in your commit (with react transition group). So for fix it i recommend you use these commands: (npm install react-transition-group --save) and (npm install @types/react-transition-group) - for types.

sergii-nosachenko (changes requested):
You need to pass all tests before submitting your work. Fix it.

PR #184, opened 2022-11-14
create third part of todo task

AndriyZhabynets:
[DEMO LINK](https://AndriyZhabynets.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,222 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useEffect, useRef, useState, useContext, useCallback, FormEvent, useMemo,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
+import { FilterTypes } from './types/FilterType';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+import { TodoList } from './components/Todolist';
+import {
+  getTodos,
+  createTodo,
+  deleteTodos,
+  updateTodo,
+} from './api/todos';
+
+export function getFilterTodos(
+  todos: Todo[],
+  filterTypes: FilterTypes,
+) {
+  const filterTodo = [...todos];
+
+  switch (filterTypes) {
+    case FilterTypes.Active:
+      return filterTodo.filter(({ completed }) => !completed);
+
+    case FilterTypes.Completed:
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,223 +1,222 @@
...
+              loadingTodoIds={loadingTodoIds}
+              changeTodo={changeTodo}
+            />
+            <Footer
+              getFilterTodo={setFilterBy}
+              selectedTab={filterBy}
+              deleteCompletedTodos={deleteCompletedTodos}
+              todos={todos}
+              todoActive={todoActive}
             />
```
Oleksandra-Shevchenko, line 209: you need to hide footer  if there are no todos :)

src/components/Header.tsx
```diff
@@ -0,0 +1,53 @@
...
+        aria-label="button"
+        className={classnames(
+          'todoapp__toggle-all', {
+            active: todoActive.length === 0,
+          },
+        )}
+        onClick={() => {
+          handleToggleAll();
+        }}
+      />
```
Oleksandra-Shevchenko, line 38: and this button too :)

Oleksandra-Shevchenko (approved):
GJ 👍

src/App.tsx
```diff
@@ -1,223 +1,222 @@
...
+      }
+    }, [todos],
+  );
+
+  const completedTodos = useMemo(() => (
+    todos.filter(todo => todo.completed)
+  ), [todos]);
+
+  const deleteCompletedTodos = useCallback(() => {
+    Promise.all(completedTodos.map(({ id }) => deleteTodo(id)))
```
Anna-Tkachenko, line 131: if you don't need the result of map method, use forEach instead of it

src/components/Footer.tsx
```diff
@@ -0,0 +1,89 @@
...
+        </a>
+      </nav>
+
+      <button
+        data-cy="ClearCompletedButton"
+        type="button"
+        className={classnames(
+          'todoapp__clear-completed',
+        )}
+        onClick={deleteCompletedTodos}
```
Anna-Tkachenko, line 82: you shouldn't use classnames here

src/components/TodoList.tsx
```diff
@@ -0,0 +1,58 @@
...
+          loadingTodoIds={loadingTodoIds}
+          isAdding={isAdding}
+          changeTodo={changeTodo}
+          newTodoField={newTodoField}
+        />
+      ))}
+
+      {isAdding && (
+        <Todos
+          key={Math.random()}
```
Anna-Tkachenko, line 41: for what purpose did you add key attribute here? and why it has random value?

Anna-Tkachenko (changes requested):

PR #185, opened 2022-11-14
buttons was implemented

vallerii:


Oleksandra-Shevchenko (changes requested):
You have problems with your tests and don`t have a demo link 😿

PR #189, opened 2022-11-15
partially implemented solution

ArtemHulkoDev:
[DEMO LINK](https://ArtemHulkoDev.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,141 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+import {
+  addTodo,
+  changeTodo,
+  getTodos,
+  removeTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { Types } from './types/Types';
+import { TodoList } from './components/TodoList/TodoList';
+import { TodoFilter } from './components/TodoFilter/TodoFilter';
+import { NewTodo } from './components/NewTodo/NewTodo';
+import { ErrorNote } from './components/ErrorNote/ErrorNote';
+
+const FILTERS: Types = {
+  all: 'all',
+  completed: 'completed',
+  active: 'active',
+};
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState(FILTERS.all);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [errorStatus, setErrorStatus] = useState(false);
+  const [isAdding, setIsAdding] = useState(false);
+  const [isRemoving, setIsRemoving] = useState(0);
+  const [tempTodo, setTempTodo] = useState<Todo>({
+    id: 0,
+    title: '',
+    userId: 0,
+    completed: false,
+  });
+
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+
+  const changeErrorStatus = () => setErrorStatus(false);
+
+  async function loadTodos() {
+    if (user) {
+      const todosFromServer = await getTodos(user.id);
+
+      setTodos(todosFromServer);
+    }
+  }
+
+  const addTodoOnServer = async (todoTitle: string) => {
+    if (user) {
+      try {
+        setIsAdding(true);
+        setTempTodo({
+          id: 0,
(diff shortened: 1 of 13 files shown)


Oleksandra-Shevchenko:
something wrong with tests 😢

Oleksandra-Shevchenko (changes requested):
😞

ArtemHulkoDev:
tests don't pass due to the react-transition-group mod installed by me in order to make smooth animations :(

Oleksandra-Shevchenko (approved):
GJ 👍

src/App.tsx
```diff
@@ -1,223 +1,272 @@
...
+};
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState(FILTERS.all);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [errorStatus, setErrorStatus] = useState(false);
+  const [isAdding, setIsAdding] = useState(false);
+  const [isRemoving, setIsRemoving] = useState<number[]>([]);
```
Anna-Tkachenko, line 35: it's not clear, why isRemoving variable, that probably should store some boolean value, stores array of numbers instead
check the variable name

src/App.tsx
```diff
@@ -1,223 +1,272 @@
...
+        });
+        setIsAdding(true);
+
+        const newTodo = await addTodo({
+          title: todoTitle,
+          userId: user.id,
+          completed: false,
+        });
+
+        setTodos(currTodos => [...currTodos, newTodo]);
```
Anna-Tkachenko, line 73: don't add todo manually in state, call method for getting actual data from api and update data in state instead of it

src/App.tsx
```diff
@@ -1,223 +1,272 @@
...
+      }
+    }
+  };
+
+  const removeTodoFromServer = async (todoId: number) => {
+    try {
+      setIsRemoving(prevIds => [...prevIds, todoId]);
+      await removeTodo(todoId);
+
+      setTodos(currTodos => currTodos.filter(todo => todo.id !== todoId));
```
Anna-Tkachenko, line 94: same here, don't change state manually, get actual data from server

src/App.tsx
```diff
@@ -1,223 +1,272 @@
...
+      setIsRemoving(prevIds => prevIds.filter(id => id !== todoId));
+    } catch {
+      setErrorMessage('Unable to delete a todo');
+      setErrorStatus(true);
+    }
+  };
+
+  const removeCompletedTodos = async () => {
+    try {
+      await Promise.all(todos.map(todo => {
```
Anna-Tkachenko, line 104: if you don't need the result of the map method, use for each instead of it

src/App.tsx
```diff
@@ -1,223 +1,272 @@
...
+      setErrorStatus(true);
+    }
+  };
+
+  const changeTitleOnServer = async (todoId: number, title: string) => {
+    try {
+      setIsRemoving(prevIds => [...prevIds, todoId]);
+      await changeTodo({ title }, todoId);
+
+      setTodos(currTodos => currTodos.map((todo: Todo) => {
```
Anna-Tkachenko, line 150: here will be also good to get actual data from server

Anna-Tkachenko (changes requested):

src/App.tsx
ArtemHulkoDev: done

src/App.tsx
ArtemHulkoDev: done

src/App.tsx
ArtemHulkoDev: done

src/App.tsx
ArtemHulkoDev: done

src/App.tsx
ArtemHulkoDev: unfortunately, it is showing an error when I try to use forEach with promise.all as it expects an array as an argument, tried to find a solution and didn't find any(

src/components/Todo/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const [isTitleClicked, setIsTitleClicked] = useState(false);
+  const [query, setQuery] = useState(todo.title);
+  const newTodoField = useRef<HTMLInputElement>(null);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const trimmedQuery = query.trim();
+
+    if (trimmedQuery === '') {
```
OleksiiBlezniuk, line 29:
```suggestion
    if (!trimmedQuery) {
```

OleksiiBlezniuk (approved):

PR #193, opened 2022-11-20
Solved. Ver.1.0

PaulVoron:
[DEMO LINK](https://PaulVoron.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,203 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useState,
+  useRef,
+} from 'react';
+
+import classNames from 'classnames';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Loader } from './components/Loader';
+import { TodoList } from './components/TodoList';
+import { Notification } from './components/Notification';
+
+import {
+  createTodo,
+  getTodos,
+  removeTodo,
+  changeTodo,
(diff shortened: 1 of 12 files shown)


Anastasiia-Svintsova (changes requested):
I waited a very long time for the page to load (almost 5 minutes). Then it took a very long time to add an item to the list. Can you please check? I checked my Internet connection -  everything is fine. If it's something with the server, add some timeout and a message that the list cannot be loaded, and try again.
Animation

src/App.tsx
```diff
@@ -23,201 +208,102 @@ export const App: React.FC = () => {
...
-          <button
-            data-cy="ClearCompletedButton"
-            type="button"
-            className="todoapp__clear-completed"
-          >
-            Clear completed
-          </button>
-        </footer>
-      </div>
+        {(todos.length !== 0 || isAdding) && (
```
Anastasiia-Svintsova, line 235:
```suggestion
        {(!!todos.length || isAdding) && (
```

src/App.tsx
```diff
@@ -1,18 +1,203 @@
...
+    if (user) {
+      getTodos(user.id).then(todosFromServer => {
+        setTodos(todosFromServer);
+        setIsLoading(false);
+      })
+        .catch(() => {
+          setIsLoading(false);
+          setNotification('Cannot load todos from server');
+        });
+    }
```
Anastasiia-Svintsova, line 48: It's better to move async function to a separate function and just call it in useEffect

Anastasiia-Svintsova (approved):

PR #196, opened 2022-12-02
Solution

bohdan-tron:
[DEMO LINK](https://bohdan-tron.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/ErrorNotifications/ErrorNotifications.tsx
+++ b/src/components/ErrorNotifications/ErrorNotifications.tsx
@@ -1,0 +1,33 @@
+import classNames from 'classnames';
+import { Error } from '../../types/Error';
+import '../../styles/index.scss';
+
+type Props = {
+  isError: Error;
+  handleErrorClose: () => void;
+};
+
+export const ErrorNotifications: React.FC<Props> = ({
+  isError,
+  handleErrorClose,
+}) => {
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={classNames(
+        'notification is-danger is-light has-text-weight-normal',
+        { hidden: !isError.status },
+      )}
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        aria-label="Hide Error"
+        onClick={handleErrorClose}
+      />
+
+      {isError.message}
+    </div>
+  );
+};
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,91 @@
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import '../../styles/index.scss';
+import { Filter } from '../../types/Filter';
+
+type Props = {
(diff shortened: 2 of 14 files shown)


src/components/ErrorNotifications/ErrorNotifications.tsx
```diff
@@ -0,0 +1,33 @@
...
+      )}
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        aria-label="Hide Error"
+        onClick={handleErrorClose}
+      />
+
```
loralevitska, line 29:
```suggestion
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const completedTodos = todos.filter((todo: Todo) => todo.completed);
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="todosCounter">
+        {todosCounter === 1 ? (
+          `${todosCounter} item left`
+        ) : (
+          `${todosCounter} items left`
+        )}
```
loralevitska, line 30: What is the difference in your code depending on the condition?

```suggestion
        `${todosCounter} item left`
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+      )}
+
+      <form onSubmit={handleTodoCreate}>
+        <input
+          data-cy="NewTodoField"
+          type="text"
+          ref={newTodoField}
+          className={classNames(
+            'todoapp__new-todo',
+          )}
```
loralevitska, line 49: Why library when there is only one class?

```suggestion
          className='todoapp__new-todo'
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,159 @@
...
+                </form>
+              ) : (
+                <>
+                  <span
+                    data-cy="TodoTitle"
+                    className="todo__title"
+                    onDoubleClick={() => {
+                      setTodoIdToChange(todo.id);
+                      setNewTitle(todo.title);
+                    }}
```
loralevitska, line 88: Move to separate function  (transfer the function to a variable, and here only write the name of the function)

loralevitska (changes requested):
good job! Let's improve

src/components/Footer/Footer.tsx
loralevitska: for user yes, but your code was duplicated in both cases. Now it's fixed. GJ!

src/components/TodoList/TodoList.tsx
loralevitska: In general, as you wrote, it is good
As you want to write use [data-attributes](https://www.pluralsight.com/guides/how-to-access-custom-attributes-from-aevent-object-in-react)
image

loralevitska (approved):
good job!

PR #197, opened 2022-12-04
made app todo part-3

Hlazman:
DEMO LINK
https://hlazman.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,222 +1,215 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Header } from './components/Header/Header';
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,222 +1,215 @@
...
+    setIsError(true);
+    setErrorMessage(message);
+    setTimeout(closeError, 3000);
+  };
+
+  const loadTodos = useCallback(async () => {
+    try {
+      if (user) {
+        await getTodos(user.id)
+          .then(data => setTodos(data));
```
vlad-demchuk, line 50: let's not use `then` method inside `try/catch` statement. just set a value using a separate variable
`const todosFromServer = await getTodos(user.id);
setTodos(todosFromServer);`

vlad-demchuk (changes requested):
I see some text after adding. Show loader while task is being added
https://user-images.githubusercontent.com/102411130/205587466-74ba3a52-b56d-4122-a551-1ef760c4b8b3.mp4

vlad-demchuk (changes requested):
and one more issue. I can add a lot of todos clicking `enter` many times
Dec-08-2022 11-34-13

nazarmatsevych (approved):

PR #201, opened 2022-12-13
Implement updating todos ability

anastasiia-tilikina:
[DEMO LINK](https://anastasiia-tilikina.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/ToDoList/ToDoList.tsx
+++ b/src/components/ToDoList/ToDoList.tsx
@@ -1,4 +1,9 @@
 import React from 'react';
+import {
+  CSSTransition,
+  TransitionGroup,
+} from 'react-transition-group';
+
 import { Todo } from '../../types/Todo';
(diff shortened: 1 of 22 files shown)


src/components/ToDoList/ToDoList.tsx
```diff
@@ -0,0 +1,63 @@
...
+        ))}
+
+        {tempTodo && (
+          <CSSTransition
+            timeout={300}
+            classNames="temp-item"
+          >
+            <ToDo todo={tempTodo} isTemp />
+          </CSSTransition>
+        )}
```
Virass, line 59: I think we can change this. When i add new todo, animation little bit glitched. Try to realize rendering new todo without temp element. 
Example you can mapping todos and for new todos, you can add boolean status, for example status isLoading. When data update is completed, status can be changed and preloader can be removed

Virass (changes requested):

PR #211, opened 2022-12-15
Completed todo app

vadymbaranov:
[DEMO LINK](https://vadymbaranov.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Loader/Loader.tsx
+++ b/src/components/Loader/Loader.tsx
@@ -8,9 +8,10 @@ type Props = {
 export const Loader: React.FC<Props> = ({ isLoading }) => (
   <div
     data-cy="TodoLoader"
-    className={classNames('modal overlay', {
-      'is-active': isLoading,
-    })}
+    className={classNames('modal overlay',
+      {
+        'is-active': isLoading,
+      })}
   >
     <div className="modal-background has-background-white-ter" />
     <div className="loader" />
--- a/src/components/Loader/Loader.tsx
+++ b/src/components/Loader/Loader.tsx
@@ -1,0 +1,18 @@
+import React from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  isLoading: boolean;
+};
+
+export const Loader: React.FC<Props> = ({ isLoading }) => (
+  <div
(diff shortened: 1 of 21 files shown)


src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,39 @@
...
+  isAdding,
+}) => {
+  const newTodoField = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    // focus the element with `ref={newTodoField}`
+    if (newTodoField.current) {
+      newTodoField.current.focus();
+    }
+  }, []);
```
viktoriia-ovcharenko, line 23: focus does not work

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,18 @@
...
+  <div
+    data-cy="TodoLoader"
+    className={classNames('modal overlay', {
+      'is-active': isLoading,
+    })}
+  >
+    <div className="modal-background has-background-white-ter" />
+    <div className="loader" />
+  </div>
+);
```
viktoriia-ovcharenko, line 18: you have one more "loader" under the list, that work, when you delete or change todo

viktoriia-ovcharenko (approved):

src/components/Loader/Loader.tsx
Largoxoxo: You mean the one that is working when you open the page and wait for todos to load?

Largoxoxo:
Problems Viktoria described originate from addition of setIsAdding on lines 51 and 53 in the App.tsx files. This is causing additional loader at the bottom to be shown every time there is any loading happening from the server. Additionally this causes problems with focus when there are no todos to be loaded. This 'solution' was added to allow todos to appear smoother on the page load (proposed by Mykola in one of the student's works). Without it they just fall down instead of transition.

denys-danyliuk (changes requested):
you should definitely fix this problem with additional loader, I believe you can do it💪

G-YaSha (approved):

PR #212, opened 2022-12-16
Solution v.01

andreyolin:
[DEMO LINK](https://andreyolin.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/AddNewTodoForm/AddNewTodoForm.tsx
+++ b/src/components/AddNewTodoForm/AddNewTodoForm.tsx
@@ -1,0 +1,72 @@
+import React, { useCallback, useEffect, useRef } from 'react';
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
+
+interface Props {
+  user: User | null,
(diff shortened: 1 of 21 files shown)


src/components/AddNewTodoForm/AddNewTodoForm.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+      const userId = user?.id;
+
+      if (title.trim().length === 0) {
+        setErrorText(ErrorText.Title);
+      }
+
+      if (!userId || title.trim().length === 0) {
+        return;
+      }
```
andrii-bassarab, line 46:
```suggestion
      if (title.trim().length) {
        setErrorText(ErrorText.Title);
      }

      if (!userId || title.trim().length) {
        return;
      }
```
I also did it this way, but the logic can be simplified

andrii-bassarab (approved):
Good job!

denys-danyliuk (approved):

PR #219, opened 2022-12-18
solution

andrii-bassarab:
[DEMO LINK](https://andrii-bassarab.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,133 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { Error } from './components/Error';
+import { Filter } from './types/Filter';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { TitleError } from './components/TitleError';
 
 export const App: React.FC = () => {
+  const [hasLoadingError, setHasLoadingError] = useState(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterBy, setFilterBy] = useState(Filter.ALL);
+  const [titleError, setTitleError] = useState(false);
+  const [isAdding, setIsAdding] = useState(false);
+  const [loadingList, setLoadingList] = useState<number[]>([]);
+
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
-  const newTodoField = useRef<HTMLInputElement>(null);
+
+  const getTodosFromServer = async () => {
+    if (user) {
+      try {
+        const todosFromServer = await getTodos(user.id);
+
+        setTodos(todosFromServer);
+      } catch {
+        setHasLoadingError(true);
+        setErrorMessage('Error... Can not load your todos');
+      }
+    }
+  };
 
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
-    if (newTodoField.current) {
-      newTodoField.current.focus();
-    }
+    getTodosFromServer();
   }, []);
+
+  const completedTodos = todos.filter(todo => todo.completed === true);
+  const activeTodos = todos.filter(todo => todo.completed === false);
+
+  let visibleTodos = todos;
+
+  switch (filterBy) {
+    case Filter.COMPLETED:
+      visibleTodos = completedTodos;
+      break;
+
+    case Filter.ACTIVE:
+      visibleTodos = activeTodos;
+      break;
+
+    default: visibleTodos = todos;
+  }
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
-      <div className="todoapp__content">
-        <header className="todoapp__header">
-          <button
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,223 +1,133 @@
...
-              type="button"
-              className="todo__remove"
-              data-cy="TodoDeleteButton"
-            >
-              ×
-            </button>
+        setTodos(todosFromServer);
+      } catch {
+        setHasLoadingError(true);
+        setErrorMessage('Error... Can not load your todos');
```
andreyolin, line 38: `setErrorMessage('Error... Can`t not load your todos');`

src/components/Error.tsx
```diff
@@ -0,0 +1,47 @@
...
+          type="button"
+          className="delete"
+          onClick={clearErrorMessage}
+        />
+
+        Unable to add a todo
+        <br />
+        Unable to delete a todo
+        <br />
+        Unable to update a todo
```
andreyolin, line 43: should all of this text appearing on error?

src/components/TodoInfo.tsx
```diff
@@ -0,0 +1,154 @@
...
+  };
+
+  const handleDoubleClick = (event: React.MouseEvent<HTMLInputElement>) => {
+    if (event.detail === 2) {
+      setIsEditing(true);
+    }
+  };
+
+  const onUpdateTodoTitle = async () => {
+    if (title === newTitle) {
```
andreyolin, line 79: also if newTitle is empty it must be deleted
image

src/App.tsx
```diff
@@ -1,223 +1,133 @@
...
+              onSetTitleError={(isError) => setTitleError(isError)}
+              onSetIsAdding={(isLoading) => setIsAdding(isLoading)}
+              isAdding={isAdding}
+              completedTodos={completedTodos}
+              activeTodos={activeTodos}
+              todos={todos}
+              loadTodos={getTodosFromServer}
+              addTodoToLoadingList={(idToAdd) => setLoadingList(
+                prevIds => [...prevIds, idToAdd],
+              )}
```
andreyolin, line 94: maybe create separate function for this?

src/App.tsx
```diff
@@ -1,223 +1,133 @@
...
+              loadTodos={getTodosFromServer}
+              isAdding={isAdding}
+              title={title}
+              user={user}
+              addTodoToLoadingList={(idToAdd) => setLoadingList(
+                prevIds => [...prevIds, idToAdd],
+              )}
+              deleteTodoOfLoadingList={(idToRemove) => setLoadingList(
+                prevIds => prevIds.filter(id => id !== idToRemove),
+              )}
```
andreyolin, line 109: same here

src/App.tsx
```diff
@@ -1,223 +1,144 @@
...
+      }
     }
+  };
+
+  useEffect(() => {
+    getTodosFromServer();
   }, []);
 
+  const completedTodos = todos.filter(todo => todo.completed);
+  const activeTodos = todos.filter(todo => !todo.completed);
```
denys-danyliuk, line 47: useMemo

src/App.tsx
```diff
@@ -1,223 +1,144 @@
...
+  let visibleTodos = todos;
+
+  switch (filterBy) {
+    case Filter.COMPLETED:
+      visibleTodos = completedTodos;
+      break;
+
+    case Filter.ACTIVE:
+      visibleTodos = activeTodos;
+      break;
```
denys-danyliuk, line 58: there is one more case 'All', you should add it

src/App.tsx
```diff
@@ -1,223 +1,144 @@
...
+
+    default: visibleTodos = todos;
+  }
+
+  const errorInfo = (errorTitle: ErrorTypes) => {
```
denys-danyliuk, line 63: useCallback

src/App.tsx
```diff
@@ -1,223 +1,144 @@
...
-              data-cy="NewTodoField"
-              type="text"
-              ref={newTodoField}
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
+      {errorMessage === ErrorTypes.LOADING
+        ? (
+          <>
+            <span>Error... Can not load your todos</span>
+          </>
```
denys-danyliuk, line 92: do you really need this?

src/components/Error.tsx
```diff
@@ -0,0 +1,38 @@
...
+  hasLoadingError: boolean,
+  errorMessage: string,
+};
+
+export const Error: React.FC<Props> = (
+  { clearErrorMessage, hasLoadingError, errorMessage },
+) => {
+  useEffect(() => {
+    setTimeout(() => clearErrorMessage(), 3000);
+  });
```
denys-danyliuk, line 16: maybe it would be better to add array with deps and add smth in it?

src/components/Header.tsx
```diff
@@ -0,0 +1,109 @@
...
+  return (
+    <header className="todoapp__header">
+      <button
+        data-cy="ToggleAllButton"
+        type="button"
+        className={classNames('todoapp__toggle-all', {
+          active: completedTodos.length === todos.length,
+        })}
+        onClick={toggleAll}
+      />
```
denys-danyliuk, line 98: Знімок екрана 2022-12-21 о 16 25 55
it doesn't work right

denys-danyliuk (changes requested):
look through your code and find where you can use useCallback and useMemo for better optimization, also you should hide this bar if there is no todos
Знімок екрана 2022-12-21 о 15 42 59

src/App.tsx
andrii-bassarab: Yes, that is the idea that if the user has problems loading his todos, then the todos list will not be displayed,
but maybe you're right

Olexandr6 (approved):

PR #222, opened 2022-12-19
 react_todo-app-with-api

MXMUK:
[DEMO LINK](https://MXMUK.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,127 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  FormEvent,
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+
+import classNames from 'classnames';
+
+import { addTodo, deleteTodo, getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { AddTodoFieldForm } from './components/AddTodoFieldForm';
+import { TodoFilters } from './types/TodoFilters';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Error } from './types/Errors';
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+  const [filter, setFilter] = useState<TodoFilters>(TodoFilters.all);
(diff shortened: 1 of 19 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,81 @@
...
+          todo={todo}
+          onDelete={onDelete}
+          onChangeStatus={onChangeStatus}
+          onEditTitle={onEditTitle}
+          isLoading={
+            completedTodosId.includes(todo.id)
+            || deletingTodoId === todo.id
+            || updatingTodoId === todo.id
+            || isUpdating
+          }
```
mitattt, line 44: i guess u should put it in variable

mitattt (approved):
looks good

src/App.tsx
```diff
@@ -1,223 +1,275 @@
...
     }
+
+    loadUsersTodos();
   }, []);
 
+  const visibleTodos = useCallback(() => {
+    const {
+      completed,
+      active,
+      all,
```
denys-danyliuk, line 71: use enum for this

src/App.tsx
```diff
@@ -1,223 +1,275 @@
...
+
+    return todos.filter(todo => {
+      switch (filter) {
+        case completed:
+          return todo.completed;
+        case active:
+          return !todo.completed;
+        case all:
+        default:
+          return todo;
```
denys-danyliuk, line 82:
```suggestion
        case completed:
          return todo.completed;
          
        case active:
          return !todo.completed;
          
        case all:
        default:
          return todo;
```

src/App.tsx
```diff
@@ -1,223 +1,275 @@
...
+            isAdding={isAdding}
+          />
+        </header>
+
+        <section
+          className="todoapp__main"
+          data-cy="TodoList"
+        >
+          <TodoList
+            todos={visibleTodos()}
```
denys-danyliuk, line 245: it's better to save this in to some variable

src/components/TodoList/TodoList.tsx
denys-danyliuk: agree

denys-danyliuk (changes requested):
you should hide this bar when there is no todos, check the working version
Знімок екрана 2022-12-21 о 15 42 59

src/App.tsx
```diff
@@ -1,223 +1,282 @@
...
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
+
+    loadUsersTodos();
   }, []);
 
+  const processedTodos = useCallback(() => {
```
denys-danyliuk, line 67: useMemo

denys-danyliuk (approved):

PR #223, opened 2022-12-19
3rd part done

Esceype:
- [DEMO LINK](https://Esceype.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,154 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useState,
+  useCallback,
+  useMemo,
+} from 'react';
+import classNames from 'classnames';
 import { AuthContext } from './components/Auth/AuthContext';
+import {
+  getTodos, addTodo, deleteTodo, updateTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { NewTodoField } from './components/Auth/NewTodoField/NewTodoField';
+import { TodoList } from './components/Auth/TodoList';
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,223 +1,223 @@
...
+    async (event: React.FormEvent<HTMLFormElement>) => {
+      event.preventDefault();
+
+      if (!title) {
+        setError('Title can\'t be empty');
+
+        return;
+      }
+
+      setIsAdding(true);
```
Kry-Gin, line 68: You set isAdding here as true, but if program catch an error (for examle after set in title several spaces '   '), it will disable button forever. You should add isAdding false if program catch an error

src/App.tsx
```diff
@@ -1,223 +1,223 @@
...
+
+        case Status.All:
+        default:
+          return true;
+      }
+    })
+  ), [status, todos]);
+
+  const completedTodos = todos.filter(todo => todo.completed);
+  const activeTodos = todos.filter(todo => !todo.completed);
```
Kry-Gin, line 118: useMemo

src/App.tsx
```diff
@@ -1,223 +1,223 @@
...
+              <span className="todo-count" data-cy="todosCounter">
+                {`${activeTodos.length} items left`}
+              </span>
+
+              <Filter status={status} onStatusChange={setStatus} />
+
+              <button
+                data-cy="ClearCompletedButton"
+                type="button"
+                disabled={completedTodos.length === 0}
```
Kry-Gin, line 202:
```suggestion
                disabled={!completedTodos.length}
```

Kry-Gin (changes requested):
Great work, overall 🚀
Let's make some correction

PR #226, opened 2022-12-19
v1 all

sergosdv1:
[DEMO LINK](https://DanilSerzh.github.io/react_todo-app-with-api/)


kryvets-vadym (approved):
GJ! But your loader doesn't work correctly when adding and removing todos. Text inside the loader must be taken from the new todo

illiarerun (approved):

PR #228, opened 2022-12-20
implemented form

Artem66:
[DEMO LINK](https://Artem66.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,0 +1,44 @@
+import React from 'react';
+
+type Props = {
+  newTodoField: React.RefObject<HTMLInputElement>,
+  handleSubmit: (event: React.FormEvent) => void,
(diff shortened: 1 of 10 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+      <form onSubmit={handleSubmit}>
+        <input
+          data-cy="NewTodoField"
+          type="text"
+          ref={newTodoField}
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          value={todoTitle}
+          onChange={onChange}
```
taraselnik, line 38: maybe better to make local state. it will help to avoid rerendereng when you text in input.

taraselnik (approved):
GJ. but better try to clear unnecessary comments

PR #233, opened 2022-12-26
Develop

Anastasiia-Smirnova:
[DEMO LINK](https://Anastasiia-Smirnova.github.io/react_todo-app-with-api/)


denys-danyliuk (changes requested):
when I change todo changes should save on blur, also esc should unfocus the field
when I want to toggle all todos loaders should not appear on todos that are already complete
overall performance when I pick 'active' or 'completed' is slow

PR #235, opened 2022-12-27
Done

ruslik87:
[DEMO LINK](https://ruslik87.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,0 +1,179 @@
+import React, { useEffect, useState, useRef } from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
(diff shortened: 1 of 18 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,175 @@
...
+      if (keyCode !== 'Escape') {
+        return;
+      }
+
+      setIsRenaming({});
+    };
+
+    return (
+      <>
+        {todos.map(todo => (
```
Virass, line 58: Move this logic to `todoItem` component

Virass (changes requested):
Nice work, but take the logic of the list elements into a separate component

nazarmatsevych (approved):

PR #246, opened 2023-01-11
Todo app with api task solution

juliiaap:
- [DEMO LINK](https://juliiaap.github.io/react_todo-app-with-api/)


qoa11a (changes requested):
Good job!
One fix need to be done:
When I check two todo items fast and consequently only one of them is checked on frontend side (if we refresh the page we see that actually both were checked).

qoa11a (approved):

PR #247, opened 2023-01-11
for mentor's review

Anastasiya145:
as we discussed during the QA call, I do not understand why the method onChangeStatus has a mistake with TS validation.
Please help me

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3494,9 +3494,9 @@
       }
     },
     "@mate-academy/scripts": {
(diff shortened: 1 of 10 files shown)


TarasHoliuk:
Hi! I can't find the function `onChangeStatus`. Please attach a screenshot with the error etc. But to get response faster better do it in fe_chat

TarasHoliuk (changes requested):
Please ask in chat if you need help
I don't have info about described error

PR #262, opened 2023-01-24
solution

fallowlone:
[DEMO LINK](https://official-artem.github.io/react_todo-app-with-api/)


Ihor-Karpyn (approved):
- remove loader on add todo or fix it. There is should be one loader on add

Good job

PR #265, opened 2023-01-25
react_todo-app-with-api

ysrckr:
[DEMO LINK](https://ysrckr.github.io/react_todo-app-with-api/)


src/components/Todo/TodoFooter.tsx
```diff
@@ -0,0 +1,62 @@
...
+        data-cy="todosCounter"
+      >
+        {todoLength > 0 && `${todoLength} items left`}
+      </span>
+
+      <nav
+        className="filter"
+        data-cy="Filter"
+      >
+        {filterStatusList.map(status => (
```
lefteris-L, line 36: I think here you can just create each link instead of map, it's only 3 options. This way you don't need extra variables, functions, types etc.

src/components/Todo/TodoHeader.tsx
```diff
@@ -0,0 +1,102 @@
...
+    }
+  };
+
+  const toggleAll = () => {
+    if (!isAllTodosCompleted) {
+      return;
+    }
+
+    toggleAllTodos();
+  };
```
lefteris-L, line 68: You should be able to toggle in any case. If all are completed, make them active and if at least 1 is active, complete all active todos

src/App.tsx
```diff
@@ -1,223 +1,26 @@
...
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
-      <div className="todoapp__content">
-        <header className="todoapp__header">
-          <button
-            data-cy="ToggleAllButton"
-            type="button"
-            className="todoapp__toggle-all active"
```
lefteris-L, line 26: this should have 'active' when all are completed

lefteris-L (approved):
When I add todos they stay in loading state until I refresh and then they're ok. I'm a little lost when looking for a function/variable but everything else looks great, well done 👍

src/components/Todo/TodoHeader.tsx
ysrckr: I thought it works one way because we are making the button inactive if not all completed

lefteris-L: Yeah but it works this way in the working version, so maybe that's the correct way. Maybe @naziamoff will answer this for us

src/components/Todo/TodoHeader.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  useEffect(() => {
+    // focus the element with `ref={newTodoField}`
+    if (newTodoField.current) {
+      newTodoField.current.focus();
+    }
+  }, []);
```
meforoma, line 75: this doesn't work - after you submit new todo, input field is not focused

meforoma: deps array is empty - hence it happens only once - on componentDidMount
so you have to add a dependency which will trigger the .focus on it's change

meforoma (changes requested):
toggleAll todos should work as in the example

https://user-images.githubusercontent.com/39308717/215431458-ae7da260-fb07-4e86-b033-35f8634d44e1.mp4

and see another comment in the code

meforoma (approved):
👏🏼

PR #266, opened 2023-01-26
Solution

UmizDemud:
Description: Add toggle button functionalites, implement editing functionality with contenteditable span


nurdanokcu (changes requested):
I have a few points to mention considering the task requirements:
-App shouldn't allow to add new task if the title is empty
-It should show loader while adding a new Todo
-While toggling, there should be a loader on all the items that are being toggled
-If there are no completed task, clear completed button shouldn't be visible.
-The number that counts how many items left should be a number of uncompleted tasks not all the tasks.
-You chose an approach not to load completed todos, 
but when the task is completed it stays on the page until page is reloaded. 
User won't know that completed tasks are being deleted until the page refresh. 
It is a good idea to inform the user about it at least or remove the completed todos from the page directly 
and if user wants to keep track of what they achieved in the past  or they want to repeat the same task every day by just toggling it, they wont have that option.
Let me know what you think.

UmizDemud:
@nurdanokcu 
I had the updates ready, pushed them just now.

I didn't think about this comment "or they want to repeat the same task every day by just toggling" before. And makes sense. I guess I'd have a module for repeating tasks and I'd add them each day upon click. I guess the user needs to give it a go once to understand the remaining details. If you wouldn't mind, I like this implementation more (Only if the server deletes completed todos periodically). I'd like to keep that detail

nurdanokcu:
> @nurdanokcu I had the updates ready, pushed them just now.
> 
> I didn't think about this comment "or they want to repeat the same task every day by just toggling" before. And makes sense. I guess I'd have a module for repeating tasks and I'd add them each day upon click. I guess the user needs to give it a go once to understand the remaining details. If you wouldn't mind, I like this implementation more (Only if the server deletes completed todos periodically). I'd like to keep that detail

Yeah usually it would be a bigger page with recurring tasks section where they want to track habits.
 It is about what type of page you or the customer wants to implement . 
 Sure sure you can keep the detail the way you want. I just wanted to get your opinion on it.

src/components/Todos/TodosFooter.tsx
```diff
@@ -0,0 +1,74 @@
...
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          data-cy="FilterLinkAll"
+          href="#/"
+          className={`filter__link ${
+            filterType === 'all' ? 'selected'
+              : ''} ${
+            !total ? 'is-disabled' : ''}`}
```
meforoma, line 29: i still feel weird about ternaries in classnames 🙂 
frankly they're hard to comprehend

src/components/Todos/TodosList.tsx
```diff
@@ -0,0 +1,52 @@
...
+      case 'all':
+        return true;
+      case 'completed':
+        return item.completed;
+      case 'active':
+        return !item.completed;
+      default:
+        return true;
+    }
+  });
```
meforoma, line 36:
```suggestion
  const filteredTodos = todos.filter(({completed}) => {
    switch (filterType) {
      case 'all':
        return true;
      case 'completed':
        return completed;
      case 'active':
        return !completed;
      default:
        return true;
    }
  });
```

meforoma (approved):
very nice
although i'm still confused with those ternary classnames 🙂

PR #269, opened 2023-01-26
react_todo-app-with-api

kolya-movchan:
[DEMO LINK](https://kolya-movchan.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,20 +1,28 @@
 /* eslint-disable no-console */
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import classNames from 'classnames';
+
 import React, {
-  useContext, useEffect, useRef, useState,
+  useContext,
+  useEffect,
+  useState,
 } from 'react';
 
 import {
-  getTodos, postTodos, deleteTodos, patchTodos,
+  getTodos,
+  postTodos,
+  deleteTodos,
+  patchTodos,
 } from './api/todos';
+
 import { AuthContext } from './components/Auth/AuthContext';
-import { Filter } from './components/React/Filter';
+import { Filter, Filters } from './components/React/Filter';
 import { NewTodo } from './components/React/NewTodo';
+
 import { TodoList } from './components/React/TodoList';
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,222 +1,302 @@
...
-  const newTodoField = useRef<HTMLInputElement>(null);
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<Filters>(Filters.All);
+  const [error, setError] = useState('');
+  const [input, setInput] = useState('');
+  const [isAdding, setIsAdding] = useState(false);
+  const [todoOnload, setTodoOnLoad] = useState<Todo | null>(null);
+  const [todoIdsOnLoad, setTodoIdsOnLoad] = useState<number[]>([]);
+  const [toggle, setToggle] = useState(false);
```
vlad-demchuk, line 41:
```suggestion
  const [isActiveToggle, setIsActiveToggle] = useState(false);
```

src/App.tsx
```diff
@@ -1,222 +1,302 @@
...
+    }
+
+    try {
+      putTodoOnLoad(id);
+
+      await patchTodos(
+        id,
+        typeof data === 'string'
+          ? { title: data }
+          : { completed: !data },
```
vlad-demchuk, line 86: better to create a separate variable instead of ternary operator in function parameters

src/components/React/Filter.tsx
```diff
@@ -0,0 +1,82 @@
...
+        selected: filter === filterName,
+      },
+    );
+  };
+
+  const activeTodos = todos.filter(todo => !todo.completed).length;
+  const isCompletedTodo = todos.some(todo => todo.completed);
+
+  return (
+    <>
```
vlad-demchuk, line 36: you don't need this react fragment

vlad-demchuk (changes requested):
Looks good!

tetiana-ivanchenko (approved):

PR #270, opened 2023-01-26
Solution

4ebupel:
[DEMO LINK](https://4ebupel.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,38 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+  useCallback,
+  useMemo,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
 
+import {
+  addTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+} from './api/todos';
+
+import { Todo } from './types/Todo';
+import { FilterType } from './types/FilterType';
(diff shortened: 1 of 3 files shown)


nikachu404 (changes requested):
Well done! 🚀  But I noticed that when you update or toggle todo, it goes into the `selectedTodoIds`, and is not removed from there when the interaction ends, so when you toggle or update another todo, it is loading with the previous todo. So I suggest to add `setSelectedTodoIds([])` to `editTodo` function at `finally` block

4ebupel:
> Well done! 🚀 But I noticed that when you update or toggle todo, it goes into the `selectedTodoIds`, and is not removed from there when the interaction ends, so when you toggle or update another todo, it is loading with the previous todo. So I suggest to add `setSelectedTodoIds([])` to `editTodo` function at `finally` block

Thanks! That's a good point, I'll fix it as soon as possible)

nikachu404 (approved):

src/App.tsx
```diff
@@ -14,210 +44,224 @@ export const App: React.FC = () => {
...
+      case FilterType.All:
+        return todos;
+
+      case FilterType.Active:
+        return todos.filter(todo => !todo.completed);
+
+      case FilterType.Completed:
+        return todos.filter(todo => todo.completed);
+      default:
+        throw new Error('Invalid type');
```
ivan-kadykalo, line 211: Use enum for error messages

ivan-kadykalo (changes requested):
There are some strange margin-left on todo edit
image

ivan-kadykalo (approved):
Good job 🔥

PR #271, opened 2023-01-27
solution

Sveta-Kryukova:
[DEMO LINK](https://Sveta-Kryukova.github.io/react_todo-app-with-api/)


Ihor-Karpyn (changes requested):
demo link

PR #287, opened 2023-01-30
Change title method implemented full, small fixes

Lun4trick:
[DEMO](https://lun4trick.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,0 +1,141 @@
+import cn from 'classnames';
+import {
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo
+  onUpdate: (todo: Todo, prop: string, value: any) => void
+  onDelete: (todoId: number) => void
+  changingTodoIds: number | number[]
+};
+
+export const TodoInfo: React.FC<Props> = ({
+  todo,
+  onUpdate,
+  onDelete,
+  changingTodoIds,
+}) => {
+  const [newTitle, setNewTitle] = useState(todo.title);
+  const [isEditing, setIsEditing] = useState(false);
+  const newTodoField = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (newTodoField.current) {
+      newTodoField.current.focus();
+    }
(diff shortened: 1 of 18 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,141 @@
...
+            }}
+          >
+            <input
+              data-cy="NewTodoField"
+              type="text"
+              key={todo.id}
+              value={newTitle}
+              onDoubleClick={() => {
+                setIsEditing(true);
+              }}
```
nurdanokcu, line 82: I guess this double click isn't needed , this is the input while the form is open to edit already. If this input is being displayed it means isEditing state is true already.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,141 @@
...
+              className="todo__title-field"
+              placeholder="What needs to be done?"
+              disabled={!isEditing}
+            />
+          </form>
+        )
+        : (
+          <>
+            <span
+              key={todo.id}
```
nurdanokcu, line 103: I couldn't understand why there is a key here. Todoinfo components is being assigned a key in the todolist while mapping.

nurdanokcu (approved):
Nicely done.

src/components/TodoInfo/TodoInfo.tsx
meforoma: ++ agree. DoubleClick is essential for the `<span` but for input - no need

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,141 @@
...
+    setIsEditing(false);
+  };
+
+  return (
+    <div
+      data-cy="Todo"
+      className={cn('todo', {
+        completed: todo.completed,
+      })}
+      key={todo.id}
```
meforoma, line 48: you're dropping too much `key` attributes
you don't seem to even need any here
you only need to provide it when `.map` in JSX
also, you're using the same (hence, NOT UNIQUE) `key` few times - which is wrong
key attribute must be unique

reconsider it's usage throughout your code

meforoma (changes requested):
few minor things to fix
other than that - well done 👍🏼

meforoma (approved):

PR #290, opened 2023-01-30
Solution

OlhaArama:
[DEMO LINK](https://OlhaArama.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,203 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext, useEffect, useState, useMemo,
+} from 'react';
+import { getTodos, removeTodo, updateTodo } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,223 +1,203 @@
...
+    if (user) {
+      getTodos(user.id)
+        .then(setTodos)
+        .catch(() => {
+          setIsError(true);
+        });
     }
   }, []);
 
+  const addTodo = (value: Todo) => {
```
vlad-demchuk, line 38: what is `value`? use better variable names

vlad-demchuk: check all such cases

vlad-demchuk (approved):

src/components/TodoDetails.tsx
```diff
@@ -0,0 +1,161 @@
...
+      newTodoField.current.focus();
+    }
+  });
+
+  const isDeleted = (todoId: number) => {
+    return deletedTodoIds?.includes(todoId);
+  };
+
+  const isLoadingTodo = (todoId: number) => {
+    if (deletedTodoIds.length !== 0) {
```
TarasHoliuk, line 38: I guess you can handle `loading` circle rendering with only one state property, like `loadingTodoIds` - and push/remove (at the start/end of action) there ids on any action (deleting/adding/editing), since you can do only one action at the time.

It'd simplify the app.
But it's just FYI - anyway - GJ :+1:

TarasHoliuk (approved):

PR #293, opened 2023-01-30
soution

victoria-kovalenko:
[DEMO LINK](https://victoria-kovalenko.github.io/react_todo-app-with-api/)


llllewvvaa (changes requested):
It seems ok, but you don't have logic for `ToggleAll` button, which changes todos status to opposite

PR #306, opened 2023-01-31
Todo info component was added, toggle and renaming was implemented

vrima1221:
[DEMO LINK](https://vrima1221.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,0 +1,131 @@
+import React, { useState, useEffect, useRef } from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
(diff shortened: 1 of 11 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,131 @@
...
+        completed: todo.completed,
+      })}
+    >
+      <label className="todo__status-label">
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          defaultChecked={todo.completed}
+          onClick={() => onTodoUpdate(todo)}
```
OlggaJul, line 89: i note that all tasks have loader while i change status for one of them

OlggaJul (changes requested):

PR #311, opened 2023-02-01
solution

kreisberg:


Sveta-Kryukova (commented):
Good job) But there are some problems) your button that shoul toggle all woks not exactly in that way that it should. It just make todos that completed to active, and thisc button woks only once( also it should with some kind of active class when it is cliked) also, I cant rename todos( and they are not deleting if I left empty string when renaming) Also add a placeholder for the case user shoul know if he wants rename todo in an empty string it should be deleted)

Sveta-Kryukova (changes requested):
.

kreisberg:
> Good job) But there are some problems) your button that shoul toggle all woks not exactly in that way that it should. It just make todos that completed to active, and thisc button woks only once( also it should with some kind of active class when it is cliked) also, I cant rename todos( and they are not deleting if I left empty string when renaming) Also add a placeholder for the case user shoul know if he wants rename todo in an empty string it should be deleted)

Everything was fixed - now u can update todos and toggle filter is working ☑️🦦

src/api/todos.ts
```diff
@@ -1,8 +1,29 @@
...
+) => {
+  return client.patch<Todo>(`/todos/${todoId}`, updateData);
+};
+
+export const todoApi = {
+  getTodos,
+  addTodo,
+  deleteTodo,
+  updateTodo,
+};
```
valerii-derkach, line 29: 🔥

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,85 @@
...
+    if (!user) {
+      showError('User not found');
+
+      return;
+    }
+
+    try {
+      await onAddTodo({
+        title,
+        userId: user?.id,
```
valerii-derkach, line 45:
```suggestion
        userId: user.id,
```

src/helpers/helpers.tsx
```diff
@@ -0,0 +1,23 @@
...
+) => {
+  if (filter === Filters.ALL) {
+    return todos;
+  }
+
+  return todos.filter((todo) => (filter === Filters.COMPLETED
+    ? todo.completed
+    : !todo.completed
+  ));
+};
```
valerii-derkach, line 16: what if we pass something else, not ACTIVE? handle all cases (use switch case)

valerii-derkach (changes requested):
check title editing - does it work as you expect?

src/components/Header/Header.tsx
kreisberg: it was TS fixes 🦦

src/api/todos.ts
kreisberg: fix updating ☑️

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,103 @@
...
+                className="todo__title"
+                onDoubleClick={() => setShouldShowInput(true)}
+              >
+                {todo.title}
+              </span>
+              <button
+                type="button"
+                className="todo__remove"
+                data-cy="TodoDeleteButton"
+                onClick={() => onDeleteTodo(todo.id)}
```
ivan-kadykalo, line 82: I believe using `on...` in the names is not the best idea. So, this name won't describe the action which will be done.
On the other hand, if I'll write `deleteTodoById`, will you understand what will happen and which data you should pass?

```suggestion
                onClick={() => deleteTodoById(todo.id)}
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+      <a
+        data-cy="FilterLinkCompleted"
+        href="#/completed"
+        className={cn(
+          'filter__link', {
+            selected: filter === Filters.COMPLETED,
+          },
+        )}
+        onClick={() => onChange(Filters.COMPLETED)}
```
ivan-kadykalo, line 63: The same as in the previous comment
```suggestion
        onClick={() => setFilterStatus(Filters.COMPLETED)}
```

ivan-kadykalo (approved):

PR #318, opened 2023-02-04
solution

AnnyKoval:
[DEMO LINK](https://AnnyKoval.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,171 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
+import { deleteTodo, getTodos, updateTodo } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { ErrorMessage } from './components/ErrorMessage';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { ErrorType } from './types/ErrorType';
+import { FilterStatus } from './types/FilterStatus';
+import { TempTodo, Todo } from './types/Todo';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
-  const newTodoField = useRef<HTMLInputElement>(null);
+  const [error, setError] = useState(ErrorType.None);
+  const [filterStatus, setFilterStatus] = useState(FilterStatus.All);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [updatingTodoIds, setUpdatingTodoIds] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<TempTodo | null>(null);
+  const [isAllCompleted, setIsAllCompleted] = useState(false);
+
+  const getTodosFromServer = async () => {
+    if (!user) {
+      return;
+    }
+
+    try {
+      const receivedTodos = await getTodos(user.id);
+
+      setTodos(receivedTodos);
+    } catch (err) {
+      setError(ErrorType.LoadingError);
+    }
+  };
 
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
-    if (newTodoField.current) {
-      newTodoField.current.focus();
-    }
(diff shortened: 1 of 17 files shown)


src/types/Todo.ts
```diff
@@ -4,3 +4,10 @@ export interface Todo {
   title: string;
   completed: boolean;
 }
+
+export type TempTodo = {
```
TarasHoliuk, line 8: Not sure you need this type

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,56 @@
...
+        return !todo.completed;
+      case FilterStatus.Completed:
+        return todo.completed;
+      default:
+        return todo;
+    }
+  });
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
```
TarasHoliuk, line 37: it's better to use semantic tags for a list, i.e `ul`, `li`

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const TodoList:React.FC<Props> = ({
+  todos,
+  filterStatus,
+  tempTodo,
+  handleDelete,
+  handleStatusChange,
+  updatingTodoIds,
+  editTodo,
+}) => {
+  const visibleTodos = todos.filter(todo => {
```
TarasHoliuk, line 25: better to do it in `App.tsx` and pass `visibleTodos` as `todos` to the `TodoList` component
In this case, `TodoList` won't know anything about the filter

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,115 @@
...
+};
+export const TodoInfo:React.FC<Props> = React.memo(({
+  onDelete,
+  todo,
+  handleStatusChange,
+  updatingTodo,
+  editTodo,
+}) => {
+  // eslint-disable-next-line no-console
+  console.log('rendering todoinfo');
```
TarasHoliuk, line 20: delete it

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,115 @@
...
+        { completed: todo.completed })}
+    >
+      <label className="todo__status-label">
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+          onChange={() => {
+            if (handleStatusChange) {
```
TarasHoliuk, line 51: I know that's because of `TempTodo`, but I guess it's better to create a function like this `const doNothing = () => {};` above component and use it as a default value for all functions you need instead of checking it every time

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,115 @@
...
+      </label>
+
+      {isEditing ? (
+        <form
+          onSubmit={(event) => {
+            event.preventDefault();
+            handleEdit();
+          }}
+        >
+          <input
```
TarasHoliuk, line 65: it should be focused when appears

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,111 @@
...
+      >
+        <input
+          data-cy="NewTodoField"
+          type="text"
+          ref={newTodoField}
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          disabled={isLoading}
+          value={newTitle}
+          onChange={(event) => setNewTitle(event.target.value)}
```
TarasHoliuk, line 106: move it `handleChange` as well

TarasHoliuk (changes requested):
Very good job, but let's improve it a bit :)

src/App.tsx
```diff
@@ -1,16 +1,202 @@
...
+      event.preventDefault();
+
+      if (!title.trim()) {
+        setErrorMessage(ErrorMessage.TitleCantBeEmpty);
+
+        return;
+      }
+
+      const addNewTodo = async () => {
+        setIsAdding(true);
```
OlEzhA0, line 83: it's good way to separate logic of adding todo and validation (above)
but would better to keep it outside handleSubmit and wrap with useCallback

OlEzhA0 (changes requested):
check this bug with empty todo
Screenshot 2023-02-06 at 11 55 08 AM

qoa11a (approved):

PR #323, opened 2023-02-08
Develop

lukacsja:
[DEMO LINK](https://lukacsja.github.io/react_todo-app-with-api/)

Everything looks to work as expected except that on doubleclick I need to click one more time to be able to edit the field. 
I'm open to any suggestions for that and everything else.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,7 +19,8 @@ import {
   getTodos,
   addTodo,
   removeTodo,
-  updateTodo,
+  updateTodoStatus,
+  updateTodoTitle,
 } from './api/todos';
 
 const USER_ID = 6160;
@@ -27,7 +28,6 @@ const USER_ID = 6160;
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [error, setError] = useState<Errors>(Errors.NoError);
-  const [isEditing] = useState(false);
   const [filteredTodos, setFilteredTodos] = useState<Todo[]>(todos);
   const [selectedFilter, setSelectedFilter] = useState<Filters>(Filters.All);
   const [newTodoTitle, setNewTodoTitle] = useState('');
@@ -38,8 +38,29 @@ export const App: React.FC = () => {
   const activeTodos = todos.filter(todo => !todo.completed);
   const completedTodos = todos.filter(todo => todo.completed);
 
+  const saveEditedTitle = (
+    todoId: number,
+    todoTitle: string,
+    setEditingTodoId: (value: number) => void,
+  ) => {
+    setLoadingTodoIds([...loadingTodoIds, todoId]);
+    updateTodoTitle(todoId, todoTitle)
+      .then(() => {
+        getTodos(USER_ID)
+          .then(todosFromServer => {
+            setTodos(todosFromServer);
+            setFilteredTodos(todosFromServer);
+            setLoadingTodoIds([]);
+            setEditingTodoId(0);
+          })
+          .catch(() => {
+            setError(Errors.CantUpdate);
+          });
+      });
+  };
+
   const toggleTodoStatus = (todoId: number, completed: boolean) => {
-    updateTodo(todoId, !completed)
+    updateTodoStatus(todoId, !completed)
       .then(() => {
         setLoadingTodoIds([...loadingTodoIds, todoId]);
         getTodos(USER_ID)
@@ -56,11 +77,11 @@ export const App: React.FC = () => {
 
   const toggleAllTodoStatus = () => {
     const everyTodoCompleted = todos.every(todo => todo.completed);
+    const noTodoCompleted = todos.every(todo => !todo.completed);
     const someTodosCompleted = todos.some(todo => todo.completed)
       && !everyTodoCompleted;
-    const noTodosCompleted = todos.every(todo => !todo.completed);
-    const updatedTodos: Todo[] = [];
 
+    const updatedTodos: Todo[] = [];
     const loadingTodoIdsToAdd: number[] = [];
 
     todos.forEach(todo => {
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,24 +1,273 @@
...
+            setFilteredTodos(todosFromServer);
+            setLoadingTodoIds([]);
+          })
+          .catch(() => {
+            setError(Errors.CantUpdate);
+          });
+      });
+  };
+
+  const toggleAllTodoStatus = () => {
```
naziamoff, line 78: very huge method, try to divide it into the smaller ones

src/components/Footer.tsx
```diff
@@ -0,0 +1,81 @@
...
+  completedTodos,
+  filterByAll,
+  filterByActive,
+  filterByCompleted,
+  selectedFilter,
+  clearCompletedTodos,
+}) => {
+  const hasFilterAll = selectedFilter === Filters.All;
+  const hasFilterActive = selectedFilter === Filters.Active;
+  const hasFilterCompleted = selectedFilter === Filters.Completed;
```
naziamoff, line 26: redundant variables, you can do it right in the classnames

src/components/Main.tsx
```diff
@@ -0,0 +1,155 @@
...
+                  type="text"
+                  className="todo__title-field"
+                  placeholder="Empty todo will be deleted"
+                  value={newTitle}
+                  onChange={changeEditField}
+                  onBlur={(event) => {
+                    event.preventDefault();
+                    handleEditSubmit(todo.id, todo.title);
+                    saveEditedTitle(todo.id, newTitle, setEditingTodoId);
+                  }}
```
naziamoff, line 84: worth a handler

src/components/Main.tsx
```diff
@@ -0,0 +1,155 @@
...
+          />
+        </label>
+
+        <div className="modal overlay">
+          <div className="modal-background has-background-white-ter" />
+          <div className="loader" />
+        </div>
+      </div> */}
+
+      {/* This todo is in loadind state */}
```
naziamoff, line 136: do you need those comments? Seems it is the last part of the task

naziamoff (changes requested):
great job, Luka 🔥 
Let's polish and you're good

lukacsja:
> about your question - there's a method for focusing on the element. Try to google, and if you won't succeed in it - let me know and I'll help

thank you. finally managed to implement it. also it doesnt save when the title doesnt change now.

src/App.tsx
lukacsja: i broke it into 4 functions. not sure if its more readable tho

src/App.tsx
```diff
@@ -1,24 +1,311 @@
...
+      });
+  };
+
+  const getUpdatedTodos = (
+    todosToUpdate: Todo[],
+    everyTodoCompleted: boolean,
+    noTodoCompleted: boolean,
+    someTodosCompleted: boolean,
+  ) => {
+    const updatedTodos: Todo[] = [];
```
meforoma, line 84: you can provide an initial value in args area
```suggestion
    todosToUpdate: Todo[] = [],
    everyTodoCompleted: boolean,
    noTodoCompleted: boolean,
    someTodosCompleted: boolean,
  ) => {
```

src/App.tsx
```diff
@@ -1,24 +1,311 @@
...
+    todosToUpdate.forEach(todo => {
+      let updatedTodo = { ...todo };
+
+      if (someTodosCompleted && !todo.completed) {
+        updatedTodo = { ...todo, completed: !todo.completed };
+      }
+
+      if (everyTodoCompleted || noTodoCompleted) {
+        updatedTodo = { ...todo, completed: !todo.completed };
+      }
```
meforoma, line 95: could be shorter - update props when constructing updatedTodo
```suggestion
      const completed = (someTodosCompleted && !todo.completed)
        || (everyTodoCompleted || noTodoCompleted)
      const updatedTodo = {
        ...todo,
        completed,
      };
```

src/App.tsx
```diff
@@ -1,24 +1,311 @@
...
+    const loadingTodoIdsToAdd: number[] = [];
+
+    todosToLoad.forEach(todo => {
+      if (someTodosCompleted && !todo.completed) {
+        loadingTodoIdsToAdd.push(todo.id);
+      }
+
+      if (everyTodoCompleted || noTodoCompleted) {
+        loadingTodoIdsToAdd.push(todo.id);
+      }
```
meforoma, line 118: you can combine evaluations like in prev comment
```suggestion
      if ((someTodosCompleted && !todo.completed)
        || (everyTodoCompleted || noTodoCompleted)) {
        loadingTodoIdsToAdd.push(todo.id);
      }
```

src/components/Main.tsx
```diff
@@ -0,0 +1,146 @@
...
+                  onChange={changeEditField}
+                  onBlur={(event) => {
+                    event.preventDefault();
+                    handleEditSubmit(todo.id, todo.title);
+                  }}
+                  onKeyDown={(event) => {
+                    if (event.key === 'Escape') {
+                      setEditingTodoId(0);
+                    }
+                  }}
```
meforoma, line 95: let's keep it as short as possible
```suggestion
                  onBlur={handleBlur}
                  onKeyDown={handleEscape}
```

meforoma (changes requested):
let's address few comments and move on

src/App.tsx
lukacsja: thank you for your comments, i created a nice and common handlerFunction for both 'blur' and ' form submit' but with this one i had hard times shortening it.
the above described case updates all of them even when there are some completed and some active todos so i think i cant go for it. 
or i am just missing something

meforoma: oh, so you mean that `everyTodoCompleted || noTodoCompleted` are dependent from `someTodosCompleted && !todo.completed`?
if so, you could carry on.

I think you should just try it and test out edge-cases that should be handled by the logic.

Have you tried to apply the suggestion? And test it on the rendered page?

lukacsja: yes of course but then i realized that i probably need either switch or if statements because sometimes on only want to update some of them (both on the page and in the api)

meforoma (approved):
nice 👍🏼

PR #327, opened 2023-02-12
Develop

SalehAkaJim:
Demo in here: [DEMO LINK](https://SalehBUD.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,24 +1,158 @@
...
+  };
+
+  const updateTodo = async (todo: Todo, update: 'title' | 'complete') => {
+    const data
+      = update === 'title' ? todo : { ...todo, completed: !todo.completed };
+
+    await patchTodo(data)
+      .then((res) => {
+        setFilteredTodos(
+          filteredTodos.map((t) => {
```
Grayfox147, line 80:
```suggestion
          (filteredTodos) => filteredTodos.map((t) => {
```
I would recommend you to use the current filteredTodos value instead of reading the variable again.

src/App.tsx
```diff
@@ -1,24 +1,158 @@
...
+      })
+      .catch(() => {
+        setError(ErrorMessages.addTodo);
+      });
+
+    setIsInputDisabled(false);
+  };
+
+  const removeTodo = async (id: number) => {
+    await deleteTodo(id)
```
Grayfox147, line 64:
```suggestion
 setTodosToDelete(current => [...current, todoId]); 
    await deleteTodo(id)
```

for you to fix the delete completed issue create a state and a setter of the state. when deleting send every ID to this state wich it would be an array of todosIds

src/App.tsx
```diff
@@ -1,24 +1,158 @@
...
+      .then(() => {
+        setFilteredTodos(filteredTodos.filter((todo) => todo.id !== id));
+      })
+      .catch(() => {
+        setError(ErrorMessages.deleteTodo);
+      });
+  };
```
Grayfox147, line 71:
```suggestion
.finally(() => {
 setTodosToDelete.filter((id) => id !== todoId),
})
  };
```
and at the end you remove every id from it, this would allow you to render only when all the todos are deleted.

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,69 @@
...
+        >
+          Completed
+        </a>
+      </nav>
+
+      {hasIncompleteTodos && (
+        <button
+          type="button"
+          className="todoapp__clear-completed"
+          onClick={clearCompleted}
```
Grayfox147, line 62:
```suggestion
          onClick={clearCompleted}
           style={{
        visibility: completedTodos?.length
          ? ('visible')
          : ('hidden'),
      }}
```
here you should check if you have completed todos, if you do the button will be visible otherwise it will be hidden.

Grayfox147 (approved):
Good implementation, although it needs some changes for it to be able to work as expected, really good job!

meforoma (changes requested):
@SalehBUD @Grayfox147 made really good point - please implement them
and deploy the most recent version after you implement changes

also you could implement .focus on the form input field after successful addTodo

meforoma (changes requested):
@SalehBUD we've got a very weird behaviour - checkout the video

https://user-images.githubusercontent.com/39308717/219297640-ac852b48-5b3f-4680-bbdf-9bc931723f7e.mov

meforoma (changes requested):
hmm, still can't approve it @SalehBUD 

https://user-images.githubusercontent.com/39308717/219338397-93503b63-1c0b-4d9b-847a-816660fae3eb.mov

SalehBUD:
> hmm, still can't approve it @SalehBUD

😅 Fixed

src/components/TodoContent/TodoContent.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+      <TodoMain
+        todos={todos}
+        tempTodo={tempTodo}
+        deleteTodo={deleteTodo}
+        updateTodo={updateTodo}
+        loadingAll={loadingAll}
+      />
+
+      {todos && (
```
meforoma, line 74: `todos` is an array, right?
`[]` is not a falsy equivalent, even if empty
BUT `todos.length` is a numeric and `0` is a falsy equivalent that you can use for conditional rendering
```suggestion
      {todos.length && (
```

src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    if (todoName === '') {
+      onError(ErrorMessages.emptyTitle);
+
+      return;
+    }
+
+    createTodo(todoName);
+    setNewTodoTitle('');
+  };
```
meforoma, line 41: why not focus input after adding new todo?
```suggestion
    setNewTodoTitle('');
    inputTitle.current?.focus();
  };
```

meforoma (approved):
approving ))
make sure to apply the array.length validation 😉

PR #328, opened 2023-02-12
Develop

Grayfox147:
[DEMO LINK](https://Grayfox147.github.io/react_todo-app-with-api/)


SalehBUD (approved):

meforoma (changes requested):
@Grayfox147 good job 👏🏼 
yet there are a few glitches and improvement - see the video and fix those

https://user-images.githubusercontent.com/39308717/219037070-810464c3-881e-45f9-b080-5d7d96b87d4e.mov

Grayfox147:
@meforoma I believe I implemented what you asked, also I dont see that behavior on my code when I run NPM start, or when I open the demopage, I am not sure why is that difference.

src/App.tsx
```diff
@@ -1,24 +1,247 @@
...
+import { TodoList } from './components/TodoList';
+import { Errors, Filter, Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 6101;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState('all');
```
naziamoff, line 22: create enum for filter, string values are really difficult to maintain

naziamoff: oh, I see you already have one. Why not just add `Filter.All` for that?

src/App.tsx
```diff
@@ -1,24 +1,247 @@
...
+            />
+          </>
+        )}
+        {error && (
+          <div
+            className={classNames(
+              'notification is-danger is-light has-text-weight-normal',
+              { hidden: !error },
+            )}
+          >
```
naziamoff, line 233:
```suggestion
          <div className={classNames(
              'notification is-danger is-light has-text-weight-normal',
              { hidden: !error },
            )}
          >
```

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,50 @@
...
+export const client = {
+  get: <T>(url: string) => request<T>(url),
+  post: <T>(url: string, data: any) => request<T>(url, 'POST', data),
+  patch: <T>(url: string, data: any) => request<T>(url, 'PATCH', data),
+  delete: (url: string) => request(url, 'DELETE'),
+};
+// on Submit I will have to use the post method, to add the todos
+// get should be when rendering to get what we have so far
+//  patch it would be to update the todo
+//  delete = remove it
```
naziamoff, line 50: do you need those comments?

naziamoff (approved):
🔥 
Just a couple of suggestions

src/App.tsx
Grayfox147: Done!

src/utils/fetchClient.ts
Grayfox147: Gone! :D

meforoma (approved):

PR #329, opened 2023-02-14
Solution

Rijo-J:
- [DEMO LINK](https://rijo-j.github.io/react_todo-app-with-api/)
missing fix behavior on todo delete while editing todo with empty input

### DIFF
--- a/src/components/Todos/Todos.tsx
+++ b/src/components/Todos/Todos.tsx
@@ -1,0 +1,85 @@
+import cn from 'classnames';
+import React from 'react';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[],
+  onDeleteTodo: (id: number) => void,
+  updatingTodos: number[],
+};
+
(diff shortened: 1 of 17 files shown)


src/components/Todos/Todos.tsx
```diff
@@ -0,0 +1,136 @@
...
+            </label>
+
+            {id === editTodoId ? (
+              <>
+                <form
+                  onSubmit={(e) => {
+                    e.preventDefault();
+                    handleTitleUpdate(id, title);
+                  }}
+                >
```
meforoma, line 78: i think this is the root cause
```suggestion
                <form
                  onSubmit={(e) => {
                    e.preventDefault();
                    e.target.value
                      ? handleTitleUpdate(id, title)
                      : deleteTodo(id);
                  }}
                >
```

Rijo-J: ✅ fixed above!

SalehBUD (approved):

PR #339, opened 2023-02-18
Develop

anna-inozemets:
[DEMO LINK](https://anna-inozemets.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,24 +1,238 @@
...
+  const hasCompletedTodos = todos.some(todo => todo.completed);
+
+  const loadTodos = async () => {
+    try {
+      const loadedTodos = await getTodos(USER_ID);
+
+      setTodos(loadedTodos);
+    } catch (error) {
+      setIsError(true);
+      closeNotification(setIsError, false, 3000);
```
Avichez, line 49: It seems to me, here it also needs to add `setErrorMessage()` something like "Unable to load todos".

src/App.tsx
```diff
@@ -1,24 +1,238 @@
...
+
+  const preparedTodos = useMemo(() => {
+    return prepareTodo(todos, filterBy);
+  }, [todos, filterBy]);
+
+  const handleAddTodo = async (todoTitle: string) => {
+    if (todoTitle.length === 0) {
+      setIsError(true);
+      setErrorMessage(ErrorMessages.TITLE);
+      closeNotification(setIsError, false, 3000);
```
Avichez, line 66: As a suggestion, you can combine this logic into one function and you then just pass inside `ErrorMessages` depending on the context.

Avichez (approved):
Great job 🙂 👍 
Just in case to pass tests needs to push also package.json and package-lock.json

1) onBlur doesn't save inserted text just throwing the Error, if I am not mistaken it should work as you did the submit function on "Enter".
image

src/App.tsx
anna-inozemets: Yeah, that's great idea, I'll fix it, thanks!)

denys-cheporniuk (changes requested):
Screenshot 2023-02-20 at 17 27 10

1. only one todo can be edited at a time
2. cursor should be in the input when input appears

denys-cheporniuk (commented):
Screenshot 2023-02-20 at 17 31 55
todo with spaces shouldn't be in the list

src/App.tsx
```diff
@@ -1,24 +1,252 @@
...
+    setTitle(newTitle);
+  };
+
+  const changeFilterBy = (filter: FilterBy) => {
+    setFilterBy(filter);
+  };
+
+  const changeIsError = (value: boolean) => {
+    setIsError(value);
+  };
```
denys-cheporniuk, line 51: useCallback should be here

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+      </a>
+    </nav>
+
+    <button
+      type="button"
+      className="todoapp__clear-completed"
+      onClick={onDeleteCopletedTodos}
+    >
+      {hasCompletedTodos && 'Clear completed'}
+    </button>
```
denys-cheporniuk, line 66: at best, the button should not be displayed if hasCompletedTodos is false, not just the text, but the entire button
And you'll probably have problems with styles when you hide the button, the styles should also be fixed

src/types/ErrorMessages.ts
```diff
@@ -0,0 +1,8 @@
+export enum ErrorMessages {
+  NONE = '',
+  LOAD = 'Can\'t load todos from server',
+  TITLE = 'Title can\'t be empty',
+  ADD = 'Unable to add a todo',
+  DELETE = 'Unable to delete a todo',
+  UPDATE = 'Unable to update a todo',
```
denys-cheporniuk, line 7: Usually, we do not store all the text inside the enum, enum is used for a short list of values
Display different text inside component depending on this enum

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+    options.body = JSON.stringify(data);
+    options.headers = {
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // we wait for testing purpose to see loaders
+  return wait(300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
```
denys-cheporniuk, line 32: every time when I see `then` in code I want to cry

denys-cheporniuk (approved):

PR #344, opened 2023-02-21
Solution

G0odvin:
[DEMO LINK](https://G0odvin.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,305 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+/* eslint-disable no-param-reassign */
+import React, {
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import {
+  createTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+} from './api/todos';
+import { AuthContext } from './components/Auth/AuthContext';
+import { NewTodoField } from './components/NewTodoField/NewTodoField';
+import { TodoFilters } from './components/TodoFilters/TodoFilters';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [isLoadError, setIsLoadError] = useState(false);
+  const [isUploadError, setIsUploadError] = useState(false);
+  const [isRemoveError, setIsRemoveError] = useState(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [filtredTodos, setFiltredTodos] = useState<Todo[]>([]);
+  const [isEmptyTitle, setIsEmptyTitle] = useState(false);
+  const [isTodoLoading, setIsTodoLoading] = useState(false);
+  const [isUpdatedError, setIsUpdatedError] = useState(false);
+  const [deletedTodosId, setDeletedTodosId] = useState<number[]>([]);
+  const [activeTodoId, setActiveTodoId] = useState<number[]>([]);
+  const [isUpdatedTodo, setIsUpdatedTodo] = useState(false);
+
+  const user = useContext(AuthContext);
+
+  const loadTodosFromServer = async () => {
+    try {
+      if (user) {
+        const getTodosFromServer = await getTodos(user.id);
+
+        setTodos(getTodosFromServer);
+      }
+    } catch (error) {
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,24 +1,305 @@
...
+        const newTodo = await createTodo({
+          title: todo.title,
+          userId: todo.userId,
+          completed: false,
+        });
+
+        setTodos(prevTodos => [...prevTodos, newTodo]);
+        setIsTodoLoading(false);
+        setTempTodo(null);
+        setActiveTodoId([]);
```
anstsot, line 65: it would be great to change the order here. Because now there is a second where the loading and newTodo is shown (check the video please)
https://user-images.githubusercontent.com/22824431/220374785-773ef058-4ee6-4902-a2c5-d5dff7907178.mov

src/App.tsx
```diff
@@ -1,24 +1,305 @@
...
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [isLoadError, setIsLoadError] = useState(false);
+  const [isUploadError, setIsUploadError] = useState(false);
+  const [isRemoveError, setIsRemoveError] = useState(false);
```
anstsot, line 24: you have different state variables for errors. I would recommend creating `enum` for `ErrorType`. Then you can just have one state `[error, setError]` for different errors and show text according to the type.

src/App.tsx
```diff
@@ -1,24 +1,305 @@
...
+        return todo;
+      });
+
+      setTodos(checkedTodoList);
+    } catch {
+      setIsUpdatedError(true);
+    }
+  };
+
+  const completedTodos = useMemo(() => {
```
anstsot, line 151: you're using `useMemo`, it's great!

src/App.tsx
```diff
@@ -1,24 +1,305 @@
...
+          <button
+            data-cy="HideErrorButton"
+            type="button"
+            className="delete"
+            onClick={() => setIsUpdatedError(false)}
+          />
+
+          Unable to update a todo
+        </div>
+      )}
```
anstsot, line 302: with `enum ErrorType` you can reduce the code and just change the text of error

anstsot (changes requested):
Looks good! Let's add few changes and make your code better

src/App.tsx
```diff
@@ -1,24 +1,233 @@
...
+} from './components/ErrorNotification/ErrorNotification';
+import { ErrorType } from './types/ErrorType';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [isUploadError, setIsUploadError] = useState(false);
+  const [error, setError] = useState(ErrorType.None);
+  const [isError, setIsError] = useState(false);
```
anstsot, line 28: you don't need these 2 values here. you can just simplify the logic and use `error` only

src/App.tsx
```diff
@@ -1,24 +1,233 @@
...
+        />
+      </div>
+
+      {isError && (
+        <ErrorNotification
+          error={error}
+          setIsError={setIsError}
+          isError={isError}
+        />
+      )}
```
anstsot, line 230: you can always render the `ErrorNotification` component and add the check inside. (check my comments there).

or just replace the check with: `error !== ErrorType.None`

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,65 @@
...
+          break;
+        case ErrorType.EmptyTitle:
+          setErrorMesage(ErrorType.EmptyTitle);
+          break;
+
+        default:
+          setErrorMesage('');
+      }
+    }, [error]);
+
```
anstsot, line 48: you can use the check in App.tsx or add it here:
```suggestion
    if (error === ErrorType.None) return null;
```

src/App.tsx
```diff
@@ -1,24 +1,233 @@
...
+        setActiveTodoId([]);
+        setIsTodoLoading(false);
+      } catch {
+        setIsUploadError(true);
+        setTodos(todos);
+        setTempTodo(null);
+        setIsTodoLoading(false);
+        setError(ErrorType.UploadError);
+        setIsError(true);
+      }
```
anstsot, line 77: this code can be simplified by using `finally`
```suggestion
        setTodos(prevTodos => [...prevTodos, newTodo]);
        setActiveTodoId([]);        
      } catch {
        setError(ErrorType.UploadError);
      } finally {
        setTempTodo(null);      
        setIsTodoLoading(false);      
      }
```

anstsot (changes requested):
Let's leave only one `error` with different types for the whole application because now there are still 3 types (isUploadError, error, isError)

witflash (approved):
Well done 👍

PR #348, opened 2023-02-21
It is not all

PolinaSukhorukova:
[DEMO LINK](https://PolinaSukhorukova.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,241 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useState, useCallback, useEffect } from 'react';
 
-const USER_ID = 0;
+// components
+import { UserWarning } from './UserWarning';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { ErrorMessage } from './components/ErrorMessage';
+
+// utils for loading and updating
+import {
+  getTodos,
+  addTodo,
+  deleteTodo,
+  updateStatusTodo,
+  updateTitleTodo,
+} from './api/todos';
+
+// utils
+import { getFilteredTodos } from './utils/getFilteredTodos';
+
+// types
+import { Todo } from './types/Todo';
+import { FilterType } from './types/FilterType';
+
+const USER_ID = 6354;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosWithLoader, setTodosWithLoader] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const [isLoadingFailed, setIsLoadingFailed] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const [filterBy, setFilterBy] = useState<FilterType>(FilterType.ALL);
+
+  const visibleTodos = getFilteredTodos(todos, filterBy);
+  const completedTodos = todos.filter(todo => todo.completed === true);
+
+  const numberOfNotCompletedTodos = todos.length - completedTodos.length;
+  const isClearAllButtonVisible = Boolean(!completedTodos.length);
+
+  const isSubmitButtonDisabled = Boolean(tempTodo?.title.length);
+
+  const addErrorMessage = useCallback((newMessage: string) => {
+    setErrorMessage(newMessage);
+  }, []);
+
+  const getAllTodos = useCallback(async () => {
+    setIsLoadingFailed(false);
+
+    try {
+      const allTodos = await getTodos(USER_ID);
+
+      setTodos(allTodos);
+    } catch {
+      setIsLoadingFailed(true);
+      addErrorMessage('Unable to update a todo');
+    }
+  }, []);
+
+  useEffect(() => {
+    getAllTodos();
+  }, []);
+
+  const clearMessage = useCallback(() => {
+    setErrorMessage('');
+  }, []);
+
+  const addNewTodo = useCallback(async (title: string) => {
+    clearMessage();
+
+    if (!title) {
+      addErrorMessage('Title can\'t be empty');
+
+      return;
+    }
+
+    const newTodo = {
+      id: 0,
+      userId: USER_ID,
+      title,
+      completed: false,
+    };
+
+    setTempTodo(newTodo);
+
+    try {
+      await addTodo(USER_ID, newTodo);
+      await getAllTodos();
+    } catch {
+      addErrorMessage('Unable to load todos');
+    } finally {
+      setTempTodo(null);
+    }
+  }, []);
+
+  const removeTodo = useCallback(async (todo: Todo) => {
+    clearMessage();
+
+    setTodosWithLoader(currentTodos => [...currentTodos, todo]);
+
+    try {
+      await deleteTodo(USER_ID, todo);
+      await getAllTodos();
+    } catch {
+      addErrorMessage('Unable to update a todo');
+    } finally {
+      setTodosWithLoader(currentTodos => (
+        currentTodos.filter(t => t.id !== todo.id)
+      ));
+    }
+  }, []);
+
+  const removeAllCompletedTodos = useCallback(async () => {
+    clearMessage();
+
+    setTodosWithLoader(currentTodos => [...currentTodos, ...completedTodos]);
+
+    try {
+      await Promise.all(
+        completedTodos
+          .map((todo: Todo) => deleteTodo(USER_ID, todo)),
+      );
+      await getAllTodos();
+    } catch {
+      addErrorMessage('Unable to delete completed todos');
+    } finally {
+      setTodosWithLoader(currentTodos => (
+        currentTodos.filter(todo => todo.completed !== true)
+      ));
+    }
+  }, [completedTodos]);
+
+  const changeStatus = useCallback(async (todoChangeStatus: Todo) => {
+    const newTodoStatus = !todoChangeStatus.completed;
+    const todoIdChangeStatus = todoChangeStatus.id;
+
+    setTodosWithLoader(currentTodos => [...currentTodos, todoChangeStatus]);
+
+    try {
+      await updateStatusTodo(todoIdChangeStatus, newTodoStatus);
+      await getAllTodos();
+    } catch {
+      addErrorMessage('Unable to update a todo');
+    } finally {
+      clearMessage();
+      setTodosWithLoader(currentTodos => (
+        currentTodos.filter(t => t.id !== todoChangeStatus.id)
+      ));
+    }
+  }, []);
+
+  const changeTitle = useCallback(async (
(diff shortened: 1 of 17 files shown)


src/App.tsx
```diff
@@ -1,24 +1,254 @@
...
+      userId: USER_ID,
+      title,
+      completed: false,
+    };
+
+    setTempTodo(newTodo);
+
+    try {
+      await addTodo(USER_ID, newTodo);
+      await getAllTodos();
```
Anastasiia-Khudych, line 101: Id for new todo will be set on the server, so you shouldn't pass newTodo with id  to `addTodo`

src/App.tsx
```diff
@@ -1,24 +1,254 @@
...
+      <div className="todoapp__content">
+        {!isLoadingFailed && (
+          <>
+            <Header
+              addNewTodo={addNewTodo}
+              addErrorMessage={addErrorMessage}
+              isSubmitButtonDisabled={isSubmitButtonDisabled}
+              toggleStatusForAllTodos={toggleStatusForAllTodos}
+              isToogleButtonVisible={isToogleButtonVisible}
+            />
```
Anastasiia-Khudych, line 224: You use `await getAllTodos()` where you set the value for `isLoadingFailed ` in `toggleStatusForAllTodos `, `changeStatus` and other functions, so maybe if there is an error after changing the todos you will not see all previously added todos at all. Maybe it's better to use check with todos length here.

src/types/ErrorMessages.ts
```diff
@@ -0,0 +1,10 @@
+export enum ErrorMessages {
+  NONE = '',
+  LOAD = 'Unable to load todos',
+  TITLE = 'Title can\'t be empty',
+  ADD = 'Unable to add the todo',
+  DELETE = 'Unable to delete the todo',
+  DELETE_COMPLETED = 'Unable to delete completed todos',
+  UPDATE = 'Unable to update a todo',
+  UPDATE_ALL = 'Unable to update status of todos',
```
Anastasiia-Khudych, line 9: I am not sure that it is a good idea to store whole sentences in enum, maybe it's better to check for ErrorMessages value in your ErrorMessage.tsx and show these sentences there.

Anastasiia-Khudych (changes requested):
Pretty good job but check the comments. The filter doesn't work as expected. There are no loaders when toggling all todos. If I try to edit the title for one particular todo, save it and then edit one more time, I will see the previous title in the input, but not the current one, and if I save it again it will be the previous title.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+  removeAllCompletedTodos,
+  numberOfNotCompletedTodos,
+  isClearAllButtonVisible,
+}) => (
+  <footer className="todoapp__footer">
+    <span className="todo-count">
+      {`${numberOfNotCompletedTodos} items left`}
+    </span>
+
+    {/* Active filter should have a 'selected' class */}
```
alexandra-protyanova, line 25: remove comments

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+        className={cn('filter__link', {
+          selected: filterBy === FilterType.COMPLETED,
+        })}
+        onClick={() => setFilterBy(FilterType.COMPLETED)}
+      >
+        Completed
+      </a>
+    </nav>
+
+    {/* don't show this button if there are no completed todos */}
```
alexandra-protyanova, line 58: haven't completed

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+    if (title !== addedTitle) {
+      changeTitle(todo, addedTitle);
+    }
+
+    if (!addedTitle) {
+      removeTodo(todo);
+    }
+
+    changeInputActiveStatus(false);
+    // setAddedTitle(title);
```
alexandra-protyanova, line 43: delete comments

src/types/ErrorMessages.ts
```diff
@@ -0,0 +1,21 @@
...
+// export enum ErrorMessages {
+//   NONE = '',
+//   LOAD = 'Unable to load todos',
+//   TITLE = 'Title can\'t be empty',
+//   ADD = 'Unable to add the todo',
+//   DELETE = 'Unable to delete the todo',
+//   DELETE_COMPLETED = 'Unable to delete completed todos',
+//   UPDATE = 'Unable to update a todo',
+//   UPDATE_ALL = 'Unable to update status of todos',
+// }
```
alexandra-protyanova, line 21: as for me, this looks much better

PolinaSukhorukova: @alexandra-protyanova  As for me, I agree with you. But after  @Anastasiia-Khudych  comment, I read a lot of information about enum, and noticed that she was right, usually, enum is used for a short list of values, not for storing all the text of error. So I decided to leave the second version.

src/App.tsx
```diff
@@ -1,24 +1,258 @@
...
+} from 'react';
+
+// components
 import { UserWarning } from './UserWarning';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { ErrorMessage } from './components/ErrorMessage';
+
+// utils for loading and updating
```
ArturErmolenko, line 15: I believe  final code should be clear without comments , you can left comment like TODO or explain hard logic in extra cases , but write comment to import its overkill 😃

src/App.tsx
```diff
@@ -1,24 +1,258 @@
...
+    setTodosWithLoader(currentTodos => [...currentTodos, todo]);
+
+    try {
+      await deleteTodo(USER_ID, todo);
+      await getAllTodos();
+    } catch {
+      addErrorMessage(ErrorMessages.DELETE);
+    } finally {
+      setTodosWithLoader(currentTodos => (
+        currentTodos.filter(t => t.id !== todo.id)
```
ArturErmolenko, line 125: `t `is bad naming

src/types/ErrorMessages.ts
ArturErmolenko: ok, in that case, let's delete the commented code

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+            onSubmit={handleSubmit}
+          >
+            <input
+              className="todo__title-field"
+              type="text"
+              value={addedTitle}
+              onChange={handleInput}
+              onBlur={handleSaveChanges}
+              onKeyUp={handleKeyUp}
+              // eslint-disable-next-line jsx-a11y/no-autofocus
```
ArturErmolenko, line 84: let's try to fix that issue I believe google help you fix it with first search result )

src/components/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,48 @@
...
+      {typeOfError === ErrorMessages.ADD && (
+        'Unable to add the todo'
+      )}
+
+      {typeOfError === ErrorMessages.DELETE && (
+        'Unable to delete the todo'
+      )}
+
+      {typeOfError === ErrorMessages.DELETE_COMPLETED && (
+        'nable to delete completed todos'
```
ArturErmolenko, line 36: typo error , that's why we don't rely on unknown string , we use enums instead

src/components/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,48 @@
...
+        'Unable to load todos'
+      )}
+
+      {typeOfError === ErrorMessages.TITLE && (
+        'Title can\'t be empty'
+      )}
+
+      {typeOfError === ErrorMessages.ADD && (
+        'Unable to add the todo'
+      )}
```
ArturErmolenko, line 29: such  code definitely cannot get into the master.
 Can make a function with a switch case that will determine exactly which error needs to be shown and transfer a ready-made message to this component

src/App.tsx
```diff
@@ -1,24 +1,258 @@
...
+  const isClearAllButtonVisible = Boolean(!completedTodos.length);
+  const isToogleButtonVisible = Boolean(!numberOfNotCompletedTodos);
+  const isSubmitButtonDisabled = Boolean(tempTodo?.title.length);
+  const isTodoListVisible = Boolean(todos.length > 0);
+
+  const addErrorMessage = useCallback((newMessage: ErrorMessages) => {
+    setErrorMessage(newMessage);
+  }, []);
+
+  const getAllTodos = useCallback(async () => {
```
ArturErmolenko, line 63: lets manage loading here

src/App.tsx
```diff
@@ -1,24 +1,258 @@
...
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <Header
```
ArturErmolenko, line 222: add Loader here while waiting for todos from server

ArturErmolenko (changes requested):

ArturErmolenko (commented):
🥲 
https://user-images.githubusercontent.com/94292654/221578152-10c4518e-81d9-4637-9184-38700720ff74.mov

src/App.tsx
```diff
@@ -1,24 +1,254 @@
...
+    getFilteredTodos(todos, filterBy)
+  ), [todos, filterBy]);
+
+  const completedTodos = useMemo(() => (
+    todos.filter(todo => todo.completed)
+  ), [todos]);
+
+  const numberOfNotCompletedTodos = todos.length - completedTodos.length;
+
+  const isClearAllButtonVisible = Boolean(!completedTodos.length);
```
alexandra-protyanova, line 50:
```suggestion
  const isClearAllButtonVisible = !!completedTodos.length;
```

src/App.tsx
```diff
@@ -1,24 +1,254 @@
...
+  const isToogleButtonVisible = Boolean(!numberOfNotCompletedTodos);
+  const isSubmitButtonDisabled = Boolean(tempTodo?.title.length);
+  const isTodoListVisible = Boolean(todos.length > 0);
```
alexandra-protyanova, line 53:
```suggestion
  const isTodoListVisible = todos.length > 0;
```

alexandra-protyanova: Compressing is boolean by default

src/utils/chooseTheError.ts
```diff
@@ -0,0 +1,29 @@
...
+    case ErrorMessages.UPDATE:
+      return 'Unable to update a todo';
+
+    case ErrorMessages.UPDATE_ALL:
+      return 'Unable to update status of todos';
+
+    default:
+      throw new Error('No such error');
+  }
+};
```
alexandra-protyanova, line 29: why don't you set values to your keys in enum, it will work the same way

alexandra-protyanova (approved):
Good job 👍

PR #350, opened 2023-02-22
Develop

Anastasiia-Khudych:
[Demo](https://anastasiia-khudych.github.io/react_todo-app-with-api/)


src/components/ErrorNotifications/ErrorNotifications.tsx
```diff
@@ -0,0 +1,66 @@
...
+        'Unable to add a todo'
+      )}
+
+      {error === Error.ONDELETE && (
+        'Unable to delete a todo'
+      )}
+
+      {error === Error.ONUPDATE && (
+        'Unable to update a todo'
+      )}
```
svitjojo, line 63: Maybe you can use your enum and set an error message for all these cases, and after use just {error}

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Filter } from '../../types/Filter';
+// import { Todo } from '../../types/Todo';
```
svitjojo, line 4: remove comments

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,88 @@
...
+    let completedValue;
+
+    if (isAllTodosCompleted) {
+      completedValue = false;
+    } else {
+      completedValue = true;
+    }
+
+    toggleAllTodosStatus(completedValue);
+  };
```
svitjojo, line 58:
```suggestion
  const handleToggleButtonClick = () => {
    toggleAllTodosStatus(isAllTodosCompleted);
  };
```

svitjojo (changes requested):
Good job 👍 
One more thing - toggle all button suppose to update only active todos, in your case, it updates all, even if they're are some of them already completed.

src/components/ErrorNotifications/ErrorNotifications.tsx
Anastasiia-Khudych: Thank you for reviewing! I fixed toggleAll problem. But I am not sure that it's a good idea to store error sentences in an enum. If I got it right enum values are usually short and used for example for sorting purposes.

src/components/ErrorNotifications/ErrorNotifications.tsx
```diff
@@ -0,0 +1,66 @@
...
+      {
+        hidden: error === Error.NONE,
+      },
+    )}
+    >
+      <button
+        type="button"
+        className="delete"
+        aria-label="Delete"
+        onClick={() => handleCloseButtonClick()}
```
denys-cheporniuk, line 41:
```suggestion
        onClick={handleCloseButtonClick}
```

src/components/ErrorNotifications/ErrorNotifications.tsx
denys-cheporniuk: agree, storing text in enum isn't a good idea

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,90 @@
...
+          >
+            Completed
+          </a>
+        </nav>
+
+        {completedTodos !== 0 && (
+          <button
+            type="button"
+            className="todoapp__clear-completed"
+            onClick={() => handleClearButtonClick()}
```
denys-cheporniuk, line 82:
```suggestion
            onClick={handleClearButtonClick}
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  return (
+    <div
+      className={classNames(
+        'todo',
+        {
+          completed,
+        },
+      )}
+      onDoubleClick={() => handleDoubleClick()}
```
denys-cheporniuk, line 118:
```suggestion
      onDoubleClick={handleDoubleClick}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,94 @@
...
+    removeDeleteId,
+    changeCompletedStatus,
+    isUpdateWaiting,
+    todosIdsToUpdate,
+    changeTodosIdsToUpdate,
+    removeUpdatedId,
+    changeTodoTitle,
+  },
+) => {
+  return (
```
denys-cheporniuk, line 40: you can avoid return here

denys-cheporniuk (changes requested):
would be great to see more useCallback, useMemo and React.memo

denys-cheporniuk (approved):

PR #354, opened 2023-02-22
solution

Kravchenko71213:
https://Kravchenko71213.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,274 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import { prepareTodos } from './utils/prepareTodos';
+import { TodoList } from './components/TodoList';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import {
+  callAddTodo,
+  callDeleteTodo,
+  callEditTodo,
+  callGetTodos,
+} from './api/todos';
+import { Notifications } from './components/Notifications';
+import { ErrorType } from './types/ErrorType';
+import { FilterBy } from './types/FilterBy';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { RequestTodo } from './types/RequestTodo';
 
-const USER_ID = 0;
+const USER_ID = 6414;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorType, setErrorType] = useState<ErrorType>(ErrorType.LOAD);
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
+  const [hasError, setHasError] = useState(false);
+  const [title, setTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todoWithLoader, setTodoWithLoader] = useState(0);
+  const [isToggle, setIsToggle] = useState(false);
+  const [isClearCompleted, setClearCompleted] = useState(false);
+  const [editingId, setEditingId] = useState(0);
+
+  const isAllTodosActive = useMemo(() => {
+    return todos.every(todo => todo.completed);
+  }, [todos]);
+
+  const [isEdit, setIsEdit] = useState(false);
+
+  const getTodos = async () => {
+    try {
+      const todosFromServer = await callGetTodos(USER_ID);
+
+      setTodos(todosFromServer);
+      setHasError(false);
+    } catch (error) {
+      setHasError(true);
+      if (error instanceof Error) {
+        setErrorType(ErrorType.LOAD);
+      }
+    }
+  };
+
+  useEffect(() => {
+    getTodos();
+  }, []);
+
+  const addTodo = useCallback(async (titleOfTodo: string) => {
+    const preparedTitle = titleOfTodo.trim();
+
+    if (!preparedTitle) {
+      setErrorType(ErrorType.TITLE);
+      setHasError(true);
+      setTitle('');
+
+      return;
+    }
+
+    const newTodo: RequestTodo = {
+      userId: USER_ID,
+      title: preparedTitle,
+      completed: false,
+    };
+
+    try {
+      setTempTodo({ ...newTodo, id: 0 });
+      await callAddTodo(newTodo);
+      await getTodos();
+      setTempTodo(null);
+      setTitle('');
+      setHasError(false);
+    } catch (error) {
+      setHasError(true);
+      setErrorType(ErrorType.ADD);
+    } finally {
+      setTempTodo(null);
+    }
+  }, []);
+
+  const deleteTodo = useCallback(async (id: number) => {
+    setTodoWithLoader(id);
+
+    try {
+      await callDeleteTodo(id);
+      await getTodos();
+    } catch (error) {
+      setErrorType(ErrorType.DELETE);
+      setHasError(true);
+    } finally {
+      setTodoWithLoader(0);
+    }
+  }, []);
+
+  const deleteCompleted = async (id: number) => {
+    setClearCompleted(true);
+
+    try {
+      await deleteTodo(id);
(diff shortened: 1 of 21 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,87 @@
...
+  onClearCompleted: (id: number) => void,
+  // isClearButtonVisible: boolean;
+};
+
+export const Footer: React.FC<Props> = React.memo(({
+  todos,
+  onFilterBy,
+  filterBy,
+  // notCompletedTodos,
+  // isClearButtonVisible,
```
tania-boiar, line 20: you could get rid off comments

tania-boiar (approved):
Good job! ⚡️

As for me, I'm not sure about your using states. It seems that there are more of them than are actually needed.
Maybe, it's ok.)
Interesting to see the feedback from the mentor.

src/App.tsx
```diff
@@ -1,24 +1,274 @@
...
+const USER_ID = 6414;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorType, setErrorType] = useState<ErrorType>(ErrorType.LOAD);
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
+  const [hasError, setHasError] = useState(false);
+  const [title, setTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todoWithLoader, setTodoWithLoader] = useState(0);
```
Anna-Tkachenko, line 35: it's not clear why `todoWithLoader` contains some number
number can represent `id` value or something else
but in this variable I expect to see some `todo`
let's think more about variable name

src/App.tsx
```diff
@@ -1,24 +1,274 @@
...
+  const [isToggle, setIsToggle] = useState(false);
+  const [isClearCompleted, setClearCompleted] = useState(false);
+  const [editingId, setEditingId] = useState(0);
```
Anna-Tkachenko, line 38: will be better to add noun here, to clarify that it's todo id

src/App.tsx
```diff
@@ -1,24 +1,274 @@
...
+
+  const isAllTodosActive = useMemo(() => {
+    return todos.every(todo => todo.completed);
+  }, [todos]);
+
+  const [isEdit, setIsEdit] = useState(false);
```
Anna-Tkachenko, line 44: move this variable to the top (where you declared all state variables)

src/App.tsx
```diff
@@ -1,24 +1,274 @@
...
+
+    const newTodo: RequestTodo = {
+      userId: USER_ID,
+      title: preparedTitle,
+      completed: false,
+    };
+
+    try {
+      setTempTodo({ ...newTodo, id: 0 });
+      await callAddTodo(newTodo);
```
Anna-Tkachenko, line 83: can we name it just `addTodo`? do we need to add extra `call`?

src/App.tsx
```diff
@@ -1,24 +1,274 @@
...
+      await getTodos();
+      setTempTodo(null);
```
Anna-Tkachenko, line 85: do we need this resetting here? you've already did it in `finally` section

src/App.tsx
```diff
@@ -1,24 +1,274 @@
...
+    } finally {
+      setTempTodo(null);
+    }
+  }, []);
+
+  const deleteTodo = useCallback(async (id: number) => {
+    setTodoWithLoader(id);
+
+    try {
+      await callDeleteTodo(id);
```
Anna-Tkachenko, line 100: same here, `call` in function name is redundant here

src/App.tsx
```diff
@@ -1,24 +1,274 @@
...
+      setTodoWithLoader(0);
+    }
+  }, []);
+
+  const deleteCompleted = async (id: number) => {
+    setClearCompleted(true);
+
+    try {
+      await deleteTodo(id);
+      setTodos((currentTodos) => currentTodos.filter(todo => todo.id !== id));
```
Anna-Tkachenko, line 115: will be better to retrieve existing todos from server

src/App.tsx
```diff
@@ -1,24 +1,274 @@
...
+      setErrorType(ErrorType.DELETE);
+    } finally {
+      setClearCompleted(false);
+    }
+  };
+
+  const editTodoStatus = useCallback(async (editedTodo: Todo) => {
+    setTodoWithLoader(editedTodo.id);
+
+    const undatedTodo = {
```
Anna-Tkachenko, line 126:
```suggestion
    const updatedTodo = {
```

src/App.tsx
```diff
@@ -1,24 +1,274 @@
...
+      setEditingId(0);
+      setIsEdit(false);
+    }
+  }, []);
+
+  const handleTitleTodo = useCallback((newTitle: string) => {
+    setTitle(newTitle);
+  }, [title]);
+
+  const handleToggleStatus = async () => {
```
Anna-Tkachenko, line 171: you've created so many functions in one component
can we separate them somehow?
move to some child components instead of declare all of them at the parent one?

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+export const Footer: React.FC<Props> = React.memo(({
+  todos,
+  onFilterBy,
+  filterBy,
+  onClearCompleted,
+}) => {
+  const completedTodos = todos.filter((todo) => todo.completed);
+  const activeTodos = todos.filter((todo) => !todo.completed);
+  const handleClearCompleted = () => {
```
Anna-Tkachenko, line 21: what is the purpose if this function? it doesn't return anything, and doesn't do anything, check it

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,83 @@
...
+    completedTodos.map((todo) => onClearCompleted(todo.id));
```
Anna-Tkachenko, line 22: I see, probably you wanted to call some action for each todo
for this purpose let's use `forEach` method instead of the `map` one

Anna-Tkachenko (changes requested):
Check the process of updating the title for todo on the demo link. It doesn't work as expected, when I update the title, and then click on it one more time - I still see the previous value, it doesn't save the new one

src/App.tsx
```diff
@@ -1,24 +1,252 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState <Error>(Error.NONE);
+  const [filterType, setFilterType] = useState<Filter>(Filter.ALL);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [isAddWaiting, setIsAddWaiting] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isDeleteWaiting, setIsDeleteWaiting] = useState(false);
+  const [todosIdsToRemove, setTodosIdsToRemove] = useState<number[]>([]);
+  const [todosIdsToUpdate, setTodosIdsToUpdate] = useState<number[]>([]);
+  const [isUpdateWaiting, setIsUpdateWaiting] = useState(false);
```
Anna-Tkachenko, line 37: can we use one variable for waiting some response from server instead of creating separate boolean variable for each action?

src/App.tsx
```diff
@@ -1,24 +1,252 @@
...
+              todosIdsToRemove={todosIdsToRemove}
+              changeTodosIdsToRemove={changeTodosIdsToRemove}
+              removeDeleteId={removeDeleteId}
+              changeCompletedStatus={changeCompletedStatus}
+              isUpdateWaiting={isUpdateWaiting}
+              todosIdsToUpdate={todosIdsToUpdate}
+              changeTodosIdsToUpdate={changeTodosIdsToUpdate}
+              removeUpdatedId={removeUpdatedId}
+              changeTodoTitle={changeTodoTitle}
+            />
```
Anna-Tkachenko, line 231: check the code, it's so many props that you pass to the child component
can we decrease the number, and move some functions and properties from main component to the child ones?

Anna-Tkachenko (changes requested):

denys-cheporniuk (approved):

PR #355, opened 2023-02-22
Develop

Coyoote:
[DEMO LINK](https://Coyoote.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,193 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import { debounce } from 'lodash';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import {
+  addTodo,
+  getTodos,
+  removeTodo,
+  updateTodo,
+} from './api/todos';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { Notification } from './components/Notification';
+import { TodoList } from './components/TodoList';
+import { ErrorType } from './types/ErrorType';
+import { FilterType } from './types/FilterType';
+import { Todo } from './types/Todo';
+import { prepareTodos } from './utils/prepareTodos';
 
-const USER_ID = 0;
+const USER_ID = 6396;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+  const [isTitleDisabled, setIsTitleDisabled] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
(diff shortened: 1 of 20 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,48 @@
...
+      {/* this buttons is active only if there are some active todos */}
+      <button
+        type="button"
+        className={classNames(
+          'todoapp__toggle-all', {
+            active: activeTodos,
+          },
+        )}
+        disabled={activeTodos}
+      />
```
Dashaverz, line 37: doesn't work

Dashaverz (changes requested):
GJ! left some comments

src/App.tsx
```diff
@@ -1,24 +1,234 @@
...
+    if (!todoTitleTrim.length) {
+      setHasError(true);
+      setErrorType(ErrorType.Title);
+      clearErrorAfterDelay();
+
+      return;
+    }
+
+    const todoToAdd = {
+      id: 0,
```
ArturErmolenko, line 87: 0 isn't valid value for id, you shouldn't send id for todo on server

src/App.tsx
```diff
@@ -1,24 +1,234 @@
...
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+
+        <Header
```
ArturErmolenko, line 194: add Loader here while we waiting for todos fetched

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,50 @@
...
+type Props = {
+  title: string,
+  todos: Todo[],
+  handleInput: (input: string) => void,
+  handleAddTodo: (todoTitle: string) => void,
+  isTitleDisabled: boolean,
+  handleAllTodosStatus: () => void,
+};
+
+export const Header: React.FC<Props> = ({
```
ArturErmolenko, line 16: memo

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,58 @@
...
+  handleDeleteTodo: (todoId: number) => void,
+  tempTodo: Todo | null,
+  handleUpdateTodo: (
+    todoId: number,
+    value: boolean | string,
+  ) => void,
+  isProcessedIds: number[],
+};
+
+export const TodoList: React.FC<Props> = ({
```
ArturErmolenko, line 18: memo

ArturErmolenko (changes requested):

src/App.tsx
Coyoote: I'm not sure if this is correct, but in my case the Loader shows up every time when updating a todo (changing title or status)

ArturErmolenko: @Coyoote that's because you should return all components when loading false:
```
  {isLoading
         ? <Loader />
         : (
           <div className="todoapp__content">
             <Header />

             <TodoList />

             {!!todos.length && (
               <Footer />
             )}
           </div>
         )}
```

ArturErmolenko (approved):
GJ 👍
Just  fix Loader I replied you in comment

PR #362, opened 2023-02-22
Todo app 3/3

andrii-andriushchenko:
[DEMO LINK](https://andrii-andriushchenko.github.io/react_todo-app-with-api/)

Have a good mood! 🌞

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,204 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
+import { Todo } from './types/Todo';
 
-const USER_ID = 0;
+import {
+  addTodo, deleteTodo, getTodos, updateTodo,
+} from './api/todos';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { FilterType } from './types/FilterType';
+import { Footer } from './components/Footer';
+import { Notification } from './components/Notification';
+import { getFilteredTodos } from './utils/getFilteredTodos';
+import { UserWarning } from './UserWarning';
+import { getActiveTodos } from './utils/getActiveTodos';
+import { getCompletedTodos } from './utils/getCompletedTodos';
+import { ErrorType } from './types/ErrorType';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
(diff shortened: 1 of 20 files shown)


Serghii29 (commented):
Good job🙌

src/App.tsx
```diff
@@ -1,24 +1,228 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
```
alexandra-protyanova, line 3:
```suggestion
import React, {
  useCallback, 
  useEffect, 
  useMemo, 
  useState,
} from 'react';
```

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+      const todosFromServer = await getTodos(userId);
+
+      setTodos(todosFromServer);
+    } catch {
+      setHasError(true);
+      setError(ErrorType.UPLOAD_ERROR);
+      // eslint-disable-next-line no-console
+      console.warn('An occur error while loading todos');
+    }
+  };
```
alexandra-protyanova, line 42: let us add loader

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+  const handleAddTodo = async (userId: number, title: string) => {
+    if (!title.trim()) {
+      setHasError(true);
+      setError(ErrorType.EMPTY_ERROR);
+
+      return;
+    }
+
+    const newTodo = {
+      id: 0,
```
alexandra-protyanova, line 61: no no no, `0` isn't valid value for id, you shouldn't send id for todo on server

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+      }
+    });
+  }, [todos]);
+
+  const changeTodo = async (todo: Todo, title?: string) => {
+    if (todo.title === title) {
+      return;
+    }
+
+    if (title?.length === 0) {
```
alexandra-protyanova, line 115:
```suggestion
    if (!title?.length) {
```

alexandra-protyanova (changes requested):

src/App.tsx
andrii-andriushchenko: Done 👍

src/App.tsx
andrii-andriushchenko: I left `title?.length === 0` because title can be `undefined`, and `!title.length` would work incorrectly, so I needed to add an additional check for `undefined`.
I think first variant is more clear)

ArturErmolenko (approved):
Well done 🔥

PR #365, opened 2023-02-22
Solution

halynarydasova:
[DEMO LINK](https://halynarydasova.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,185 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, {
+  useContext,
+  useEffect,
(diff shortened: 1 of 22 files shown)


src/App.tsx
```diff
@@ -1,24 +1,185 @@
...
+import { TodoStatus } from './types/TodoStatus';
+import { Footer } from './components/Footer';
+import { TodoItem } from './components/Todo';
+import { ErrorMessages } from './types/ErrorMessages';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
qoa11a, line 22: Don't disable eslint. Fix the error itself rather than disabling the rule.

qoa11a (approved):
Good job! 👍 
One small comment, but I still approve. You don't need to re-request a review when you fix it.

PR #372, opened 2023-02-23
realised react_todo-app with api

andrii-kurash:
[DEMO LINK](https://andrii-kurash.github.io/react_todo-app-with-api/)

Have a good time🔥

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,190 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 
-const USER_ID = 0;
+import { getTodos, deleteTodos, updateTodoOnServer } from './api/todos';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Filter } from './components/Filter';
+import { Error } from './components/Error';
+import { Todo, TodoStatus } from './types/Todo';
+import { ErrorMessage } from './types/ErrorMessage';
+import { filterTodos } from './components/FilterTodo';
+import { UserIdContext } from './utils/context';
+import { ChangeFunction } from './types/ChangeFunction';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedFilter, setSelectedFilter] = useState(TodoStatus.All);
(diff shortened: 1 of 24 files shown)


src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+
+  const showError = useCallback((error: ErrorMessage) => {
+    setErrorType(error);
+  }, []);
+
+  const hideError = useCallback(() => {
+    setErrorType(ErrorMessage.None);
+  }, []);
+
+  const AddTodo = useCallback(
```
Anna-Tkachenko, line 60: don't start variable name for common values from capital letter
```suggestion
  const addTodo = useCallback(
```

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,189 @@
...
+    setIsWaiting(true);
+
+    try {
+      await deleteTodos(id);
+      DeleteTodo(id);
+    } catch {
+      showError(ErrorMessage.Delete);
+      setIsWaiting(false);
+
+      onError?.();
```
Anna-Tkachenko, line 50: will be better to set some boolean variable and depending on it call / not call some function

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,189 @@
...
+              }}
+              onDoubleClick={() => setIsEdit(true)}
+            >
+              {title}
+            </span>
+
+            <button
+              type="button"
+              className="todo__remove"
+              onClick={() => handleDeleteTodo()}
```
Anna-Tkachenko, line 171: do you need wrapper here?

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,88 @@
...
+  isClearCompleted,
+  DeleteTodo,
+  ChangeTodo,
+  showError,
+  hideError,
+  isAllToggled,
+}) => {
+  const userId = useContext(UserIdContext);
+
+  const creatingTodo = useMemo(
```
Anna-Tkachenko, line 34: check variable name, if it's todo let's name it as `createdTodo`

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,88 @@
...
+    () => ({
+      id: 0,
+      userId,
+      title: creatingTodoTitle,
+      completed: false,
+    }),
+    [userId, creatingTodoTitle],
+  );
+
+  const isLoading = (isTodoCompleted: boolean): boolean => {
```
Anna-Tkachenko, line 44: `isLoading` - not the best name for function
function name should contain some action, `getIsLoading`, for example

Anna-Tkachenko (changes requested):
When I open the page - I see infinite loader in todo list, check it

PR #376, opened 2023-02-23
Solution

max-rozzhalovets:


src/App.tsx
```diff
@@ -1,24 +1,156 @@
...
+      userId: USER_ID,
+      completed: false,
+    };
+
+    setTempTodo({ ...todoToAdd, id: 0 });
+
+    try {
+      const response = await addTodo(USER_ID, todoToAdd);
+
+      setTodos(currentTodos => [...currentTodos, response]);
```
Anna-Tkachenko, line 64: will be better to make request here to api for all todos

src/App.tsx
```diff
@@ -1,24 +1,156 @@
...
+    }
+  };
+
+  const deleteChosenTodo = async (todoId: number) => {
+    try {
+      await deleteTodo(USER_ID, todoId);
+
+      setTodos((currentTodos) => (
+        currentTodos.filter(({ id }) => id !== todoId)
+      ));
```
Anna-Tkachenko, line 78: same here
after each change will be better to get all updated data from api

src/App.tsx
```diff
@@ -1,24 +1,156 @@
...
+              onDelete={deleteChosenTodo}
+              onChange={changeChosenTodo}
+              tempTodo={tempTodo}
+            />
+          )
+        }
+
+        {
+          todos.length !== 0 && (
+            <Footer
```
Anna-Tkachenko, line 141: can we combine these conditions?

src/components/TodoComponent.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const handleTitleDoubleClick = (
+    event: React.MouseEvent<HTMLSpanElement>,
+  ) => {
+    switch (event.detail) {
+      case 2:
+        setIsInputVisible(true);
+        break;
+      default:
+        break;
+    }
```
Anna-Tkachenko, line 54: do you need switch / case here?

src/components/todosList.tsx
```diff
@@ -0,0 +1,111 @@
...
+              classNames="temp-item"
+            >
+              <TodoComponent todo={tempTodo} />
+            </CSSTransition>
+          )
+        }
+
+      </TransitionGroup>
+
+      {/* This todo is not completed */}
```
Anna-Tkachenko, line 51: delete unused comments

Anna-Tkachenko (changes requested):
- Let's add loader on each todo while processing changing status for all todos

PR #378, opened 2023-02-23
solution

victorionov:
- [DEMO LINK](https://victorionov.github.io/react_todo-app-with-api)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { Todo } from './types/Todo';
 import { Header } from './Components/Header';
 import { TodoList } from './Components/Todolist';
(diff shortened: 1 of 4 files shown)


Serghii29 (changes requested):
Good job!

but your app does not work correctly. They are not updated as required and the footer does not work as required. review your code again and test for errors

src/App.tsx
```diff
@@ -1,24 +1,193 @@
...
+    setErrorMessage(error);
+  };
+
+  const clearError = () => {
+    setHasError(false);
+  };
+
+  const fetchTodos = async () => {
+    try {
+      setHasError(false);
```
Anna-Tkachenko, line 34: should we change `hasError` variable in `catch` block?

src/App.tsx
```diff
@@ -1,24 +1,193 @@
...
+    const newTodo: Todo = {
+      id: 0,
+      userId: USER_ID,
+      title: inputQuery,
+      completed: false,
+    };
+
+    setTodosToShow([
+      ...todosToShow,
+      newTodo]);
```
Anna-Tkachenko, line 66:
```suggestion
    setTodosToShow([
      ...todosToShow,
      newTodo,
    ]);
```

Anna-Tkachenko (changes requested):
Check the behaviour of loader component, for now I don't see it's usage. It should be visible when something is processing with todos: updating (editing) text or status and so on.
Also, when I click on `clear completed` button - there are no any changes, check it

denys-cheporniuk (changes requested):
Uploading Screen Recording 2023-03-06 at 13.14.23.mov…

editing should work by double click, in your case somehow it works by triple click
Also, after editing I see todo only with spaces

PR #382, opened 2023-02-25
solution

Oleksandr-Maly:
[DEMO LINK](https://oleksandr-maly.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,172 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { CSSTransition, TransitionGroup } from 'react-transition-group';
+import {
+  addTodos,
+  deleteTodos,
+  editTodos,
+  getTodos,
+} from './api/todos';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { Notification } from './components/Notification';
+import { TodoItem } from './components/TodoItem';
+import { Errors } from './types/Errors';
+import { Todo } from './types/Todo';
+import { TodoStatus } from './types/TodoStatus';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 6398;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedStatus, setSelectedStatus] = useState(TodoStatus.All);
+  const [errorType, setErrorType] = useState(Errors.None);
+  const [hasErrorNotification, setHasErrorNotification] = useState(false);
+  const [isUpdatingIds, setIsUpdatingIds] = useState<number[]>([]);
+
+  const hideErrorNotifications = () => setHasErrorNotification(false);
+  const showErrorNotification = (error:Errors) => {
+    setErrorType(error);
+    setHasErrorNotification(true);
+    setTimeout(hideErrorNotifications, 3000);
+  };
+
+  const fetchAllTodos = async () => {
+    try {
+      const todosFromServer = await getTodos(USER_ID);
+
+      setTodos(todosFromServer);
+    } catch (error) {
+      showErrorNotification(Errors.Get);
+    }
+  };
+
+  useEffect(() => {
+    fetchAllTodos();
+  }, []);
+
+  const addNewTodo = async (title: string) => {
+    try {
+      const newTodo = {
+        title,
+        userId: USER_ID,
+        completed: false,
+      };
+
+      await addTodos(newTodo);
+      await fetchAllTodos();
+    } catch {
+      showErrorNotification(Errors.Add);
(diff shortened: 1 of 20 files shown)


src/App.tsx
```diff
@@ -1,24 +1,172 @@
...
 
-const USER_ID = 0;
+const USER_ID = 6398;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedStatus, setSelectedStatus] = useState(TodoStatus.All);
+  const [errorType, setErrorType] = useState(Errors.None);
+  const [hasErrorNotification, setHasErrorNotification] = useState(false);
+  const [isUpdatingIds, setIsUpdatingIds] = useState<number[]>([]);
```
Anna-Tkachenko, line 26: probably some boolean variables can start with `is`
let's think more about variable name for this ids

src/App.tsx
```diff
@@ -1,24 +1,172 @@
...
+      // setIsLoaded(true);
+      setIsUpdatingIds([]);
+    } catch (error) {
+      showErrorNotification(Errors.Edit);
+    }
+  };
+
+  let visibleTodos = [...todos];
+
+  if (selectedStatus === TodoStatus.Active) {
```
Anna-Tkachenko, line 97: if you need to process some logic according to the value of some variable, will be better to use switch / case
and let's move the code for preparing todos in separate function

src/components/Footer/Foter.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <span className="todo-count">
+        {`${activeTodosCount} items left`}
+      </span>
+
+      <StatusFilter
+        selectStatus={selectStatus}
+        selectedStatus={selectedStatus}
+      />
+
+      {/* don't show this button if there are no completed todos */}
```
Anna-Tkachenko, line 29: check the comment

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,39 @@
...
+      message = '';
+  }
+
+  return (
+    <div className="notification is-danger is-light has-text-weight-normal">
+      <button
+        aria-label="close notification"
+        type="button"
+        className="delete"
+        onClick={() => hideErrorNotifications()}
```
Anna-Tkachenko, line 34: do you need wrapper here?

Anna-Tkachenko (changes requested):
Check the functionality:
- we shouldn't be able to add empty todo (with empty title)
- when I click on `completed` button (if there is no completed todos) - I can't return back to the active one, I didn't see buttons and filters

src/App.tsx
```diff
@@ -1,24 +1,180 @@
...
+  const addNewTodo = async (title: string) => {
+    if (title.trim()) {
+      try {
+        const newTodo = {
+          title,
+          userId: USER_ID,
+          completed: false,
+        };
+
+        await addTodos(newTodo);
```
Anna-Tkachenko, line 74: check in previous task description how should work the process of adding new todos, you didn't complete all requirements (check points with temp todo)

src/App.tsx
```diff
@@ -1,24 +1,180 @@
...
+    try {
+      await editTodos(todo.id, editedTodo);
+      await fetchAllTodos();
+      setUpdatingTodoIds([]);
+    } catch (error) {
+      showErrorNotification(Errors.Edit);
+    }
+  };
+
+  const visibleTodos = filterTodos(todos, selectedStatus);
```
Anna-Tkachenko, line 114: don't forget about rendering optimization hooks

src/components/AddTodoForm/AddTodoForm.tsx
```diff
@@ -0,0 +1,46 @@
...
+    setTodoTitle('');
+  };
+
+  const handleSubmit = (event:React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    setHasTitleError(!todoTitle);
+
+    if (!hasTitleError) {
+      addNewTodo(todoTitle);
```
Anna-Tkachenko, line 27: should we show some loader with temp todo here? for now as user I don't see that something is processing after form submitting

src/components/Footer/Foter.tsx
Anna-Tkachenko: check the comment

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,24 +1,200 @@
...
+  const addNewTodo = useCallback(async (title: string) => {
+    if (title.trim()) {
+      try {
+        const newTodo = {
+          title,
+          userId: USER_ID,
+          completed: false,
+        };
+
+        setTempTodo({ ...newTodo, id: 0 });
```
Anna-Tkachenko, line 80: will be great if user will see this temp todo immediately in the todo list (as it's described in the task)

Anna-Tkachenko (approved):

PR #386, opened 2023-02-28
Final

IoghiStudio:
[DEMO LINK](https://IoghiStudio>.github.io/react_todo-app-with-api/)
Still working on it.


src/components/App/App.tsx
```diff
@@ -0,0 +1,215 @@
...
+  //       .catch(() => {
+  //         setError('Unable to update a todo');
+  //         setTimeout(() => setError(''), 3000);
+  //       })
+  //       .finally(() => {
+
+  //       });
+  //   },
+  //   [],
+  // );
```
naziamoff, line 129: don't leave unnecessary comments in your code

naziamoff (changes requested):
looks good, but fix errors from the previous PR. You can fix them here, no need to do it in the previous PR

PR #389, opened 2023-03-04
solution

Deniskakaka:
https://deniskakaka.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,153 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
+import { v4 as uuid } from 'uuid';
+import { deleteTodo, getTodos, postTodo } from './api/todos';
+import { Error } from './components/Error/Error';
+import { Filter } from './components/Filter/Filter';
+import { Item } from './components/Todo/Todo';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 6470;
 
 export const App: React.FC = () => {
+  const [titleTodo, setTitleTodo] = useState('');
+  const [listTodo, setListTodo] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [tempTodo, setTempTodo] = useState(false);
+
+  const getListTodo = async (filter?: boolean, type?: string) => {
+    try {
+      const result = await getTodos(USER_ID, filter);
+
+      if (type === 'active') {
+        setListTodo(result.filter((el:Todo) => !el.completed));
+      } else {
+        setListTodo(result);
+      }
+    } catch (e) {
+      setError('Oops, something were wrong, please try again later');
+    }
+  };
+
+  useEffect(() => {
+    getListTodo();
+  }, []);
+
+  const clearError = useCallback(() => setError(''), []);
+
+  const onSubmitHandler = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (titleTodo) {
+      setTempTodo(true);
+      try {
+        await postTodo(
+          { title: titleTodo, userId: USER_ID, completed: false },
+        );
+
+        getListTodo();
+        setTitleTodo('');
+        setTempTodo(false);
+      } catch (e) {
+        setError('Oops, something were wrong, please try again later');
+      }
+    }
+  };
+
+  const onChangeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitleTodo(event.target.value);
+  };
+
+  const deleteTodoHandler = (id: number) => {
+    deleteTodo(id).then(() => getListTodo());
+  };
+
+  const deleteAllCompleteTodo = () => {
+    listTodo.forEach((todo: Todo) => {
+      if (todo.completed) {
+        deleteTodoHandler(todo.id);
+      }
+    });
+  };
+
+  const renderClearButton = useMemo(() => {
+    if (listTodo.some((todo: Todo) => todo.completed)) {
+      return (
+        <button
+          type="button"
+          className="todoapp__clear-completed"
+          onClick={deleteAllCompleteTodo}
+        >
+          Clear completed
+        </button>
+      );
+    }
+
+    return <></>;
+  }, [listTodo]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          <button
+            type="button"
+            className="todoapp__toggle-all active"
+          />
+          <form onSubmit={onSubmitHandler}>
+            <input
+              type="text"
+              className="todoapp__new-todo"
+              placeholder="What needs to be done?"
+              value={titleTodo}
+              onChange={onChangeHandler}
+            />
+          </form>
+        </header>
+
+        <section className="todoapp__main">
+          {
+            listTodo.map((todo: Todo) => (
+              <Item
+                todo={todo}
+                key={todo.id}
+                deleteTodo={deleteTodoHandler}
+                updateTodos={getListTodo}
+              />
+            ))
+          }
+          {tempTodo && (
+            <Item
+              todo={{
+                id: +uuid(),
+                userId: USER_ID,
+                title: titleTodo,
+                completed: false,
+              }}
+              temp
+            />
+          )}
+        </section>
+        <footer className="todoapp__footer">
+          <span className="todo-count">
+            {`${listTodo.length} items left`}
+          </span>
+
+          <Filter setFilter={getListTodo} />
+
+          {renderClearButton}
+        </footer>
+      </div>
+
+      {error && <Error errorText={error} errorClear={clearError} />}
+    </div>
   );
 };
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -1,0 +1,26 @@
+import { useEffect, FC } from 'react';
+
+/* eslint-disable jsx-a11y/control-has-associated-label */
+type Props = {
+  errorText: string,
+  errorClear: () => void
+};
+export const Error: FC<Props> = ({ errorClear, errorText }) => {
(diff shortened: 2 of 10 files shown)


VitaliyBondarenko1982 (changes requested):
In this PR exist same problems? as in [
react_todo-app-add-and-delete](https://github.com/mate-academy/react_todo-app-add-and-delete/pull/328)

- Don't show the footer if there are no todos in the list. Also, don't show the clear completed button if there are no completed todos
-  todo disappears for a moment before adding
- todo should disappear from the list after the error
-  strange behavior for the input and todo item (when slow internet, todo loading, but you can change input and adding todo in same time)
- show some error for the deleting process as well
- The items left amount shouldn't change from filter to filter

src/App.tsx
```diff
@@ -1,24 +1,173 @@
...
+              }}
+              temp
+            />
+          )}
+        </section>
+
+        {showFooter ? (
+          <footer className="todoapp__footer">
+            <span className="todo-count">
+              {`${length} items left`}
```
etojeDenys, line 160: i think we can manually calculate the length here instead of using useState

src/App.tsx
```diff
@@ -1,24 +1,173 @@
...
+
+  useEffect(() => {
+    getTodos(USER_ID).then(res => {
+      setLength(res.length);
+      if (!res.length) {
+        setShowFooter(false);
+      } else {
+        setShowFooter(true);
+      }
+    });
```
etojeDenys, line 49: why do you need this second useEffect if you call getTodos in `getListTodo`

src/App.tsx
```diff
@@ -1,24 +1,173 @@
...
+  }, [listTodo]);
+
+  const clearError = useCallback(() => {
+    setTempTodo(false);
+    setError('');
+  }, []);
```
etojeDenys, line 55: we don`t need to cover such small function in useCallback

src/App.tsx
```diff
@@ -1,24 +1,173 @@
...
+
+  const onChangeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitleTodo(event.target.value);
+  };
+
+  const deleteTodoHandler = (id: number) => {
+    deleteTodo(id).then(() => getListTodo());
+  };
+
+  const deleteAllCompleteTodo = () => {
```
etojeDenys, line 83:
```suggestion
  const deleteAllCompletedTodos = () => {
```

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,26 @@
...
+  errorClear: () => void
+};
+export const Error: FC<Props> = ({ errorClear, errorText }) => {
+  useEffect(() => {
+    const idTimer = setTimeout(errorClear, 3000);
+
+    if (errorText) {
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      idTimer;
+    }
```
etojeDenys, line 15: can we remove this?

src/App.tsx
```diff
@@ -1,24 +1,173 @@
...
+  const [error, setError] = useState('');
+  const [tempTodo, setTempTodo] = useState(false);
+  const [showFooter, setShowFooter] = useState(true);
+  const [length, setLength] = useState(0);
+
+  const getListTodo = async (filter?: boolean, type?: string) => {
+    try {
+      const result = await getTodos(USER_ID, filter);
+
+      if (type === 'active') {
```
etojeDenys, line 27: why you don't use here TypeFilter enum?

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,57 @@
+import classNames from 'classnames';
+import { useState } from 'react';
+
+type Props = {
+  setFilter: (filter?: boolean, type?: string) => void
```
etojeDenys, line 5:
```suggestion
  setFilter: (filter?: boolean, type?: TypeFilter) => void
```

src/App.tsx
```diff
@@ -1,24 +1,173 @@
...
 
 export const App: React.FC = () => {
+  const [titleTodo, setTitleTodo] = useState('');
+  const [listTodo, setListTodo] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [tempTodo, setTempTodo] = useState(false);
+  const [showFooter, setShowFooter] = useState(true);
+  const [length, setLength] = useState(0);
+
+  const getListTodo = async (filter?: boolean, type?: string) => {
```
etojeDenys, line 23: we can just check if `type !== TypeFilter.All`, and don't use filter parameter

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <nav className="filter">
+      <a
+        href="#/"
+        className={classNames('filter__link', {
+          selected: typeFilter === TypeFilter[0],
+        })}
+        onClick={() => chooseFilter(TypeFilter[0])}
+      >
+        All
```
etojeDenys, line 32:
```suggestion
        {TypeFilter.All}
```

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,57 @@
...
+  };
+
+  return (
+    <nav className="filter">
+      <a
+        href="#/"
+        className={classNames('filter__link', {
+          selected: typeFilter === TypeFilter[0],
+        })}
+        onClick={() => chooseFilter(TypeFilter[0])}
```
etojeDenys, line 30:
```suggestion
        onClick={() => chooseFilter(TypeFilter.All)}
```

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,123 @@
...
+          updateTodos(false, typeFilter);
+        }
+
+        setLoader(false);
+        setChange(false);
+      });
+    }
+  };
+
+  const changeTitleTodoHandler = (item: Todo, event: any) => {
```
etojeDenys, line 53: we can just get this `todo`(first argument) without passing it as parameter

and don't forget to add some type for event

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,123 @@
...
+    <div className={classNames('todo', {
+      completed: todo.completed,
+    })}
+    >
+      <label className="todo__status-label">
+        <input
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+          onClick={() => changeTodoCompleteHandler(todo, false)}
```
etojeDenys, line 74: same about this function

etojeDenys (changes requested):
feel free to ask me about it in the slack chat 

still can reproduce:
- todo disappears for a moment before adding
- The items left amount shouldn't change from filter to filter

src/App.tsx
Deniskakaka: because like this i count all elements in getListTodo i have only todo which i choose in filter, i make request on back where choose type filter

src/App.tsx
Deniskakaka: Are you serious? How am I supposed to guess who's going to check my problem? This requirement was given to me in the first task from there I copied it in this form by adding setTempTodo

src/components/Error/Error.tsx
Deniskakaka: i try but linter Swearing Expected an assignment or function call and instead saw an expression, and since it is a timer I can not call it as a function other than to ignore this line for him I could not find, well, unless you go to the settings linter but I think this is not a good idea

src/App.tsx
Deniskakaka: i try, but how i wrote down, in second useEffect i count all elements, i can create async function and add in html, but why if i can just in one place count all elements and check need render footer or no

src/components/Todo/Todo.tsx
Deniskakaka: if i get todo not how arguments i can't return old value title if push esc

src/App.tsx
```diff
@@ -1,24 +1,173 @@
...
+                userId: USER_ID,
+                title: titleTodo,
+                completed: false,
+              }}
+              temp
+            />
+          )}
+        </section>
+
+        {showFooter ? (
```
Viktor-Kost, line 157: you can use `length` instead of using another statement as `showFooter`

src/App.tsx
Viktor-Kost: isn't fixed

src/components/Error/Error.tsx
Viktor-Kost: What if you use like this? Just as an attempt
```
useEffect(() => {
  let idTimer: ReturnType<typeof setTimeout>;

  if (errorText) {
    idTimer = setTimeout(errorClear, 3000);
  }

  return () => clearTimeout(idTimer);
}, [errorText, errorClear]);
```

or even like this

```
const clearError = () => {
  if (errorText) {
    errorClear();
  }
};

useEffect(() => {
  const idTimer = setTimeout(clearError, 3000);

  return () => clearTimeout(idTimer);
}, [errorText, errorClear]);
```

src/App.tsx
Viktor-Kost: It looks a little bit confused. You should have one `useEffect` hook for getting todos. If you have some dynamic elements like filter_type or something you can use function params or hook dependencies

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const [title, setTitle] = useState(todo.title);
+
+  const activeLoader = () => {
+    if (deleteTodo) {
+      deleteTodo(todo.id);
+      setLoader(true);
+    }
+  };
+
+  const changeTodoCompleteHandler = (changeName:boolean) => {
```
Viktor-Kost, line 27: It is better to name like this if you use a boolean
```suggestion
  const changeTodoCompleteHandler = (isChangedName:boolean) => {
```

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const result = todo;
+
+    setLoader(true);
+    if (!changeName) {
+      result.completed = !todo.completed;
+    } else {
+      result.title = title;
+    }
```
Viktor-Kost, line 35: ternary operator

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,127 @@
...
+  temp?: boolean,
+  deleteTodo?: (id: number) => void,
+  updateTodos?: (type?: string) => void
+};
+
+export const Item: React.FC<Props> = ({
+  todo, temp, deleteTodo, updateTodos,
+}) => {
+  const [loader, setLoader] = useState(false);
+  const [change, setChange] = useState(false);
```
Viktor-Kost, line 17: isLoading
isInputChanging

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const result = todo;
+
+    setLoader(true);
+    if (!changeName) {
+      result.completed = !todo.completed;
+    } else {
+      result.title = title;
+    }
+
+    if (updateTodos) {
```
Viktor-Kost, line 37: you've `setLoader(true) `previously. What will happen if you don't have `updateTodos` ? 
You need to control your loader

PR #398, opened 2023-03-15
Solution p3 need he

KrisMakarovska:
[ DEMO LINK](KrisMakarovska.github.io/react_todo-app-with-api/)
 
 Please help to fix next issues:
 1. onBlur and onSubmit title changes: if the new title is the same as the old one just cancel editing. The editing id was canceled but the request is still running on the server. How to cancel the request? Also, the markup looks strange 
 
 2. I can’t activate and deactivate the loader on several todos. Only on all of them. (Use state ‘loader’). It happens in 3 cases:
 - when I click the toggle all button on several active todos  -  when I click on the checkbox on some todo. - when I submit title changes 
 
 3. I can’t make my form not on focus while I delete empty todo.

### DIFF
--- a/src/TodoInfo.tsx
+++ b/src/TodoInfo.tsx
@@ -1,0 +1,123 @@
+import classnames from 'classnames';
+import React, { useState } from 'react';
+import { Todo } from './types/Todo';
+
+type Props = {
+  todo: Todo,
+  onDelete: (id: number) => void,
+  onStatusChange: (todo: Todo) => void,
+  onTitleChange: (todo: Todo, todoTitle: string) => void,
+  loader: boolean,
+  // setLoader: (loader: boolean) => void,
+};
+
+export const TodoInfo: React.FC<Props> = ({
+  todo,
+  onDelete,
+  onStatusChange,
+  onTitleChange,
+  loader,
+  // setLoader,
+}) => {
+  const { id, title, completed } = todo;
+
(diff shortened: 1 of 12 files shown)


Volodymyr-Melnyk-Fron10der (changes requested):
Hi, Khrystyna! Please ask all your questions in the Slack as it's easier for communication and you will get all answers much faster🙌🏼

src/TodoInfo.tsx
```diff
@@ -0,0 +1,119 @@
...
+export const TodoInfo: React.FC<Props> = ({
+  todo,
+  onDelete,
+  onStatusChange,
+  onTitleChange,
+  isLoading,
+}) => {
+  const { id, title, completed } = todo;
+
+  // const [isBeingDeleted, setIsBeingDeleted] = useState(false);
```
polosanya, line 22: Remove the comments

src/TodoInfo.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const [titleChanger, setTitleChanger] = useState(false);
+  const [newTitle, setNewTitle] = useState(title);
+  const [focusDisable, setFocusDisable] = useState(false);
```
polosanya, line 25: Get used to name your boolean state according to these rules
https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

polosanya (approved):

PR #399, opened 2023-03-16
Third step, first try

Vasyl-Pavlenko:
[DEMO LINK](https://vasyl-pavlenko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,165 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
+import {
+  getTodos,
+  createTodo,
+  deleteTodo,
+  updateTodo,
+} from './api/todos';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { TodoList } from './components/TodoList';
+import { FilteredBy } from './types/FilteredBy';
+import { Todo } from './types/Todo';
+import { TodoNotification } from './components/TodoNotification';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 6662;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isTodoLoaded, setIsTodoLoaded] = useState(false);
+  const [filterBy, setFilterBy] = useState<FilteredBy>(FilteredBy.ALL);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isError, setIsError] = useState(false);
+
+  const isActiveTodos = todos.some(todo => todo.completed);
+  const countActiveTodos = todos.filter(todo => !todo.completed).length;
+
+  const getTodosFromServer = async () => {
+    try {
+      const todosFromServer = await getTodos(USER_ID);
+
+      setTempTodo(null);
+      setIsTodoLoaded(true);
+      setTodos(todosFromServer);
+    } catch (error) {
+      setErrorMessage('Data couldn\'t be loaded from the server');
+    }
+  };
+
+  const createTodoOnServer = async (query: string) => {
+    const data = {
+      title: query,
+      userId: USER_ID,
+      completed: false,
+    };
+
+    setIsTodoLoaded(false);
+    setTempTodo({
+      id: 0,
+      ...data,
+    });
+    try {
+      await createTodo(USER_ID, data);
+      getTodosFromServer();
+    } catch (error) {
+      setTempTodo(null);
+      setErrorMessage('Unable to add a todo');
+    }
+  };
+
+  const removeTodoOnServer = async (id: number) => {
+    try {
+      await deleteTodo(id);
+      getTodosFromServer();
+    } catch (error) {
+      setTempTodo(null);
+      setErrorMessage('Unable to delete a todo');
+    }
+  };
+
+  const updateTodoOnServer = async (todo: Todo, isChange?: boolean) => {
+    const todoId = todo.id;
+    const todoCopy = {
+      ...todo,
+    };
+
+    if (isChange === true) {
+      todoCopy.completed = !todo.completed;
+    }
+
+    try {
+      await updateTodo(todoId, todoCopy);
+      getTodosFromServer();
+    } catch (error) {
+      setTempTodo(null);
+      setErrorMessage('Unable to update a todo');
+    }
+  };
+
+  const updateAllTodos = async () => {
+    todos.forEach(todo => {
+      updateTodoOnServer(todo, true);
+    });
+  };
+
+  const clearCompletedTodo = async () => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    completedTodos.forEach(todo => removeTodoOnServer(todo.id));
+  };
+
+  const visibleTodos: Todo[] = useMemo(() => {
+    switch (filterBy) {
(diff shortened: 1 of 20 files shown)


src/App.tsx
```diff
@@ -1,24 +1,165 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isTodoLoaded, setIsTodoLoaded] = useState(false);
+  const [filterBy, setFilterBy] = useState<FilteredBy>(FilteredBy.ALL);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isError, setIsError] = useState(false);
+
+  const isActiveTodos = todos.some(todo => todo.completed);
```
mykhalenych, line 26: change to useMemo

src/App.tsx
```diff
@@ -1,24 +1,165 @@
...
+  const countActiveTodos = todos.filter(todo => !todo.completed).length;
```
mykhalenych, line 27: here as well

mykhalenych (changes requested):
Screenshot 2023-03-16 at 17 11 14

your loader doen't work correctly, I got an error, and infinity loader

Vasyl-Pavlenko:
I added useMemo.

Also i added errorMessage for case offline, but in the same time i couldn't understand how to make my input not disabled?

I tried to add isLoader(false) into my catch. It's doesn't work

src/App.tsx
```diff
@@ -1,24 +1,176 @@
...
+
+  const getTodosFromServer = async () => {
+    try {
+      const todosFromServer = await getTodos(USER_ID);
+
+      setTempTodo(null);
+      setIsTodoLoaded(true);
+      setTodos(todosFromServer);
+    } catch (error) {
+      setIsError(true);
```
loralevitska, line 44: I guess it will fix your infinit loaders. Fix loading everywhere
```suggestion
      setIsError(true);
      setIsTodoLoaded(false);
```

src/App.tsx
```diff
@@ -1,24 +1,176 @@
...
+  const visibleTodos: Todo[] = useMemo(() => {
+    switch (filterBy) {
+      case FilteredBy.ACTIVE:
+        return todos.filter(todo => !todo.completed);
+      case FilteredBy.COMPLETED:
+        return todos.filter(todo => todo.completed);
+      case FilteredBy.ALL:
+        return [...todos];
+      default:
+        throw new Error('Unexpected filter type');
```
loralevitska, line 123:
```suggestion
    switch (filterBy) {
      case FilteredBy.ACTIVE:
        return todos.filter(todo => !todo.completed);
        
      case FilteredBy.COMPLETED:
        return todos.filter(todo => todo.completed);
        
      case FilteredBy.ALL:
        return [...todos];
        
      default:
        throw new Error('Unexpected filter type');
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,74 @@
...
+  updateAllTodos,
+  setErrorMessage,
+  setIsError,
+}) => {
+  const [query, setQuery] = useState<string>('');
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (query.length === 0) {
```
loralevitska, line 29:
```suggestion
    if (!query.length) {
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+    }
+  };
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  return (
+    <div
+      className={classNames('todo', { completed: completed === true })}
```
loralevitska, line 83:
```suggestion
      className={classNames('todo', { completed: !!completed })}
```

OR
If completed always boolean
```suggestion
      className={classNames('todo', { completed })}
```

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
```
loralevitska, line 4: Remove comments everywhere. Try it write obvious code that explain all we need without comments)

loralevitska (changes requested):

- After error disable loading 
image
- Remove default outline 
image
- Your toggleAll doesn't work correct 
- Every adding new todo fetch 2 queris. Should be 1 query per todo. Fix it

If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

src/App.tsx
```diff
@@ -1,24 +1,192 @@
...
+    };
+
+    setTempTodo({
+      ...data,
+      id: 0,
+    });
+    try {
+      setIsTodoLoaded(true);
+      await createTodo(USER_ID, data);
+      await getTodosFromServer();
```
DarkMistyRoom, line 68: would be better for performance to reduce requests on the server and update the array on the place. Same for all other manipulations with todos

src/App.tsx
```diff
@@ -1,24 +1,192 @@
...
+    };
+
+    if (isChange) {
+      todoCopy.completed = !todo.completed;
+    }
+
+    try {
+      setIsTodoLoaded(true);
+      await updateTodo(todoId, todoCopy);
+      getTodosFromServer();
```
DarkMistyRoom, line 105:
```suggestion
      await updateTodo(todoId, todoCopy);
      await getTodosFromServer();
```

src/App.tsx
```diff
@@ -1,24 +1,192 @@
...
+    todos.forEach(todo => {
+      updateTodoOnServer(todo, true);
+    });
+  };
+
+  const clearCompletedTodo = async () => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    completedTodos.forEach(todo => removeTodoOnServer(todo.id));
+  };
```
DarkMistyRoom, line 125: what about Loader? you should show it while the request is handling

src/types/SendTodo.ts
```diff
@@ -0,0 +1,5 @@
+export interface SendTodo {
+  userId: number,
+  title: string,
+  completed: boolean,
+}
```
DarkMistyRoom, line 5: you can place it in the Todo.ts

DarkMistyRoom (changes requested):
- I've just added todo with only whitespaces

image

- Also would be great to add focusing on input on editing

image

- Check how toggle all works in the example page

src/App.tsx
```diff
@@ -1,24 +1,219 @@
...
+    };
+
+    if (isChange) {
+      todoCopy.completed = !todo.completed;
+    }
+
+    try {
+      setIsTodoLoaded(true);
+      await updateTodo(todoId, todoCopy);
+      await getTodosFromServer();
```
IvanFesenko, line 108: This case will gives multiple request. Just update local state without request all todos

src/App.tsx
```diff
@@ -1,24 +1,219 @@
...
+  const visibleTodos: Todo[] = useMemo(() => {
+    switch (filterBy) {
+      case FilteredBy.ACTIVE:
+        return todos.filter(todo => !todo.completed);
+
+      case FilteredBy.COMPLETED:
+        return todos.filter(todo => todo.completed);
+
+      case FilteredBy.ALL:
+        return [...todos];
```
IvanFesenko, line 162: No need to create copy
```suggestion
        return todos;
```

IvanFesenko (approved):

PR #400, opened 2023-03-19
solution

bohdan-gavrilyk:
https://bohdan-gavrilyk.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,200 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  getTodos,
+  createTodo,
+  removeTodo,
+  changeTodo,
+}
+  from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { Filter } from './types/Filter';
+import { ErrorAction } from './types/ErrorAction';
+import { Notification } from './components/Notification';
 
-const USER_ID = 0;
+const USER_ID = 6655;
+
+function filteredTodos(todos:Todo[], filter:Filter) {
+  const returnArr = [...todos];
+
+  switch (filter) {
+    case Filter.ACTIVE:
+      return returnArr.filter(todo => !todo.completed);
+    case Filter.COMPLETED:
+      return returnArr.filter(todo => todo.completed);
+    default:
+      return returnArr;
+  }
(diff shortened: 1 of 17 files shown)


src/App.tsx
```diff
@@ -1,24 +1,200 @@
...
+      <div className="todoapp__content">
+        <Header
+          onSubmit={addTodo}
+          isDisabled={disabledInput}
+          errorInput={() => {
+            setErrorMessage(ErrorAction.TITLE);
+            setTimeout(() => {
+              setErrorMessage(ErrorAction.NONE);
+            }, 3000);
+          }}
```
VitaliyBondarenko1982, line 172: it is bad practise to make anonymous arrow functions in JSX
better to create function in component and pass reference here

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const handleSubmit = () => {
+    if (!query) {
+      errorInput();
+
+      return;
+    }
+
+    onSubmit(query);
+    setQuery('');
+  };
```
VitaliyBondarenko1982, line 30:
```suggestion
  const handleSubmit = (e: FormEvent<HTMLFormElement>) => {
   e.preventDefault();
   
    if (!query) {
      errorInput();

      return;
    }

    onSubmit(query);
    setQuery('');
  };
```
and just pass reference to onSubmit

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,63 @@
...
+          'todoapp__toggle-all',
+          { active: hidenToggle },
+        )}
+      />
+
+      <form onSubmit={(event) => {
+        event.preventDefault();
+        handleSubmit();
+      }}
+      >
```
VitaliyBondarenko1982, line 51:
```suggestion
      <form onSubmit={submitHandler}
      >
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <input
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          disabled={isDisabled}
+          value={query}
+          onChange={(event) => setQuery(event.target.value)}
```
VitaliyBondarenko1982, line 58: make named function and pass reference here
```suggestion
          onChange={(event) => setQuery(event.target.value)}
```

VitaliyBondarenko1982 (changes requested):
Looks good to me.
Suggest some little refactor.
Please check all cases with anonymous arrow functions in jsx

vlad-demchuk (approved):
Looks good

It would be good to apply some optimization like `memo` and `Promise.All` for bulk actions (`toogleAll, deleteTodoCompleted`)

PR #401, opened 2023-03-19
finished

barantarasNEW:
https://barantarasNEW.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,198 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 
-const USER_ID = 0;
+import {
+  createTodo,
+  getTodos,
+  deleteTodo,
+  USER_ID,
+  updateTodo,
+} from './api/todos';
+
+import { Errors } from './types/Errors';
+import { Filters } from './types/Filters';
+import { Todo } from './types/Todo';
+
+import Header from './Components/Header/Header';
+import Footer from './Components/Footer/Footer';
+import TodoList from './Components/TodoList/TodoList';
+import Notification from './Components/Notification/Notification';
+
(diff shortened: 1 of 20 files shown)


src/Components/Header/Header.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+      <form onSubmit={onSubmit}>
+        <input
+          ref={inputRef}
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          value={value}
+          onChange={e => setValue(e.target.value)}
+          onBlur={() => setSubmitState(Submits.none)}
```
VitaliyBondarenko1982, line 83: make functions onChange, onBlur and pass reference here

src/App.tsx
```diff
@@ -1,24 +1,198 @@
...
+    });
+
+    Promise.all([...delCompletedTodos])
+      .then(() => {
+        setTodos(currTodos => currTodos.filter(todo => !todo.completed));
+      }).finally(() => setLoadingTodo([]));
+  }, [completedTodos]);
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const changeTodo = useCallback(async (id: number, data: any) => {
```
VitaliyBondarenko1982, line 119: no need to disable eslint.
suggest to pass full Todo in changeTodo function
```suggestion
  const changeTodo = useCallback(async (todo: Todo) => {
```

src/Components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+      {tempTodo && (
+        <CSSTransition
+          key={0}
+          timeout={300}
+          classNames="temp-item"
+        >
+          <TodoInfo
+            todo={tempTodo}
+            removeTodo={removeTodo}
```
VitaliyBondarenko1982, line 54: maybe better to pass noop function (`() => {}`) for tempTodo, and not allow user to try remove todo, which not exists on server yet

VitaliyBondarenko1982 (changes requested):
Looks good.
Suggest some improvements.

etojeDenys (approved):
LGTM

PR #403, opened 2023-03-23
Solution

OksanaBaloh:
[DEMO LINK](https://OksanaBaloh.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,204 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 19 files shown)


src/App.tsx
```diff
@@ -1,24 +1,204 @@
...
+      setTodos(prevTodos => prevTodos.filter(todo => todo.id !== id));
+    } catch (error) {
+      setErrorMessage(ErrorType.Delete);
+    } finally {
+      setUpdatingIds([0]);
+    }
+  };
+
+  const removeCompletedTodos = () => {
+    todos.forEach(async (todo) => {
```
vlad-demchuk, line 108: it would be better to use `Promise.All`

vlad-demchuk (approved):
GJ

PR #406, opened 2023-03-25
TODO APP WITH API v1.0

Mykola-Hadupiak:
[TODO APP](https://Mykola-Hadupiak.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,208 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import { Todo } from './types/Todo';
+import { Error } from './types/Error';
+import { FilterTodos } from './types/FilterTodos';
+import {
+  getTodos,
+  createTodo,
+  deleteTodo,
+  updateTodo,
+} from './api/todos';
 import { UserWarning } from './UserWarning';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
 
-const USER_ID = 0;
+const USER_ID = 6550;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState(Error.NONE);
+  const [filterBy, setFilterBy] = useState(FilterTodos.ALL);
+  const [activeInput, setActiveInput] = useState(true);
+
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [idUpdating, setIdUpdating] = useState([0]);
+
+  useEffect(() => {
+    const getTodosFromServer = async () => {
+      try {
+        setErrorMessage(Error.NONE);
+        const todosFromServer = await getTodos(USER_ID);
+
+        setTodos(todosFromServer);
+      } catch (error) {
+        setErrorMessage(Error.LOAD);
+      }
+    };
+
+    getTodosFromServer();
+  }, []);
+
+  const countTodos = todos.length;
+
+  const activeTodos = useMemo(
+    () => todos.filter(todo => !todo.completed).length,
+    [todos],
+  );
+
+  const visibleTodos: Todo[] = useMemo(() => {
+    switch (filterBy) {
+      case FilterTodos.ACTIVE:
+        return todos.filter(todo => !todo.completed);
+      case FilterTodos.COMPLTED:
+        return todos.filter(todo => todo.completed);
+      default:
+        return todos;
+    }
+  }, [todos, filterBy]);
+
+  const addTodo = (title: string) => {
+    if (!title.trim()) {
+      setErrorMessage(Error.TITLE);
+
+      return;
+    }
+
+    const createNewTodo = async () => {
+      try {
+        setErrorMessage(Error.NONE);
+
+        const newTodo = {
+          userId: USER_ID,
+          title,
+          completed: false,
+        };
+
+        setActiveInput(false);
+        setTempTodo({ ...newTodo, id: 0 });
+        const createdTodo = await createTodo(USER_ID, newTodo);
+
+        setTodos(prevTodos => [...prevTodos, createdTodo]);
+      } catch (error) {
+        setErrorMessage(Error.ADD);
+      } finally {
+        setTempTodo(null);
+        setActiveInput(true);
+      }
+    };
+
+    createNewTodo();
+  };
+
+  const removeTodo = async (id: number) => {
+    try {
+      setErrorMessage(Error.NONE);
+      setIdUpdating(prevId => [...prevId, id]);
+
+      await deleteTodo(USER_ID, id);
+
+      setTodos(todos.filter(todo => todo.id !== id));
+    } catch (error) {
+      setErrorMessage(Error.DELETE);
+    } finally {
+      setIdUpdating([0]);
+    }
+  };
+
+  const removeCompletedTodos = () => {
+    todos.forEach(async (todo) => {
+      if (todo.completed) {
+        await removeTodo(todo.id);
+      }
+    });
+  };
+
+  const handleUpdate = async (id: number, data: string | boolean) => {
+    try {
+      setErrorMessage(Error.NONE);
+
+      setIdUpdating(prevId => [...prevId, id]);
+      if (typeof data === 'boolean') {
+        await updateTodo(USER_ID, id, { completed: data });
+      } else {
+        await updateTodo(USER_ID, id, { title: data });
+      }
+
+      setTodos(prevTodos => prevTodos.map(todo => {
+        if (todo.id === id) {
+          const updatedTodo = typeof data === 'boolean' ? {
+            ...todo,
+            completed: data,
+          } : {
+            ...todo,
+            title: data,
+          };
+
+          return updatedTodo;
+        }
+
+        return todo;
+      }));
+    } catch {
+      setErrorMessage(Error.UPDATE);
+    } finally {
+      setIdUpdating([0]);
+    }
+  };
+
+  const toggleAll = () => {
+    let toggleAllTodos = todos;
+
+    if (activeTodos) {
+      toggleAllTodos = toggleAllTodos.filter(todo => !todo.completed);
+    }
+
+    toggleAllTodos.forEach(async (todo) => {
+      await handleUpdate(todo.id, !todo.completed);
+    });
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <Header
+          addTodo={addTodo}
+          isInputActive={activeInput}
+          hasActive={!!activeTodos}
+          hasTodos={!!countTodos}
+          toggleAll={toggleAll}
+        />
+
+        {(!!countTodos || !!tempTodo) && (
+          <>
+            <TodoList
+              todos={visibleTodos}
+              tempTodo={tempTodo}
+              removeTodo={removeTodo}
+              idUpdating={idUpdating}
+              handleUpdate={handleUpdate}
+            />
+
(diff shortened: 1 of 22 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,59 @@
...
+            {filterLink.charAt(0).toUpperCase() + filterLink.slice(1)}
+          </a>
+        ))}
+      </nav>
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        style={{
+          opacity: +Boolean(completedTodos),
+          cursor: !completedTodos ? 'auto' : 'pointer',
```
Volodymyr-Melnyk-Fron10der, line 50: handle truth case first

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,59 @@
...
+          >
+            {filterLink.charAt(0).toUpperCase() + filterLink.slice(1)}
+          </a>
+        ))}
+      </nav>
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        style={{
+          opacity: +Boolean(completedTodos),
```
Volodymyr-Melnyk-Fron10der, line 49: it might be more relevant
```suggestion
          opacity: +Boolean(completedTodos?.length),
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,118 @@
...
+      case title:
+        return;
+      case '':
+        removeTodo(id);
+
+        return;
+
+      default:
+        handleUpdate(id, str);
+    }
```
Volodymyr-Melnyk-Fron10der, line 42: looks a little weird actually) maybe it would be better to use if statement

Volodymyr-Melnyk-Fron10der (changes requested):

src/components/Footer/Footer.tsx
Mykola-Hadupiak: i have number here

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+const USER_ID = 6550;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState(Error.NONE);
+  const [filterBy, setFilterBy] = useState(FilterTodos.ALL);
+  const [activeInput, setActiveInput] = useState(true);
+
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [idUpdating, setIdUpdating] = useState([0]);
```
OlEzhA0, line 26: why do we need to keep 0 here?
```suggestion
  const [updatingIds, setUpdatingIds] = useState([0]);
```

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+import { ErrorNotification } from './components/ErrorNotification';
 
-const USER_ID = 0;
+const USER_ID = 6550;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState(Error.NONE);
+  const [filterBy, setFilterBy] = useState(FilterTodos.ALL);
+  const [activeInput, setActiveInput] = useState(true);
```
OlEzhA0, line 23: `isInputActive` or `isActiveInput` would be better yo boolean vars

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [idUpdating, setIdUpdating] = useState([0]);
+
+  useEffect(() => {
+    const getTodosFromServer = async () => {
```
OlEzhA0, line 29: 👍🏼  for `async/await`
also you can use IIFE in such cases
```
(async () => {
  try {
  
  } catch (err) {

  }
})()
```

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+        setErrorMessage(Error.LOAD);
+      }
+    };
+
+    getTodosFromServer();
+  }, []);
+
+  const countTodos = todos.length;
+
+  const activeTodos = useMemo(
```
OlEzhA0, line 45:
```suggestion
  const activeTodosLength = useMemo(
```

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+    () => todos.filter(todo => !todo.completed).length,
+    [todos],
+  );
+
+  const visibleTodos: Todo[] = useMemo(() => {
```
OlEzhA0, line 50:
```suggestion
  const visibleTodos = useMemo(() => {
```

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+    switch (filterBy) {
+      case FilterTodos.ACTIVE:
+        return todos.filter(todo => !todo.completed);
+      case FilterTodos.COMPLTED:
+        return todos.filter(todo => todo.completed);
+      default:
+        return todos;
```
OlEzhA0, line 57: add empty lines between blocks
```suggestion
      case FilterTodos.ACTIVE:
        return todos.filter(todo => !todo.completed);

      case FilterTodos.COMPLTED:
        return todos.filter(todo => todo.completed);

      default:
        return todos;
```

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+      setTodos(todos.filter(todo => todo.id !== id));
+    } catch (error) {
+      setErrorMessage(Error.DELETE);
+    } finally {
+      setIdUpdating([0]);
+    }
+  };
+
+  const removeCompletedTodos = () => {
+    todos.forEach(async (todo) => {
```
OlEzhA0, line 106: try to avoid using some `async` callbacks for array methods
read this [link](https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971)

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+  ) => {
+    try {
+      setErrorMessage(Error.NONE);
+
+      setIdUpdating(prevId => [...prevId, id]);
+      await updateTodo(USER_ID, id, data);
+
+      setTodos(prevTodos => prevTodos.map(todo => {
+        if (todo.id === id) {
+          const updatedTodo = 'completed' in data ? {
```
OlEzhA0, line 125: why do we need this check? why we can't add completed for each todo but with `false` value?

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+  const toggleAll = () => {
+    let toggleAllTodos = todos;
+
+    if (activeTodos) {
+      toggleAllTodos = toggleAllTodos.filter(todo => !todo.completed);
+    }
+
+    toggleAllTodos.forEach(async (todo) => {
+      await handleUpdate(todo.id, { completed: !todo.completed });
+    });
```
OlEzhA0, line 154: the same as I mentioned above

src/components/Footer/Footer.tsx
OlEzhA0, line 49: :)
```suggestion
          opacity: +!!completedTodos,
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const Header: React.FC<Props> = ({
+  addTodo,
+  hasActive,
+  isInputActive,
+  hasTodos,
+  toggleAll,
+}) => {
+  const [title, setTitle] = useState('');
+  const todoInput = useRef<HTMLInputElement | null>(null);
```
OlEzhA0, line 20:
```suggestion
  const todoInput = useRef<HTMLInputElement>(null);
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    addTodo(title);
+    setTitle('');
+  };
+
+  useEffect(() => {
+    if (todoInput.current) {
```
OlEzhA0, line 30: mb here should be condition for `&& isInputActive`?
for me it seems like you trying to trigger some state and after ping your input, but would be better to pass `ref` to this component and create some function on parent level

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,65 @@
...
+      todoInput.current.focus();
+    }
+  }, [isInputActive]);
+
+  const handleChangeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
```
OlEzhA0, line 35: useCallback?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  const todoTitle = useRef<HTMLInputElement | null>(null);
+
+  useEffect(() => {
+    if (todoTitle.current) {
+      todoTitle.current.focus();
+    }
+  }, [isTodoUpdating]);
+
+  const changeTitle = (str: string | undefined) => {
```
OlEzhA0, line 34:
```suggestion
  const changeTitle = (str?: string) => {
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,119 @@
...
+    if (str === undefined || str === title) {
```
OlEzhA0, line 35:
```suggestion
    if (!str || str === title) {
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export const TodoList: React.FC<Props> = ({
+  todos,
+  tempTodo,
+  removeTodo,
+  idUpdating,
+  handleUpdate,
+}) => {
+  return (
+
```
OlEzhA0, line 25:
```suggestion
```

OlEzhA0 (changes requested):
would be better to save previous todo version in case of failure editing
https://user-images.githubusercontent.com/60295799/227792266-66ed07da-6d17-4dd8-ac84-8632112165c3.mov

src/App.tsx
Mykola-Hadupiak: this is from previous task **immediately after sending a request create a todo with `id: 0`**

DarkMistyRoom: How is this connected? You should set temp todo on adding, but it is an array with updatingIds. If you have questions or don't understand it clearly, ask in the chat

src/App.tsx
DarkMistyRoom: not fixed
image

src/App.tsx
DarkMistyRoom: not fixed
image

src/App.tsx
DarkMistyRoom: not fixed
image

DarkMistyRoom (changes requested):
Fix all comments from the previous review!
And fix this bug: `clear completed left one todo on the screen
https://www.loom.com/share/893193abfd2c41cf8c14b47ab1fecc8d

src/App.tsx
Mykola-Hadupiak: sorry, didnt see this 6 comments, idk why, it was hidden

src/App.tsx
Mykola-Hadupiak: because, in data i have or completed or title
so i change it ( if i have completed, i have to set new completed (true or false))

src/App.tsx
Mykola-Hadupiak: if i leave here 0, only then i have loading animation when i add todo

src/App.tsx
```diff
@@ -1,24 +1,207 @@
...
+
+      default:
+        return todos;
+    }
+  }, [todos, filterBy]);
+
+  const addTodo = async (title: string) => {
+    if (!title.trim()) {
+      setErrorMessage(Error.TITLE);
+
```
OleziO, line 64: Remove unnecessary empty line.

OleziO: Check other similar cases.

src/App.tsx
```diff
@@ -1,24 +1,207 @@
...
+  ) => {
+    try {
+      setErrorMessage(Error.NONE);
+
+      setIdUpdating(prevId => [...prevId, id]);
+      await updateTodo(USER_ID, id, data);
+
+      setTodos(prevTodos => prevTodos.map(todo => {
+        if (todo.id === id) {
+          const updatedTodo = 'completed' in data ? {
```
OleziO, line 129: Try to refactor this code to avoid this check.

src/components/Header/Header.tsx
OleziO: Not resolved, try to change your logic like proposed in the comment, if you have any questions - feel free to ask them in `fe_chat`

src/App.tsx
Mykola-Hadupiak: image

TarasHoliuk (approved):
LGTM :+1:

VitaliyBodnar (approved):

PR #409, opened 2023-03-29
Solution

IVxVI:
[DEMO LINK](https://ivxvi.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoItem/TodoItem.tsx
+++ b/src/components/TodoItem/TodoItem.tsx
@@ -1,0 +1,127 @@
+import classNames from 'classnames';
+import { useCallback, useState } from 'react';
+
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo;
+  removeTodo: (id: number) => Promise<void>;
+  loadingTodoId: number[];
+  loading?: boolean;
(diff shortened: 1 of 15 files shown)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,127 @@
...
+  updateTodo: (id: number, titleOrCompleted: string | boolean) => Promise<void>;
+  setErrorText: (text: string) => void;
+};
+
+export const TodoItem: React.FC<Props> = ({
+  todo, removeTodo, loadingTodoId, loading, updateTodo, setErrorText,
+}) => {
+  const { id, title, completed } = todo;
+  const [todoTitle, setTodoTitle] = useState(title);
+  const [prevTitle, setPrevTitle] = useState(title);
```
anstsot, line 20: do you need 2 states? maybe you can use `todo.title` as previous value

anstsot (approved):
Great job, approved!

src/components/TodoItem/TodoItem.tsx
IVxVI: I do, because my prevTitle saves title before handleEdit changes it to event.target.value, so I handled it like this)

PR #411, opened 2023-03-31
Todo App was created

lilia-obushenko:
[DEMO LINK](https://lilia-obushenko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,78 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import { FC } from 'react';
+import { useApp } from './hooks/useApp';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { ErrorModal } from './components/ErrorModal/ErrorModal';
+import { LoadContext } from './Context/LoadContext';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 6748;
 
-export const App: React.FC = () => {
+export const App: FC = () => {
+  const {
+    todos,
+    error,
+    tempTodo,
+    filterType,
+    addNewTodo,
+    deleteTodo,
+    unableField,
+    activeTodos,
+    visibleTodos,
+    loadingTodos,
+    setFilterType,
+    completedTodos,
+    updateTodoStatus,
+    updateAllStatus,
+    handleCloseError,
+    handleDeleteCompleted,
+  } = useApp();
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+    <LoadContext.Provider value={loadingTodos}>
+      <div className="todoapp">
+        <h1 className="todoapp__title">
+          todos
+        </h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+        <div className="todoapp__content">
+          <Header
+            onAdd={addNewTodo}
+            disabled={unableField}
+            activeTodos={activeTodos.length}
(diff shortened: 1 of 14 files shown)


src/hooks/useApp.tsx
```diff
@@ -0,0 +1,170 @@
...
+} from '../api/todos';
+
+import { FilterType } from '../types/FilterType';
+import { ErrorType } from '../types/ErrorType';
+import { Todo } from '../components/TodoItem/Todo';
+import { getFilteredTodos } from '../helpers/helpers';
+
+const USER_ID = 6748;
+
+export const useApp = () => {
```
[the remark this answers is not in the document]
OleksiiBlezniuk: There is such approach as using "smart and dumb components" in react, "smart" component handle all states, requests, managing data and render "dumb" component. "dumb" component only receives needed props and render all needed components/elements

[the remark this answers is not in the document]
OleksiiBlezniuk: I suggest this approach will be more descriptive then moving all logic to separate hook

src/hooks/useApp.tsx
```diff
@@ -0,0 +1,170 @@
...
+      const response = await getTodos(USER_ID);
+
+      setTodos(response);
+    } catch {
+      showError(ErrorType.LOAD);
+      setTodos([]);
+    }
+  }, []);
+
+  const addNewTodo = useCallback(
```
OleksiiBlezniuk, line 58: If we talk about custom hooks - it's a good example of what can you move to it, separate function that requires useCallback can be moved to separate hook in order to split your logic from big component to smaller logical parts

src/types/ErrorType.tsx
OleksiiBlezniuk, line 1: Btw, you could create single file `App.typedefs.ts` and store all types there instead of creating new file for each enum or type

src/hooks/useApp.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+const USER_ID = 6748;
+
+export const useApp = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState<FilterType>(FilterType.ALL);
+  const [error, setError] = useState<ErrorType>(ErrorType.NONE);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [unableField, setUnableField] = useState(false);
+  const [loadingTodos, setLoadingTodos] = useState([0]);
```
OleksiiBlezniuk, line 28: but why 0 if nothing is loading?

OleksiiBlezniuk, line 28:
```suggestion
  const [loadingTodos, setLoadingTodos] = useState<number[]>([]);
```

OleksiiBlezniuk (changes requested):
GJ! Left few comments
Also when I want to cancel updating and click somewhere else (input is blurred) - update request is still sent, please fix it

https://user-images.githubusercontent.com/84275496/230332545-3ab8fd52-6c7f-444a-a87c-9bf22674e01a.mov

src/App.tsx
```diff
@@ -1,24 +1,200 @@
...
+
+        setTodos(prevTodos => [...prevTodos, todo]);
+      } catch {
+        showError(ErrorType.ADD);
+      } finally {
+        setUnableField(false);
+        setTempTodo(null);
+      }
+    }, [],
+  );
```
OleksiiBlezniuk, line 91: This can be moved to separate hook too

src/App.tsx
```diff
@@ -1,24 +1,200 @@
...
+
+  const {
+    activeTodos,
+    completedTodos,
+  } = useFilteredTodos({ todos });
+
+  const showError = (errorType: ErrorType) => {
+    setError(errorType);
+    setTimeout(() => setError(ErrorType.NONE), 3000);
+  };
```
OleksiiBlezniuk, line 43: Just a suggestion, we can move this function with error state to separate hook that will manage errors (have state with it, reset it after 3 seconds) and return errorType and function to set it

OleksiiBlezniuk (approved):
Well done!

src/App.tsx
lilia-obushenko: Done)

src/App.tsx
lilia-obushenko: Done)

PR #414, opened 2023-04-02
React_todo-app-with-api

07Spirit:
[DEMO LINK](https://07Spirit.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,199 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
+import debounce from 'lodash.debounce';
+import { TodoCreate } from './components/TodoCreate';
+import { TodoInfo } from './components/TodoInfo';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { client } from './utils/fetchClient';
 
-const USER_ID = 0;
+const USER_ID = '6757';
+
+enum SortType {
+  completed,
+  active,
+  all,
+}
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>();
+  const [selectedForm, setSelectedForm] = useState(SortType.all);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [countComplited, setCountComplited] = useState(false);
+  const [countNotComplited, setCountNotComplited] = useState(false);
+
+  useEffect(() => {
+  }, [todosFromServer]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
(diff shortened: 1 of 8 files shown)


polosanya (changes requested):
Same comments here
image
image
todos
image
image
todos-add

slavanosatskyi (changes requested):
There are weird dots appear when you press a couple of whitespaces after some word. See video
https://user-images.githubusercontent.com/6526149/229552747-2de77bab-4a1e-4878-82c9-6e724f62bd2d.mov

it is possible to select for editing several todos simultaneously 
Screenshot 2023-04-03 at 18 09 42

07Spirit:
I dont see any dots.

07Spirit:
https://user-images.githubusercontent.com/101456317/230272086-19769a68-cbd0-4f9a-a6a0-c87d16f66bc9.mov

anstsot (changes requested):
Good work but I found several issues for improvement (check the video)
1. after adding new todo item there is a delay and user is a bit confused. it's better to show the todo in the list with loader
2. some todo items are showing only one space but on editing there are more. maybe it would be better to format?
3. the loader is showing (edit endpoint is called) even if I haven't changed the todo. please add the check to avoid unnecessary calls.
4. if I'm adding empty todo - the input if frozen for some time and the error message isn't shown

https://user-images.githubusercontent.com/22824431/230338250-ef5bec65-a135-421e-92bf-b4014b52aace.mov

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  return (
+    <div
+      className="notification is-danger is-light has-text-weight-normal"
+    >
+      <button
+        type="button"
+        className="delete"
+        onClick={deleteError}
+
```
OleziO, line 24: Remove empty line.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,151 @@
...
+    setIsLoading(true);
+    client.delete(`/todos/${id}`)
+      .catch(() => setErrorMessage('Unable to delete a todo'))
+      .finally(() => {
+        askTodos(url, () => setIsLoading(false));
+      });
+  };
+
+  const handleEdit = (text: string) => {
+    if (text !== '') {
```
OleziO, line 49: Replace with `text`, it will be the same as `text !== ''`

OleziO (changes requested):

src/App.tsx
```diff
@@ -1,24 +1,143 @@
...
+    [todosFromServer],
+  );
+
+  const countComplited = useMemo(
+    () => todosFromServer?.some((todo: Todo) => todo.completed),
+    [todosFromServer],
+  );
+
+  const fetchTodos = (pass: string, callback?: () => void): void => {
+    client
```
anstsot, line 31: you can change loading state here
```suggestion
  const fetchTodos = (pass: string, callback?: () => void): void => {
    setIsLoading(true);
    client
```

anstsot (changes requested):
Great job, just the last few small fixes to add.
Filters should be always in the middle.
image

witflash (approved):
Good job!
As I see you have fixed all the problems. Well done 👍

PR #415, opened 2023-04-03
solution

ivanm2706:
[DEMO LINK](https://ivanm2706.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/Components/Footer/Footer.tsx
+++ b/src/Components/Footer/Footer.tsx
@@ -1,0 +1,57 @@
+import { FC } from 'react';
+import classNames from 'classnames';
+import { FilterType } from '../../types/FilterType';
+
+type Props = {
+  amountCompletedTodos: number,
+  todosLength: number,
+  filterType: FilterType,
+  setFilterType: React.Dispatch<React.SetStateAction<FilterType>>,
+  onClear: () => void,
+};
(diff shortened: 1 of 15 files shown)


src/Components/Footer/Footer.tsx
```diff
@@ -0,0 +1,57 @@
...
+};
+
+const Footer: FC<Props> = ({
+  amountCompletedTodos: completedTodos,
+  todosLength,
+  filterType,
+  setFilterType,
+  onClear,
+}) => {
+  const filterTypes = [FilterType.All, FilterType.Active, FilterType.Completed];
```
etojeDenys, line 20: we can create this constant out of the component

src/Components/Footer/Footer.tsx
```diff
@@ -0,0 +1,57 @@
...
+            href={`#/${type}`}
+            data-select={type}
+            className={classNames(
+              'filter__link',
+              { selected: filterType === type },
+            )}
+            onClick={() => setFilterType(type)}
+          >
+            {type}
+          </a>
```
etojeDenys, line 41: you forgot about key prop

etojeDenys (changes requested):
almost done

PR #417, opened 2023-04-06
solution

webdevw:
[DEMO LINK](https://webdevw.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,283 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
+import { CSSTransition, TransitionGroup } from 'react-transition-group';
+import {
+  getTodos,
+  createTodo,
+  updateTodo,
+  deleteTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header';
+import { Main } from './components/Main';
+import { Footer } from './components/Footer';
+import { Error } from './components/Error';
+import { Filter } from './types/Filter';
+import { ErrorsMessages } from './types/ErrorsMessages';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 6725;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isHeaderDisabled, setIsHeaderDisabled] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo>();
+  const [processings, setProcessings] = useState<number[]>([]);
+  const [currentFilter, setCurrentFilter] = useState<Filter>(Filter.all);
+
+  const removeMessage = () => {
+    setErrorMessage('');
+  };
+
+  const createErrorMessage = (title: ErrorsMessages) => {
+    setErrorMessage(title);
+  };
+
+  const showErrorMessage = (message: ErrorsMessages) => {
+    createErrorMessage(message);
+    setTimeout(() => removeMessage(), 3000);
+  };
+
+  const loadTodos = async () => {
+    setErrorMessage('');
+
+    try {
+      const dataFromServer = await getTodos();
+
+      setTodos(dataFromServer);
+    } catch (error) {
+      showErrorMessage(ErrorsMessages.Load);
+    }
+  };
+
+  const setFilter = (filter: Filter) => {
+    setCurrentFilter(filter);
+  };
+
+  const filterTodos = (filteringProperty: Filter) => {
+    return (
+      filteringProperty === Filter.all
+        ? todos
+        : todos.filter(todo => {
+          switch (filteringProperty) {
+            case Filter.active:
+              return !todo.completed;
+            case Filter.completed:
+              return todo.completed;
+            default:
+              return null;
+          }
+        })
+    );
+  };
+
+  const filteredTodos: Todo[] = useMemo(
+    () => filterTodos(currentFilter),
+    [todos, currentFilter],
+  );
+
+  const addTodo = async (newTodoTitle: string) => {
+    setErrorMessage('');
+    setIsHeaderDisabled(true);
+
+    try {
+      setTempTodo({
+        id: 0,
+        userId: USER_ID,
+        title: newTodoTitle,
+        completed: false,
+      });
+      const newTodoPost = await createTodo({
+        title: newTodoTitle,
+        userId: USER_ID,
+        completed: false,
+      });
+
+      setTodos((items) => [...items, newTodoPost]);
+    } catch (error) {
+      showErrorMessage(ErrorsMessages.Post);
+    }
+
+    setTempTodo(undefined);
+    setIsHeaderDisabled(false);
+  };
+
+  const filterTodosById = (todoElements: Todo[], id: number) => {
+    return todoElements.filter((item) => item.id !== id);
+  };
+
+  const removeTodo = async (id: number) => {
+    setErrorMessage('');
+    setProcessings((prev) => [...prev, id]);
+
+    try {
+      const response = await deleteTodo(id);
+
+      if (response) {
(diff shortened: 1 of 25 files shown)


src/App.tsx
```diff
@@ -1,24 +1,283 @@
...
+      const isTitleChanged = response.title !== todoEdited?.title;
+      const isCompletedChanged = response.completed !== todoEdited?.completed;
+
+      if (isTitleChanged || isCompletedChanged) {
+        setTodos((oldTodos) => {
+          const newTodos = oldTodos.map((el) => ({
+            ...el,
+            completed: el.id === id ? response.completed : el.completed,
+            title: el.id === id ? response.title : el.title,
+          }));
```
Volodymyr-Melnyk-Fron10der, line 169:
```suggestion
          const newTodos = oldTodos.map((todo) => todo.id === id ? updatedTodo : todo);
```

src/App.tsx
```diff
@@ -1,24 +1,283 @@
...
+
+    setProcessings([]);
+  };
+
+  const handleChecker = async (id: number, data: Partial<Todo>) => {
+    setErrorMessage('');
+    setProcessings((prev) => [...prev, id]);
+
+    try {
+      const response = await updateTodo(id, data);
```
Volodymyr-Melnyk-Fron10der, line 158:
```suggestion
      const updatedTodo = await updateTodo(id, data);
```

src/App.tsx
```diff
@@ -1,24 +1,283 @@
...
+      const todoEdited = todos.find((todo) => todo.id === id);
+      const isTitleChanged = response.title !== todoEdited?.title;
+      const isCompletedChanged = response.completed !== todoEdited?.completed;
```
Volodymyr-Melnyk-Fron10der, line 161: Useless checks. Because if nothing isn't changed so why did you send the request? You should check it first

src/App.tsx
```diff
@@ -1,24 +1,283 @@
...
+        });
+      }
+    } catch (error) {
+      showErrorMessage(ErrorsMessages.Update);
+    }
+
+    setProcessings([]);
+  };
+
+  const handlerSwitch = async (AllChecked: boolean) => {
```
Volodymyr-Melnyk-Fron10der, line 183:
```suggestion
  const handlerSwitch = async (isAllChecked: boolean) => {
```

src/App.tsx
```diff
@@ -1,24 +1,283 @@
...
+
+    try {
+      const changedElements = (!AllChecked
+        ? await Promise.all(
+          todosIdArray.map(el => updateTodo(el, { completed: true })),
+        ) : (
+          await Promise.all(
+            todosIdArray.map(el => updateTodo(el, { completed: false })),
+          )
+        ));
```
Volodymyr-Melnyk-Fron10der, line 205:
```suggestion
      const changedElements = await Promise.all(todosIdArray.map(el => updateTodo(el, { completed: AllChecked })),
   
```

src/App.tsx
```diff
@@ -1,24 +1,283 @@
...
+
+      if (changedElements) {
+        setTodos((oldTodos) => {
+          const newTodos = oldTodos.map((el) => ({
+            ...el,
+            completed: (
+              todosIdArray.includes(el.id)
+                ? !el.completed
+                : el.completed
```
Volodymyr-Melnyk-Fron10der, line 214: check previous comment in handleChecker

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+type Props = {
+  errorMessage: string,
+  removeMessage: () => void,
+};
+
+export const Error: React.FC<Props> = (
+  { errorMessage, removeMessage },
+) => {
+  const isError = errorMessage !== '';
```
Volodymyr-Melnyk-Fron10der, line 12: there should be the case such `errorMessage === ''` because you should render the Error compnent at all if there is no errorMessage

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { ErrorsMessages } from '../../types/ErrorsMessages';
+
+type Props = {
+  todos: Todo[],
+  addTodo: (newTodoTitle: string) => Promise<void>,
+  switchBtn: (AllChecked: boolean) => void
```
Volodymyr-Melnyk-Fron10der, line 9: what is switchBtn? make the name with action part like 'onSwitch' or smth like that

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,126 @@
+import classNames from 'classnames';
+import React, { useState } from 'react';
+import { Todo } from '../../types/Todo';
+import { Loader } from '../Loader';
+import { ErrorsMessages } from '../../types/ErrorsMessages';
+
+type Props = {
+  todo: Todo,
+  handleChecker: (id: number, data: Partial<Todo>) => void,
```
Volodymyr-Melnyk-Fron10der, line 9:
```suggestion
  onCheck: (id: number, data: Partial<Todo>) => void,
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,126 @@
...
+        </form>
+      ) : (
+        <span
+          className="todo__title"
+          onDoubleClick={() => setIsFormActive(true)}
+        >
+          {title}
+        </span>
+      )}
+      {isFormActive || (
```
Volodymyr-Melnyk-Fron10der, line 111:
```suggestion
      {!isFormActive && (
```

Volodymyr-Melnyk-Fron10der: it's easier to read in which case we render compinent, not in which we do not

src/types/Filter.ts
```diff
@@ -0,0 +1,5 @@
+export enum Filter {
+  all = 'all',
```
Volodymyr-Melnyk-Fron10der, line 2:
```suggestion
  All = 'all',
```

Volodymyr-Melnyk-Fron10der: fix bellow

Volodymyr-Melnyk-Fron10der (changes requested):

src/App.tsx
```diff
@@ -1,24 +1,283 @@
...
+
+  const removeCompletedTodos = async () => {
+    setErrorMessage('');
+
+    const completedTodos = (
+      todos.filter(el => el.completed)
+        .map(item => item.id)
+    );
+
+    setProcessings(completedTodos);
```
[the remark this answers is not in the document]
webdevw: but why & for What?

src/App.tsx
webdevw: Thanks

src/App.tsx
webdevw: check my solution pls

src/App.tsx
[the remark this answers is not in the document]
Volodymyr-Melnyk-Fron10der: if you are working with async/await structure it's the best practice to do it like that 

```
const func = async () => {
  try {
    setLoading(true);
    
    // other commenads
    
    const res = await request(body);
    
    setResult(res);
  } catch(err) {
    setError(err.message || 'Oops, smth went wrong. Try again pls')
  } finally {
    setLoading(false)
  }
}
```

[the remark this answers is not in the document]
Volodymyr-Melnyk-Fron10der: it's better for user to see the loading right away because there coud be cases when preparing the request body object could take some time as well , yes it's just milliseconds, but anyway)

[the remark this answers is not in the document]
Volodymyr-Melnyk-Fron10der: and there could be the case when error is caused by your code, not by the failed request, some syntax errors and so on - so it would not break the UX - user will se the error anyway and will be able to report it or just will be aware that his action was failed

Volodymyr-Melnyk-Fron10der (approved):
Well done, move on🔥

src/App.tsx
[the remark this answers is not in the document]
webdevw: Ok, thanks

PR #420, opened 2023-04-07
todo app is done

rashyd-hasratov:
[DEMO LINK](https://rashyd-hasratov.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,75 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
 
-const USER_ID = 0;
+import { getTodos } from './api/todos';
+import { UserWarning } from './UserWarning';
+import { TodoList } from './components/TodoList';
+import { ErrorMessage } from './components/ErrorMessage';
+import { Loader } from './components/Loader';
+import { AppContext } from './components/AppContext';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [isLoadingTodos, setIsLoadingTodos] = useState(false);
+
+  // eslint-disable-next-line no-unused-vars
+  const {
+    userId,
+    allTodos,
+    setAllTodos,
+    setShouldShowError,
+    showError,
+  } = useContext(AppContext);
+
+  const loadTodosFromServer = useCallback(async () => {
+    setIsLoadingTodos(true);
+    setShouldShowError(false);
+
+    try {
+      const todosFromServer = await getTodos(userId);
+
+      setAllTodos(todosFromServer);
+    } catch {
+      showError('Unable to load todos');
+    } finally {
+      setIsLoadingTodos(false);
+    }
+  }, []);
+
+  useEffect(() => {
+    loadTodosFromServer();
+  }, []);
+
+  if (!userId) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">
+        todos
+      </h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <Header />
+
+        {isLoadingTodos && (
+          <Loader />
+        )}
+
+        <TodoList />
+
+        {allTodos.length > 0 && (
+          <Footer />
+        )}
+      </div>
+
+      <ErrorMessage />
+    </div>
   );
 };
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,59 @@
+import React, { useCallback, useContext } from 'react';
+import classNames from 'classnames';
+
+import { deleteTodo } from '../../api/todos';
+import { AppContext } from '../AppContext';
+import { TodosFilter } from '../TodosFilter';
+
+export const Footer: React.FC = () => {
+  const {
+    allTodos,
+    setAllTodos,
+    activeTodos,
+    completedTodosIds,
+    loadingTodosIds,
+    setLoadingTodosIds,
+    showError,
+    setShouldShowError,
+  } = useContext(AppContext);
+
+  const clearCompletedTodos = useCallback(async () => {
+    setShouldShowError(false);
+    setLoadingTodosIds(prevIds => [...prevIds, ...completedTodosIds]);
+
+    try {
+      await Promise.all(completedTodosIds.map(id => deleteTodo(id)));
+
+      setAllTodos(activeTodos);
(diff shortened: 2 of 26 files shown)


src/App.tsx
```diff
@@ -1,24 +1,75 @@
...
+import { Loader } from './components/Loader';
+import { AppContext } from './components/AppContext';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [isLoadingTodos, setIsLoadingTodos] = useState(false);
+
+  // eslint-disable-next-line no-unused-vars
```
tiserett, line 20: ??

src/App.tsx
```diff
@@ -1,24 +1,75 @@
...
+  const {
+    userId,
+    allTodos,
+    setAllTodos,
+    setShouldShowError,
+    showError,
```
tiserett, line 26: I believe setShouldShowError should be named as setShowError

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+      <TodosFilter />
+
+      <button
+        type="button"
+        className={classNames(
+          'todoapp__clear-completed',
+          { 'is-invisible': !completedTodosIds.length },
+        )}
+        onClick={handleClearCompletedTodosButtonClick}
```
tiserett, line 53: better get rid of ButtonClick in the end

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,229 @@
...
+    setShouldShowError,
+    loadingTodosIds,
+    setLoadingTodosIds,
+  } = useContext(AppContext);
+
+  const isLoadingState = useMemo(() => (
+    loadingTodosIds.includes(id)
+  ), [loadingTodosIds]);
+
+  const deleteTodoFromServer = useCallback(async () => {
```
tiserett, line 44: not sure about using useCallback here, since you're not passing these functions to any optimized components

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,229 @@
...
+        prevTodo.id !== id
+      )));
+    } catch {
+      showError('Unable to delete a todo');
+    } finally {
+      setLoadingTodosIds([0]);
+    }
+  }, [loadingTodosIds, allTodos]);
+
+  const toggleTodoStatus = useCallback(async () => {
```
tiserett, line 61: same as above

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,229 @@
...
+        return prevTodo;
+      }));
+    } catch {
+      showError('Unable to update a todo');
+    } finally {
+      setLoadingTodosIds([0]);
+    }
+  }, [loadingTodosIds, allTodos]);
+
+  const updateTodoTitle = useCallback(async (updatedTitle: string) => {
```
tiserett, line 85: you know what to read about

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+      setAllTodos(prevTodos => prevTodos.map(prevTodo => {
+        if (prevTodo.id === id) {
+          return {
+            ...prevTodo,
+            completed: !completed,
+          };
+        }
+
+        return prevTodo;
```
tiserett, line 76: I believe it will look clearer with ternary operator

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+        setAllTodos(prevTodos => prevTodos.map(prevTodo => {
+          if (prevTodo.id === id) {
+            return {
+              ...prevTodo,
+              title: updatedTitle,
+            };
+          }
+
+          return prevTodo;
```
tiserett, line 101: same with ternary as above

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,229 @@
...
+      </label>
+
+      {isEditingMode
+        ? (
+          <form
+            onSubmit={(event) => {
+              event.preventDefault();
+
+              finishUpdating();
+            }}
```
tiserett, line 184: create separate handler

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <section className="todoapp__main">
+      {visibleTodos.map(todo => {
+        const { id } = todo;
+
+        return (
+          <TodoInfo key={id} todo={todo} />
+        );
+      })}
```
tiserett, line 39: I think it's better to write it without destructuring, but in much shorter way

src/types/FilterMode.ts
```diff
@@ -0,0 +1,5 @@
+export enum FilterMode {
```
tiserett, line 1: I believe FilterType is a better name

tiserett (changes requested):
Good job, but there are ways for improvement 🚀

src/App.tsx
rashyd-hasratov: If I don't use it, it doesn't allow me to deploy the project. There is written that I imported useContext, but I don't use it (I do use it). The mentor's advice was to disable eslint rule for this line of code

src/App.tsx
rashyd-hasratov: `shouldShowError` has boolean type and decides if error is shown
`setShouldShowError` is used to change the value of the state

`showError` is just a helper function that changes `shouldShowError` state value and sets a timer

image
image

src/components/TodoInfo/TodoInfo.tsx
rashyd-hasratov: Before I fix it I want to clarify the moment :harold:

`TodoInfo` component has own states and the components rerenders every time these states are changed. If we don't `useCallback` we recreate this function every render time. Isn't it bad?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,227 @@
...
+      await patchTodo(id, { completed: !completed });
+
+      setAllTodos(prevTodos => prevTodos.map(prevTodo => {
+        return (prevTodo.id === id)
+          ? {
+            ...prevTodo,
+            completed: !completed,
+          }
+          : prevTodo;
+      }));
```
OleksiiBlezniuk, line 75: Create method for update todo and reuse it for both updating status and title, this logic is same

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,227 @@
...
+    if (newTitleTrimmed) {
+      updateTodoTitle(newTitleTrimmed);
+    } else {
+      deleteTodoFromServer();
+    }
+  }, [newTitle]);
+
+  const handleStatusChange = () => {
+    toggleTodoStatus();
+  };
```
OleksiiBlezniuk, line 121: Why do you need to create this wrapper?

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,40 @@
...
+        case FilterType.Active:
+          return !completed;
+
+        case FilterType.Completed:
+          return completed;
+
+        default:
+          return true;
+      }
+    })
```
OleksiiBlezniuk, line 26: Would be nice to move it to separate function

OleksiiBlezniuk (approved):

PR #424, opened 2023-04-09
React_todo-app-with-api

SerhiiVovchenko:
[DEMO LINK](https://SerhiiVovchenko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,209 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useEffect, useState, useCallback, useMemo,
+} from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  getTodos, addTodo, deleteTodo, patchTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { counterOfActiveTodos, filterTodos } from './helpers/helpers';
+import { FilterParam } from './types/FilterParam';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorType } from './types/ErrorType';
+import { Error } from './components/Error';
+import { Loader } from './Loader';
 
-const USER_ID = 0;
+const USER_ID = 6755;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState(false);
+  const [filterType, setFilterType] = useState<FilterParam>(FilterParam.All);
+  const [errorType, setErrorType] = useState(ErrorType.None);
+  const [title, setTitle] = useState('');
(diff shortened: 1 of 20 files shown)


src/components/Error/Error.tsx
```diff
@@ -0,0 +1,65 @@
...
+        setErrorMessage('');
+        break;
+    }
+  }, [errorType]);
+
+  useEffect(() => {
+    const timer = setTimeout(() => onNotificationClose(), 3000);
+
+    return () => clearTimeout(timer);
+  });
```
polosanya, line 41: The useEffect should have an empty dependency array to mimic componentDidMount behavior, otherwise, the timer might be created multiple times.

src/types/ErrorType.ts
```diff
@@ -0,0 +1,8 @@
+export enum ErrorType {
+  None = 'none',
+  Load = 'load',
+  Add = 'add',
+  Delete = 'delete',
+  Update = 'update',
+  EmptyTitle = 'empty title',
```
polosanya, line 7: Use camelCase for enum members. Change 'empty title' to 'emptyTitle'.

src/App.tsx
```diff
@@ -1,24 +1,209 @@
...
+      setIsOnRender(true);
+    } finally {
+      setLoadingTodo([0]);
+      setIsToggleAll(false);
+      setIsOnRender(false);
+    }
+  };
+
+  const toggleCompletedAllTodo = () => {
+    todos.map(todo => {
```
polosanya, line 141: Use forEach instead of map here, as map should be used when you transform each element of an array and return a new array. You're not using the returned value of map in this case.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,88 @@
...
+      default:
+        return setFilterType(FilterParam.All);
+    }
+  };
+
+  const removeCompletedTodos = () => {
+    if (checkCompletedTodo(todos)) {
+      const completedTodos = filterTodos(todos, FilterParam.Completed);
+
+      completedTodos.map(todo => removeTodo(todo.id));
```
polosanya, line 46: Use `forEach` instead of `map` for iterating over `completedTodos` since you are not returning any value from the loop.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const [newTodoTitle, setNewTodoTitle] = useState(title);
+  const [isCancel, setIsCancel] = useState(false);
+
+  useEffect(() => {
+    setCompleted(todo.completed);
+    setNewTodoTitle(title);
+    setIsCancel(false);
+  }, [todo.completed, title, isCancel]);
+
+  const handleChenge = async (TodoId: number, todoCompleted: boolean) => {
```
polosanya, line 42:
```suggestion
  const handleChange = async (TodoId: number, todoCompleted: boolean) => {
```

polosanya (changes requested):

agniya-i (approved):
I approved, but here are several small comments

PR #427, opened 2023-04-11
Solution

DenysDubas:
(https://DenysDubas.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,201 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, {
+  useState, useEffect, useMemo, useCallback,
+} from 'react';
+import classNames from 'classnames';
 
-const USER_ID = 0;
+import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { getTodos, addTodo, deleteTodo } from './api/todos';
+import { Loader } from './component/Loader/Loader';
+import { TodoList } from './component/TodoList/TodoList';
+import { Footer } from './component/Footer/Footer';
+import { LoadType } from './types/LoadType';
+import { Header } from './component/Header/Header';
+
+const USER_ID = 6908;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [typeOfLoad, setTypeOfLoad] = useState<LoadType>(LoadType.All);
+  const [isLoading, setIsLoading] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [disableInput, setDisableInput] = useState(false);
+  const [loadId, setLoadId] = useState<number[]>([]);
+
+  const getTodosFromServer = async (userId: number) => {
+    try {
+      setIsLoading(true);
+      const getTodo = await getTodos(userId);
+
+      setTodos(getTodo);
+    } catch {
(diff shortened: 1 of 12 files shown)


DenysDubas:
> (https://DenysDubas.github.io/react_todo-app-with-api/)

(https://DenysDubas.github.io/react_todo-app-with-api/)

src/App.tsx
```diff
@@ -1,24 +1,201 @@
...
+        }
+
+        return { ...todo, ...value };
+      }));
+    } catch {
+      setErrorMessage('Unable to update a todo');
+    } finally {
+      setLoadId(prev => prev.filter(num => num !== id));
+    }
+  }, []);
```
KuharchukDmytro, line 128: this blocks try/catch/finally and async keyword make no sense, because for them you need some function which will return promise, in this function you should update the todo on server, not just in the app (make updateTodo function in api/todos.ts file)

KuharchukDmytro (changes requested):
good job, but there are some important issues

first of all, you dont implement updating data function, so the title editing and todos toggling work inccorectly
secondly, you have a lot of bugs with mistakes catching, their text and clearing work inccorectly, for example:
(its dont dissappear after 3 seconds and text are wrong)
image
thirdly, loading animation, when i add/delete/toggle... doesnt work

also there some little issues, title editing submit by enter doesnt work, filter buttons must always be in center, new todos adding to a server always has id 0 (in this issue i dont confident, so you can skip it)

but otherwise you did it nice👍

KuharchukDmytro (approved):
gj!

src/App.tsx
```diff
@@ -1,24 +1,208 @@
...
+    } catch {
+      setErrorMessage(Error.Load);
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  const addNewTodo = useCallback(async (title: string) => {
+    if (!title.trim()) {
+      setErrorMessage('Title can not be empty');
```
OleksiiBlezniuk, line 47: Move error messages to an enum, that way will allow you to avoid magic strings, to be sure that you've not mistaken, and to maintain it in one place.

OleksiiBlezniuk: Especially if you already have such enum

src/App.tsx
```diff
@@ -1,24 +1,208 @@
...
+const USER_ID = 6908;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [typeOfLoad, setTypeOfLoad] = useState<LoadType>(LoadType.All);
+  const [isLoading, setIsLoading] = useState(true);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [disableInput, setDisableInput] = useState(false);
+  const [loadId, setLoadId] = useState<number[]>([]);
```
OleksiiBlezniuk, line 30:
```suggestion
  const [loadingIds, setLoadingIds] = useState<number[]>([]);
```

src/App.tsx
```diff
@@ -1,24 +1,208 @@
...
+        <Header
+          addNewTodo={addNewTodo}
+          disableInput={disableInput}
+          handleEditAll={handleToggleAll}
+        />
+
+        {isLoading && (
+
+          <Loader isLoading={isLoading} />
+        )}
```
OleksiiBlezniuk, line 167:
```suggestion
        {isLoading && (
          <Loader isLoading={isLoading} />
        )}
```

OleksiiBlezniuk (approved):

OleksiiBlezniuk (changes requested):
Missclick
Error message should be reset in 3 seconds after appearing

PR #428, opened 2023-04-11
React todo app with API

Anton-Tretiak:
[DEMO LINK](https://anton-tretiak.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,24 +1,245 @@
...
+      }));
+    } catch {
+      setError('Unable to update a todo');
+      clearError();
+    } finally {
+      setLoadingIds(state => state.filter(el => el !== id));
+    }
+  };
+
+  const toggleAll = () => {
```
valeriiayarzhemska, line 126: be more accurate with naming, toggle what?

src/App.tsx
```diff
@@ -1,24 +1,245 @@
...
+    switch (filterMode) {
+      case FilterType.Active:
+        return allTodos.filter(todo => !todo.completed);
+
+      case FilterType.Completed:
+        return allTodos.filter(todo => todo.completed);
+
+      default:
+        return allTodos;
+    }
```
valeriiayarzhemska, line 152: u can create helpers file for functions like this

src/App.tsx
```diff
@@ -1,24 +1,245 @@
...
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const handleFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!query.trim()) {
+      setError("Title can't be empty");
```
valeriiayarzhemska, line 163: i think its better to create enum for errors

valeriiayarzhemska (changes requested):
gj, but a few issues to make it better)

Oleksandr-kopaevich:
Hey @Anton-Tretiak some points of the task checklists are not done:

toggleAll button should have active class only if all the todos are completed;
**Note:** Currently it's always active

save changes when the field loses focus (onBlur);
**Note:** When a field loses focus changes are not saved and are just lost.

or the deletion error message if we tried to delete the todo.
**Note:** After clicking the button `Clear completed` and getting an error in API (you can just disable network in dev tools to emulate it) completed items are removed from the local react state, but they must stay, because the request to remove failed.

src/components/Footer/Footer.tsx
Oleksandr-kopaevich, line 1: I would call it `FormFooter` or TodoFooter or anything like that, because when it's named just Footer then I expect it to be the global footer that used on every (or most) pages

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,46 @@
...
+        );
+      })}
+    </section>
+
+    {tempTodo && (
+      <TodoInfo
+        todo={tempTodo}
+        key={tempTodo.id}
+        isLoading
+        onDelete={() => {}}
```
Oleksandr-kopaevich, line 41: If some props aren't required then they must be defined like that in types. No need to pass empty functions. Then inside TodoInfo you can call function only if it's defined

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+}
+
+// To have autocompletion and avoid mistypes
+type RequestMethod = 'GET' | 'POST' | 'PATCH' | 'DELETE';
+
+function request<T>(
+  url: string,
+  method: RequestMethod = 'GET',
+  data: any = null, // we can send any data to the server
```
Oleksandr-kopaevich, line 17: This is a good place to use typescript generics here instead of any. But it would be out of the scope of this task, just for general knowledge.

src/App.tsx
Oleksandr-kopaevich, line 1: App.tsx it's the global file that shouldn't contain so many details and logic only something global. Usually, it's used only to connect other components.

src/App.tsx
```diff
@@ -1,24 +1,245 @@
...
+    }
+  };
+
+  const removeCompleted = () => {
+    const completed = todos.filter(todo => todo.completed);
+
+    completed.forEach(todo => {
+      removeTodo(todo.id)
+        .then(() => setTodos(todos.filter(item => !item.completed)))
+        .catch(() => setError('Unable to delete todos'));
```
Oleksandr-kopaevich, line 101: You are calling delete requests for many Todos, and some may fail while others will work. It's not correct to write 'Unable to delete todos' because this failure is related to one specific todo.

Oleksandr-kopaevich (changes requested):
A good job in terms of task solving, but you have to fix and work more on your code structure and organization skills.

src/components/TodoList/TodoList.tsx
Andr1yK, line 41: if u use an empty function u can create global const to avoid needless re-renders

PR #431, opened 2023-04-12
Develop react_todo-app-with-api

sergey-lipikhin:
[DEMO LINK](https://sergey-lipikhin.github.io/react_todo-app-with-api/)


src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,31 @@
...
+  onTodoToggleAll: (isCompleted: boolean) => Promise<void>;
+};
+
+export const TodoHeader: React.FC<Props> = React.memo(({
+  isToggleActive,
+  onTodoAdd,
+  onTodoToggleAll,
+}) => (
+  <header className="todoapp__header">
+    {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
Ambulence, line 17: try to remove

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <FilterButton
+          filterOption={TodoCompletionType.Active}
+          isSelected={TodoCompletionType.Active === selectedFilterOption}
+          onFilterSelect={onFilterSelect}
+        />
+
+        <FilterButton
+          filterOption={TodoCompletionType.Completed}
+          isSelected={TodoCompletionType.Completed === selectedFilterOption}
+          onFilterSelect={onFilterSelect}
```
Ambulence, line 42: I think you can try enum here to avoid DRY

Ambulence (approved):
Very good!!!!

src/App.tsx
```diff
@@ -1,24 +1,216 @@
...
+
+import { filterTodos, getActiveTodosCount } from './common/helpers';
+
+const USER_ID = 6955;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<TodoRich[]>([]);
+  const [error, setError] = useState<ErrorType>(ErrorType.None);
+  const [selectedFilterOption, setSelectedFilterOption]
+    = useState(TodoCompletionType.All);
```
tiserett, line 30: you can rewrite selectedFilterOption to filterType to make it much shorter 😉

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,47 @@
...
+          'notification',
+          'is-danger',
+          'is-light',
+          'has-text-weight-normal',
+          {
+            hidden: error === ErrorType.None,
+          },
+        )}
+      >
+        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
tiserett, line 37: wqhy don't you just add aria label?

src/components/TodoEditForm/TodoEditForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+  render() {
+    const title = this.state.title ?? '';
+
+    return (
+      <>
+        <form
+          onSubmit={(submitEvent) => {
+            submitEvent.preventDefault();
+            this.handleTodoUpdate();
+          }}
```
tiserett, line 93: create separate handler

src/components/TodoEditForm/TodoEditForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+        >
+          <input
+            type="text"
+            className="todo__title-field"
+            placeholder="Empty todo will be deleted"
+            value={title}
+            onChange={(changeEvent) => this.setState({
+              title: changeEvent.target.value,
+            })}
```
tiserett, line 102: create separate handler

src/components/TodoHeader/TodoHeader.tsx
tiserett, line 17: same as I described before, aria label

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,96 @@
...
+    onTodoUpdate(id, { title: newTitle.trim() })
+  );
+
+  const handleEditingSkip = async () => (
+    onTodoUpdate(id, { mode: TodoMode.None })
+  );
+
+  const handleEditingStart = async () => (
+    onTodoUpdate(id, { mode: TodoMode.Editing })
+  );
```
tiserett, line 44: 3 same functions, you can make shrink them in one

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,58 @@
...
+  <section className="todoapp__main">
+    <TransitionGroup>
+      {todos.map(todo => (
+        <CSSTransition
+          key={todo.id}
+          timeout={300}
+          classNames="item"
+        >
+          <TodoItem
+            key={todo.id}
```
tiserett, line 36: I can be mistaken, but you don't have to pass key here, since you passed it inside of the CSSTransition

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,58 @@
...
+      ))}
+
+      {tempTodo && (
+        <CSSTransition
+          key={tempTodo.id}
+          timeout={300}
+          classNames="temp-item"
+        >
+          <TodoItem
+            key={tempTodo.id}
```
tiserett, line 51: same as above

tiserett (changes requested):
not a bad work, but I found some ways to make it even better

tiserett:
image
get rid of this arrow, if there are no todos

src/components/TodoItem/TodoItem.tsx
sergey-lipikhin: I believe it's better to leave the existing functions as they are.

**Main point**: I think it helps to prioritize **readability** and adopt these functions for use on **other levels** of the element hierarchy.

1. There is already a generalized function `onTodoUpdate` that can receive specific todo data, and the three functions serve as wrappers that invoke it. Using more specific function names like `Toggle`, `SkipEditing`, and `StartEditing` makes the code more readable.

2. Two of the functions are passed forward to the next component, and it would be more convenient to simply apply `onEditingSkip()` rather than look at the implementation one level up to figure out what parameters should be passed in.

3. One of the three functions `handleEditingStart` is used as a handler in the current component. If I follow your suggestion, I should need to keep this handler as it is + add your suggested generalized function. This would result in reducing the number of functions from 3 to 2, which is not a significant improvement.

src/types/TodoRich.ts
```diff
@@ -0,0 +1,6 @@
+import { Todo } from './Todo';
+import { TodoMode } from './TodoMode';
+
+export interface TodoRich extends Todo {
```
OleksiiBlezniuk, line 4:
```suggestion
export interface TodoWithMode extends Todo {
```

OleksiiBlezniuk: Strange name, it doesn't give any information about type

src/App.tsx
```diff
@@ -1,24 +1,216 @@
...
+import { Todo } from './types/Todo';
+import { TodoMode } from './types/TodoMode';
+import { TodoCompletionType } from './types/TodoCompletionType';
+import { ErrorType } from './types/ErrorType';
+import { TodoRich } from './types/TodoRich';
+import { TodoRichEditable } from './types/TodoRichEditable';
+
+import { filterTodos, getActiveTodosCount } from './common/helpers';
+
+const USER_ID = 6955;
```
OleksiiBlezniuk, line 24: Looks like constant that can be moved to separate component

src/types/TodoRichEditable.ts
```diff
@@ -0,0 +1,3 @@
+import { TodoRich } from './TodoRich';
+
+export type TodoRichEditable = Partial<Omit<TodoRich, 'id' & 'userId'>>;
```
OleksiiBlezniuk, line 3:
```suggestion
export type TodoDataToUpdate = Partial<Omit<TodoRich, 'id' & 'userId'>>;
```

src/App.tsx
```diff
@@ -1,24 +1,216 @@
...
+
+  const handleTodoDelete = async (todoId: number) => {
+    updateTodoLocal(todoId, { mode: TodoMode.Loading });
+
+    try {
+      await deleteTodoOnServer(USER_ID, todoId);
+      deleteTodoLocal(todoId);
+    } catch {
+      setError(ErrorType.DeleteTodo);
+      updateTodoLocal(todoId, { mode: TodoMode.None });
```
OleksiiBlezniuk, line 105: But why should we update todo if we removed it?

src/App.tsx
```diff
@@ -1,24 +1,216 @@
...
+    todoId: number,
+    updatedData: TodoRichEditable,
+  ) => {
+    const keys = Object.keys(updatedData);
+
+    if (keys.length === 1 && keys[0] === 'mode') {
+      updateTodoLocal(todoId, updatedData);
+
+      return;
+    }
```
OleksiiBlezniuk, line 119: If you need only reset loading for todo - do it in separate function

src/App.tsx
```diff
@@ -1,24 +1,216 @@
...
+    } finally {
+      updateTodoLocal(todoId, { mode: TodoMode.None });
+    }
+  };
+
+  const handleCompletedTodosDelete = async () => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    await Promise.all(
+      completedTodos.map(todo => handleTodoDelete(todo.id)),
```
OleksiiBlezniuk, line 140:
```suggestion
      completedTodos.map(async todo => handleTodoDelete(todo.id)),
```

src/App.tsx
```diff
@@ -1,24 +1,216 @@
...
+    };
+
+    setTempTodo(newTodo);
+
+    try {
+      const newTodoFromServer = await addTodoOnServer(USER_ID, newTodo);
+
+      addTodoLocal(newTodoFromServer);
+    } catch {
+      setError(ErrorType.AddTodo);
```
OleksiiBlezniuk, line 91: let's reset this error  after 3 seconds
Screenshot 2023-04-18 at 16 46 27

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,47 @@
...
+  error,
+  onClose,
+}) => {
+  const errorMessage = errorMessages[error];
+
+  useEffect(() => {
+    const timerId = setInterval(onClose, 3000);
+
+    return () => clearInterval(timerId);
+  }, [onClose]);
```
OleksiiBlezniuk, line 22: Why do you need interval for it? Use timeout instead
Also better to reset error in App component (create one method that will add and reset error message) and render this component only when there is some errorMessage

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <>
```
OleksiiBlezniuk, line 25: You don't need this fragment

src/components/TodoEditForm/TodoEditForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+  onTodoDelete: () => Promise<void>;
+};
+
+type State = {
+  title: string | null;
+};
+
+export class TodoEditForm extends React.Component<Props, State> {
+  state = {
+    title: null,
```
OleksiiBlezniuk, line 16: why not empty string?

src/components/TodoEditForm/TodoEditForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+  onEditingSkip: () => void;
+  onTodoTitleUpdate: (newTitle: string) => Promise<void>;
+  onTodoDelete: () => Promise<void>;
+};
+
+type State = {
+  title: string | null;
+};
+
+export class TodoEditForm extends React.Component<Props, State> {
```
OleksiiBlezniuk, line 14: But why do you need class component for this form?

OleksiiBlezniuk: Wouldn't it be enough to use functional component?

src/components/TodoEditForm/TodoEditForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+  state = {
+    title: null,
+  }
+
+  private inputRef: React.RefObject<HTMLInputElement> = React.createRef();
+
+  static getDerivedStateFromProps(props: Props, state: State) {
```
OleksiiBlezniuk, line 21: are you sure you need this method?

src/components/TodoEditForm/TodoEditForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    const {
+      handleFormSubmit,
+      handleTitleChange,
+      handleTodoUpdate,
+      inputRef,
+    } = this;
+
+    return (
+      <>
```
OleksiiBlezniuk, line 106: you don't need this fragment

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        disabled={!hasCompletedTodos}
+        onClick={onCompletedTodosDelete}
+      >
+        Clear completed
+      </button>
+    </footer>
+  </>
```
OleksiiBlezniuk, line 55: you don't need this fragment

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+import { TodoRich } from '../../types/TodoRich';
+import { TodoMode } from '../../types/TodoMode';
+import { TodoRichEditable } from '../../types/TodoRichEditable';
+
+type Props = {
+  todo: TodoRich;
+  onTodoDelete?: (todoId: number) => Promise<void>;
+  onTodoUpdate?: (todoId: number, updatedData: TodoRichEditable)
+  => Promise<void>;
```
OleksiiBlezniuk, line 15:
```suggestion
  onTodoUpdate?: (
    todoId: number,
    updatedData: TodoRichEditable
  ) => Promise<void>;
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,96 @@
...
+  todo: {
+    id,
+    title,
+    completed,
+    mode,
+  },
+  onTodoDelete = () => {},
+  onTodoUpdate = () => {},
+}) => {
+  const hadndleTodoDelete = async () => onTodoDelete(id);
```
OleksiiBlezniuk, line 28:
```suggestion
  const handleTodoDelete = async () => onTodoDelete(id);
```

OleksiiBlezniuk (changes requested):
GJ, left few comments

src/App.tsx
sergey-lipikhin: This statement is in `catch` block - if we didn't remove `todo` in case of an error, we will leave the `todo` item in the list and just change its mode to `None`

PR #436, opened 2023-04-12
solution

Lilu580:
[DEMO LINK](https://Lilu580.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -164,8 +164,8 @@ export const App: React.FC = () => {
   }
 
   return (
-    <div className="todoapp">
-      <h1 className="todoapp__title">todos</h1>
+    <div className="main mr-4 ml-4">
+      <h1 className="title is-1 has-text-centered">Todos</h1>
 
       <div className="todoapp__content">
         <Header
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -38,7 +38,7 @@ export const Footer: React.FC<Props> = ({
       {completedTodos.length > 0 && (
         <button
           type="button"
-          className="button is-primary is-small"
+          className="button is-primary is-small "
           onClick={onDeleteCompletedTodos}
         >
           Clear completed
--- a/src/components/Nav/Nav.tsx
+++ b/src/components/Nav/Nav.tsx
@@ -9,7 +9,7 @@ type Props = {
 
 export const Nav: React.FC<Props> = ({ filterType, onSetFilterType }) => {
   return (
-    <nav className="filter">
+    <nav className="filter ">
       <button
         type="button"
         className={classNames(
--- a/src/components/Notification/Notification.tsx
+++ b/src/components/Notification/Notification.tsx
@@ -10,7 +10,7 @@ export const Notification: React.FC<Props>
 = ({ errorMessage, removeErrorMessage }) => {
   return (
     <div className={classNames(
-      'notification is-danger is-light has-text-weight-normal',
+      'notification is-primary',
       { hidden: !errorMessage },
     )}
     >
--- a/src/styles/filter.scss
+++ b/src/styles/filter.scss
@@ -1,5 +1,6 @@
 .filter {
   display: flex;
+  box-sizing: border-box;
 
   &__link {
     margin: 3px;
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -4,7 +4,6 @@ iframe {
 
 body {
   min-width: 230px;
-  max-width: 550px;
   margin: 0 auto;
 }
 


Yevhenii0536:
https://www.loom.com/share/a17deda3f5d241b6b6fd586ba8acce12

- when you completed all todos loader running only on one todo
- deleting process doesn't work. need reload to see changes
- when completed all todos counter in footer doesn't show how many todos we have
- autofocus onBlur doesn't work

Yevhenii0536:
image

src/App.tsx
```diff
@@ -1,24 +1,178 @@
...
+      setError('Unable to delete a todo');
+      setTimeout(() => {
+        setError('');
+      }, 3000);
+    } finally {
+      setIsWaiting(0);
+    }
+  }, []);
+
+  const changeTodo = useCallback(async (id: number, data: Partial<Todo>) => {
```
Yevhenii0536, line 78: Omit here more useful then Partial

src/App.tsx
```diff
@@ -1,24 +1,178 @@
...
+      todos.forEach(el => {
+        changeTodo(el.id, { completed: false });
+      });
+    } else {
+      const notDoneTodos = todos.filter(el => !el.completed);
+
+      notDoneTodos.forEach(element => {
+        changeTodo(element.id, { completed: true });
+      });
+    }
```
Yevhenii0536, line 110: if you want to disable all loading todos at one time

- pass all ids to array before running async function for each todo
- replace forEach with map
- use Promise.all for this map
- await this promise
- then set empty array to loading todo ids

src/App.tsx
```diff
@@ -1,24 +1,178 @@
...
+  }, [todos]);
+
+  const removeComletedTodos = () => {
+    setIsDeletingCompleted(true);
+
+    completedTodos.forEach(todo => {
+      deleteTodos(todo.id)
+        .then(() => {
+          setTodos(todos.filter(item => !item.completed));
+        })
```
Yevhenii0536, line 120: also here

src/App.tsx
```diff
@@ -1,24 +1,178 @@
...
+    });
+  };
+
+  const removeError = () => {
+    setError('');
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
```
Yevhenii0536, line 136: you don't need it
```suggestion

```

src/App.tsx
```diff
@@ -1,24 +1,178 @@
...
+        })
+        .finally(() => {
+          setIsDeletingCompleted(false);
+        });
+    });
+  };
+
+  const removeError = () => {
+    setError('');
+  };
```
Yevhenii0536, line 132: you can separately setError from this function

Yevhenii0536:
https://www.loom.com/share/cad5dea181074a668a026d8769b395f6
after infinity loop input doesn't work

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,90 @@
...
+            },
+          )}
+          onClick={() => {
+            if (filterType !== FilterType.All) {
+              onSetFilterType(FilterType.All);
+            }
+          }}
+        >
+          All
+        </a>
```
Yevhenii0536, line 42: make NAV as separately component

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,70 @@
...
+    };
+
+    setTitle('');
+
+    onAdd(newTodo);
+  };
+
+  return (
+    <header className="todoapp__header">
+      {/* this buttons is active only if there are some active todos */}
```
Yevhenii0536, line 43: keep clean

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,131 @@
...
+      </label>
+
+      {isEditing
+        ? (
+          <form onSubmit={handleSubmit}>
+            <input
+              className="todo__title-field"
+              type="text"
+              value={todoTitle}
+              onChange={(event) => setTodoTitle(event.target.value)}
```
Yevhenii0536, line 91: make handler for this action

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,54 @@
+import { Todo } from '../../types/Todo';
+import { TodoItem } from '../TodoItem';
+import { FilterType } from '../../types/FilterType';
+import { getFilteredTodos } from '../../utils/helpers';
+
+type Props = {
+  todos: Todo[]
+  activeTodos: number
+  filterType: FilterType
+  tempTodo: Todo | undefined,
```
Yevhenii0536, line 10: better to use null instead undefined

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    const newTodo = {
+      title,
+      completed: false,
+      userId,
+    };
+
+    setTitle('');
+
+    onAdd(newTodo);
```
Yevhenii0536, line 38: how about tempTodo?

Yevhenii0536 (changes requested):

src/components/Nav/Nav.tsx
```diff
@@ -0,0 +1,65 @@
...
+          },
+        )}
+        onClick={() => {
+          if (filterType !== FilterType.Completed) {
+            onSetFilterType(FilterType.Completed);
+          }
+        }}
+      >
+        Completed
+      </a>
```
Yevhenii0536, line 62: You make same actions every link. Make helper and separate component of links

Yevhenii0536 (approved):

Oleksandr-kopaevich:
> https://www.loom.com/share/a17deda3f5d241b6b6fd586ba8acce12
> 
> * when you completed all todos loader running only on one todo
> * deleting process doesn't work. need reload to see changes
**> * when completed all todos counter in footer doesn't show how many todos we have**
> * autofocus onBlur doesn't work
> * arrow must be hidden when no todos in the list

hey @Yevhenii0536, you've done a nice and deep review, with lots of comments, suggestions, and even videos. I agree with all of your messages except for one small point. Just be more attentive to requirements the next time. If something isn't said there, then double-check the demo of the working example which link is usually in the task description. 
We don't show how much todos we have, but how much is left to do.
image

Oleksandr-kopaevich (changes requested):
I've also left some comments about how you can improve the code style.

OleksiiBlezniuk (approved):
I don't think that it's a good idea to make a list so wide, it looks bad on big screens. Better to set the width and center this block
Screenshot 2023-05-25 at 17 12 36

PR #440, opened 2023-04-13
Raw version without toggleAll

Yevhenii0536:
[DEMO LINK](https://yevhenii0536.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,176 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
+import {
+  deleteTodo,
+  getTodos,
+  updateTodoID,
+  updateTodoTitle,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './Components/TodoList/TodoList';
+import { filterTodos } from './utils/helpers/filterTodos';
+import { Footer } from './Components/Footer/Footer';
+import { Header } from './Components/Header/Header';
+import { ErrorMessage } from './Components/ErrorMessage/ErrorMessage';
+import { USER_ID } from './utils/UserID';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todoList, setTodoList] = useState<Todo[]>([]);
+  const [filteredBy, setFilteredBy] = useState('All');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [titleTodo, setTitleTodo] = useState('');
+  const [deletedTodos, setDeletedTodos] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [activeInput, setActiveInput] = useState(true);
+
+  const visibleTodos = useMemo(() => {
+    return filterTodos(todoList, filteredBy);
+  }, [todoList, filteredBy]);
+
+  const handleChangeFilterBy = (value: string) => {
+    setFilteredBy(value);
+  };
+
+  const handleChangeStatus = useCallback(
+    async (todoID: number) => {
+      try {
+        const foundTodo = todoList.find((todo) => todo.id === todoID);
+
+        setDeletedTodos((prevIDs) => [...prevIDs, todoID]);
+        await updateTodoID(todoID, !foundTodo?.completed);
+
+        const newTodos = todoList.map((todo) => {
+          if (foundTodo?.id === todo.id) {
+            foundTodo.completed = !foundTodo.completed;
+
+            return foundTodo;
+          }
+
+          return todo;
+        });
+
+        setTodoList(newTodos);
+      } catch {
+        setErrorMessage('Unable to update a todo');
+      } finally {
+        setDeletedTodos([]);
+      }
+    },
+    [todoList],
+  );
+
+  const handleUpdateTitleTodo = useCallback(
+    async (todoID: number, todoNewTitle: string) => {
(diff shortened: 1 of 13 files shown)


src/Components/Footer/Footer.tsx
```diff
@@ -0,0 +1,80 @@
+import classNames from 'classnames';
+import { Links } from '../../utils/enum';
+import { Todo } from '../../types/Todo';
+import { deleteTodo } from '../../api/todos';
+
+type Props = {
+  filteredBy: string;
+  todosLengh: number;
```
OleksiiBlezniuk, line 8: There is a typo in the property name `todosLengh`. Consider renaming it to `todosLength` to avoid confusion.

src/Components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,162 @@
...
+        setIsChangedTitle(true);
+        await onChangeNewTitle?.(id, todoTitle);
+      } catch {
+        setErrorMessage('Unable to edit a todo');
+      } finally {
+        setIsChangedTitle(false);
+      }
+    }
+
+    return '';
```
OleksiiBlezniuk, line 58: There is no need to return an empty string in the 'handleUpdateTittleTodo' function. You can simply remove this line.

src/Components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  return (
+    <>
+      {isClicked ? (
+        <div
+          className={classNames('todo', {
+            completed: todo.completed,
+          })}
+          key={todo.id}
+          onDoubleClick={() => setIsClicked((prev) => !prev)}
```
OleksiiBlezniuk, line 90: The 'onDoubleClick' event should be placed on the outer div (line 121) to handle both cases when the input is clicked and not clicked, instead of duplicating the code.

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+  );
+
+  const handleUpdateTitleTodo = useCallback(
+    async (todoID: number, todoNewTitle: string) => {
+      if (!todoNewTitle) {
+        try {
+          setLoadingTodos([todoID]);
+          const newTodos = todoList.filter((todo) => todo.id !== todoID);
+
+          await deleteTodo(`/todos/${todoID}`);
```
OleksiiBlezniuk, line 86: Create functions for making request, endpoints can be changed and it is better to maintain it in one place

src/Components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,162 @@
...
+    event: React.ChangeEvent<HTMLInputElement>,
+  ) => {
+    const {
+      target: { value },
+    } = event;
+
+    setTodoTitle(value);
+  };
+
+  const handleUpdateTittleTodo = async (
```
OleksiiBlezniuk, line 40:
```suggestion
  const handleUpdateTitleTodo = async (
```

src/api/todos.ts
```diff
@@ -0,0 +1,34 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = async (userId: number) => {
+  const todos = await client.get<Todo[]>(`/todos?userId=${userId}`);
+
+  return todos;
```
OleksiiBlezniuk, line 7:
```suggestion
  return client.get<Todo[]>(`/todos?userId=${userId}`);
```

src/api/todos.ts
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export const addTodo = async (todo: Omit<Todo, 'id'>) => {
+  const add = await client.post<Todo>('/todos', todo);
+
+  return add;
+};
+
+export const deleteTodo = async (todoID: string) => {
+  const deletedTodo = await client.delete(todoID);
```
OleksiiBlezniuk, line 17: why can't we pass endpoint only here instead of passing everywhere where we use this function?

src/utils/enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum Links {
```
OleksiiBlezniuk, line 1:
```suggestion
export enum FilterBy {
```

src/utils/helpers/filterTodos.tsx
```diff
@@ -0,0 +1,19 @@
+import { Todo } from '../../types/Todo';
+
+export const filterTodos = (todos: Todo[], filterBy: string) => {
+  if (todos) {
```
OleksiiBlezniuk, line 4: But this parameter is required, why should we check if it passed? Let typescript do it for you

OleksiiBlezniuk (changes requested):

Oleksandr-kopaevich (commented):
hey @Yevhenii0536, this point from the checklist wasn't done:
`toggleAll button should have active class only if all the todos are completed;`
Please be more careful with details when reading the tasks.
Also here is a demo of a small UI bug
https://stg.guidde.com/share/playbooks/3ZHYCRnU1KbfmVCyJCJxiu?origin=SRXXVVidcvdgopt8HY8pMu1ZCnr1

Oleksandr-kopaevich (changes requested):
Please, fix all the review comments before submitting a new review. Maybe you forgot to push your changes?

src/Components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,162 @@
...
+  };
+
+  const onBlurUpdateTittleTodo = () => {
+    setIsClicked(false);
+
+    if (todoTitle !== todo.title) {
+      return onChangeNewTitle?.(id, todoTitle);
+    }
+
+    return '';
```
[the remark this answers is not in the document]
Yevhenii0536: image

OleksiiBlezniuk (approved):
Will approve, but not all comments were fixed, next time pay more attention to it

PR #442, opened 2023-04-13
react_todo-app-with-api solution

stanislavkulakovskyi:
[DEMO LINK](https://stanislavkulakovskyi.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,267 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useState, useCallback, useEffect, useMemo,
+} from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import {
+  getTodos, postTodo, deleteTodo, updateTodo,
+} from './api/todos';
+import { TodoList } from './components/TodoList';
+import { FilterTypes } from './types/FilterTypes';
+import { Footer } from './components/Footer';
 
-const USER_ID = 0;
+const USER_ID = 6962;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
+  const [query, setQuery] = useState('');
+  const [errorText, setErrorText] = useState('');
+  const [filterType, setFilterType] = useState(FilterTypes.ALL);
+  const [loading, setLoading] = useState(false);
+  const [loadingTodoId, setLoadingTodoId] = useState<number[]>([0]);
+
+  const closeError = () => setTimeout(() => setErrorText(''), 3000);
+  const handleCloseButton = () => setErrorText('');
+  const completedTodos = todos.filter(todo => todo.completed);
+  const activeTodos = todos.filter(todo => !todo.completed);
+
+  const fetchTodos = useCallback(async () => {
+    try {
+      const todosFromServer = await getTodos(USER_ID);
+
+      setTodos(todosFromServer);
+    } catch (error) {
+      setErrorText('Unable to load todos');
+      closeError();
+    }
+  }, []);
+
+  useEffect(() => {
+    fetchTodos();
+  }, [fetchTodos]);
+
+  const addTodo = useCallback(async (title: string) => {
+    try {
+      setLoading(true);
+
+      if (!title.length) {
+        setErrorText("Title can't be empty");
+        closeError();
+
+        return;
+      }
+
+      const newTodo = {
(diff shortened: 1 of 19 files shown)


src/api/todos.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const postTodo = (data: NewTodo) => {
+  return client.post<Todo>('/todos', data);
+};
+
+export const deleteTodo = (id: number) => {
+  return client.delete(`/todos/${id}`);
+};
+
+export const updateTodo = (id: number, data: object) => {
```
OleksiiBlezniuk, line 17: Instead of using 'object' as a type, you should use existing todo interface with Partial

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,62 @@
...
+          return (
+            <CSSTransition
+              key={todo.id}
+              timeout={200}
+              classNames="item"
+            >
+              <TodoItem
+                key={todo.id}
+                todo={todo}
+                loading={false}
```
OleksiiBlezniuk, line 33: The `loading` prop of the `TodoItem` should be dynamically set based on the `loadingTodoId` array. You can use `loadingTodoId.includes(todo.id)` to check if the current `todo.id` is being loaded.

src/App.tsx
```diff
@@ -1,24 +1,267 @@
-/* eslint-disable max-len */
```
OleksiiBlezniuk, line 1: Remove unused eslint-disable comments at lines 1 and 2. The linter should not report any issues related to `max-len` and `jsx-a11y/control-has-associated-label`.

src/App.tsx
```diff
@@ -1,24 +1,267 @@
...
+  const activeTodos = todos.filter(todo => !todo.completed);
+
+  const fetchTodos = useCallback(async () => {
+    try {
+      const todosFromServer = await getTodos(USER_ID);
+
+      setTodos(todosFromServer);
+    } catch (error) {
+      setErrorText('Unable to load todos');
+      closeError();
```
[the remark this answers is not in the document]
OleksiiBlezniuk: And use enum for errors

src/App.tsx
```diff
@@ -1,24 +1,267 @@
...
+            }
+
+            return { ...todo, title: data };
+          }
+
+          return todo;
+        }));
+      } catch (error) {
+        if (typeof data === 'string' && data.length === 0) {
+          setErrorText('');
```
OleksiiBlezniuk, line 116: This function should be unique to make possible to reuse it, so better to create wrapper for title submit function and handle it there

src/App.tsx
```diff
@@ -1,24 +1,267 @@
...
+
+  const toggleAllCompleted = async () => {
+    try {
+      const activeIds = activeTodos.map(activeTodo => activeTodo.id);
+      const completedIds = completedTodos.map(
+        completedTodo => completedTodo.id,
+      );
+
+      let toggleTodos;
+      let updatedTodos;
```
OleksiiBlezniuk, line 136: In that case you should add types of this variables

src/App.tsx
```diff
@@ -1,24 +1,267 @@
...
+        setLoadingTodoId(completedIds);
+
+        toggleTodos = completedTodos.map(completedTodo => (
+          updateTodo(completedTodo.id, { ...completedTodo, completed: false })
+        ));
+
+        updatedTodos = todos.map(todo => (
+          completedIds.includes(todo.id)
+            ? { ...todo, completed: false } : todo
+        ));
```
OleksiiBlezniuk, line 158: But only data is different here, let's reuse code

src/App.tsx
```diff
@@ -1,24 +1,267 @@
...
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
OleksiiBlezniuk, line 210: fix if using aria-label

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,126 @@
...
+  );
+
+  const handleBlur = useCallback(() => {
+    if (todoTitle.length === 0) {
+      removeTodo(id).then(() => setIsUpdating(false))
+        .catch(() => setErrorText('Unable to remove a todo'));
+    } else {
+      updateTodo(id, todoTitle).then(() => setIsUpdating(false))
+        .catch(() => setErrorText('Unable to update a todo'));
+    }
```
OleksiiBlezniuk, line 50: Just call same function as you call on submit

PR #444, opened 2023-04-13
Todo app

MaksymMaliuk:
[DEMO LINK](https://MaksymMaliuk.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,7 +7,7 @@ import React, {
 import { Notification } from './components/Notification';
 import { UserWarning } from './UserWarning';
 import { Todo } from './types/Todo';
-import { ToodoList } from './components/TodoList/TodoList';
+import { TodoList } from './components/TodoList/TodoList';
 import {
   getTodos,
   postTodo,
@@ -71,7 +71,7 @@ export const App: React.FC = () => {
     try {
       const result = await postTodo(newTodo);
 
-      setTodos(state => [...state, result]);
+      setTodos((state) => [...state, result]);
     } catch (error) {
       setErrorMessage('Unable to add a todo');
       clearErrorMessage();
@@ -201,7 +201,7 @@ export const App: React.FC = () => {
       <div className="todoapp__content">
         {todos.length > 0 && (
           <>
-            <ToodoList
+            <TodoList
               temporaryTodo={temporaryTodo}
               onDeleteTodo={removeTodo}
               onUpdateTodo={handleUpdate}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -163,17 +163,16 @@ export const App: React.FC = () => {
   const removeCompletedTodos = useCallback(() => {
     const completedTodos = todos.filter(todo => todo.completed);
 
-    completedTodos.map(todo => {
-      return deleteTodo(todo.id)
-        .then(() => {
-          setTodos(todos.filter(task => !task.completed));
-        })
-        .catch(() => {
-          setErrorMessage('Unable to remove todo');
-          setTimeout(() => {
-            setErrorMessage('');
-          }, 3000);
-        });
+    completedTodos.map(async todo => {
+      try {
+        await deleteTodo(todo.id);
+        setTodos(todos.filter(task => !task.completed));
+      } catch {
+        setErrorMessage('Unable to remove todo');
+        setTimeout(() => {
+          setErrorMessage('');
+        }, 3000);
+      }
     });
   }, [todos]);
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,11 +7,12 @@ import React, {
 import { Notification } from './components/Notification';
 import { UserWarning } from './UserWarning';
 import { Todo } from './types/Todo';
-import { ToodList } from './components/TodoList/TodoList';
+import { ToodoList } from './components/TodoList/TodoList';
 import {
   getTodos,
   postTodo,
-  deleteTodo, patchTodo,
+  deleteTodo,
+  patchTodo,
 } from './api/todos';
 import { Loader } from './components/Loader';
 import { FilterType } from './types/FilterEnum';
@@ -29,6 +30,7 @@ export const App: React.FC = () => {
   const [isLoading, setIsLoading] = useState(true);
   const [loadingIds, setLoadingIds] = useState<number[]>([]);
   const [selectedFilter, setSelectedFilter] = useState(FilterType.All);
+  const [isAllCompleted, setIsAllCompleted] = useState(false);
 
   useEffect(() => {
     const getTodosFromServer = async () => {
@@ -49,7 +51,13 @@ export const App: React.FC = () => {
     getTodosFromServer();
   }, []);
 
-  const addTodo = (todoTitle: string) => {
+  const clearErrorMessage = () => {
+    setTimeout(() => {
+      setErrorMessage('');
+    }, 3000);
+  };
+
+  const addTodo = async (todoTitle: string) => {
     const newTodo = {
       title: todoTitle,
       userId: USER_ID,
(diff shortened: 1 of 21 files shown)


stanislavkulakovskyi (approved):
Good job!

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+      })
+      .catch(() => {
+        setErrorMessage('Unable to add a todo');
+        setTimeout(() => {
+          setErrorMessage('');
+        }, 3000);
+      })
+      .finally(() => {
+        setIsInputDisabled(false);
+        setTemporaryTodo(undefined);
```
OleksiiBlezniuk, line 75: Let's use one approach, use either promise chaining methods or async/await syntax (async/await is more preferable)

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+  };
+
+  const handleFormSubmit = useCallback((event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!title) {
+      setErrorMessage('Tittle cannot be empty');
+      setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
```
OleksiiBlezniuk, line 86: I saw same logic before, let's create method for it

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+    setTitle('');
+  }, [title]);
+
+  const removeTodo = useCallback(async (id: number) => {
+    setLoadingIds(state => [...state, id]);
+    try {
+      await deleteTodo(id);
+      setTodos(todos.filter(todo => todo.id !== id));
+    } catch {
+      setErrorMessage('Unable to delete a todo');
```
OleksiiBlezniuk, line 101: Better to save such messages to constants or even to enums, avoid magic strings

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+      setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    } finally {
+      setLoadingIds(state_1 => state_1.filter(el => el !== id));
```
OleksiiBlezniuk, line 106: strange_name_1

OleksiiBlezniuk: don't use numbers in variables names

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+    try {
+      await patchTodo(id, data);
+
+      setTodos(state => state.map(todo => {
+        if (todo.id === id) {
+          return { ...todo, ...data };
+        }
+
+        return todo;
+      }));
```
OleksiiBlezniuk, line 136: It's a bad idea to use local data after update to set it to state
Data can be changed on server side or because of some error it can be not processed etc.
Use data from API as source of truth, `patch` returns updated data, use it instead of local

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+    } catch {
+      setErrorMessage('Unable to add a todo');
+      setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    } finally {
+      setLoadingIds(state => state.filter(el => el !== id));
```
OleksiiBlezniuk, line 143: use more descriptive names, what is el?

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+    }
+  };
+
+  const toggleAllCompleted = useCallback(() => {
+    const allCompleted = todos.every(todo => todo.completed);
+
+    if (allCompleted) {
+      todos.map(todoEl => {
+        return handleUpdate(todoEl.id, { completed: false });
+      });
```
OleksiiBlezniuk, line 153: Why map? If you want to return promises and await them all - wrap it to Promise.all, if no - use forEach

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+    } else {
+      const notCompleted = todos.filter(el => !el.completed);
```
OleksiiBlezniuk, line 155:
```suggestion
      const notCompleted = todos.filter(todo => !todo.completed);
```

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+      try {
+        await deleteTodo(todo.id);
+        setTodos(todos.filter(task => !task.completed));
+      } catch {
+        setErrorMessage('Unable to remove todo');
+        setTimeout(() => {
+          setErrorMessage('');
+        }, 3000);
+      }
+    });
```
OleksiiBlezniuk, line 176: Shouldn't we await until this promises are resolved?

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,51 @@
+import classNames from 'classnames';
+import { FilterType } from '../../types/FilterEnum';
+
+type Props = {
+  activeTodos: number;
+  completedTodos: number
+  selectedFilter: FilterType
+  onSelectFilter: (filterType: FilterType) => void
+  onDeleteComplete: () => void
```
OleksiiBlezniuk, line 9:
```suggestion
  completedTodos: number,
  selectedFilter: FilterType,
  onSelectFilter: (filterType: FilterType) => void,
  onDeleteComplete: () => void,
```

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,31 @@
+import classNames from 'classnames';
+
+type Props = {
+  errorMessage: string
+  onDelete: () => void
+};
```
OleksiiBlezniuk, line 6:
```suggestion
type Props = {
  errorMessage: string,
  onDelete: () => void,
};
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,51 @@
...
+};
+
+export const Footer: React.FC<Props> = ({
+  onDeleteComplete,
+  onSelectFilter,
+  completedTodos,
+  selectedFilter,
+  activeTodos,
+}) => {
+  return (
```
OleksiiBlezniuk, line 19:
```suggestion
}) => (
```

src/App.tsx
```diff
@@ -1,24 +1,228 @@
...
+import { Header } from './components/Header/Header';
+import { Footer } from './components/Footer';
 
-const USER_ID = 0;
+const USER_ID = 7006;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [title, setTitle] = useState('');
```
OleksiiBlezniuk, line 26: why do you need this state in this component? move form related state to component with state and just pass function that should be run when form is submited

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,108 @@
...
+        { completed },
+        { editing: isEditing },
+      )
+    }
+    >
+      <label className="todo__status-label">
+        <input
+          type="checkbox"
+          className="todo__status"
+          onClick={() => onUpdate(id, { completed: !completed })}
```
OleksiiBlezniuk, line 66: maybe onChange?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+      {isEditing ? (
+        <input
+          type="text"
+          className="todo__title-field"
+          value={todoTitle}
+          onChange={handleTitleChange}
+          onBlur={handleTitleSubmit}
+          onKeyDown={handleKeyDown}
+        />
```
OleksiiBlezniuk, line 78: Why do you have input without form?

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+        case FilterType.Active:
+          return !completed;
+
+        case FilterType.Completed:
+          return completed;
+
+        default:
+          return true;
+      }
```
OleksiiBlezniuk, line 37: Better to move it to separate function outside component

OleksiiBlezniuk (changes requested):
Input should be focused on double click
Also add loader when active todos are removing by `clear active` button
https://user-images.githubusercontent.com/84275496/233627535-ff144c29-ce6b-4455-a563-b78f7b5aa077.mov

juljaVysotska (changes requested):
Good job! I like your code! I found some moments in your work, fix them. Furthermore, I'd recommend you read more about TS, especially about any. Best regards 💛

PR #449, opened 2023-04-13
todo-app-with-api done

Stepan-TS:
[DEMO LINK](Stepan-TS.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,211 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
+import {
+  getTodos,
+  deleteTodo,
+  postTodo,
+  updateTodo,
+} from './api/todos';
+import { ErrorMessage } from './components/ErrorMessage/ErrorMessage';
+import { TodoFilter } from './components/TodoFilter/TodoFilter';
+import { Header } from './components/Header/Header';
+import { Loader } from './components/Loader/Loader';
+import { TodoList } from './components/TodoList/TodoList';
+import { USER_ID } from './types';
+import { TodosError } from './types/Errors';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { filterTodos } from './helpers';
+import { SortTodoBy } from './types/SortTodoBy';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const [sortTodosBy, setSortTodosBy] = useState(SortTodoBy.Default);
+  const [loadingTodosIds, setLoadingTodosIds] = useState<number[]>([]);
+
+  const [isLoading, setIsLoading] = useState(false);
+  const [loadingError, setLoadingError] = useState(false);
+  const [errorText, setErrorText] = useState('');
+
+  const [disableField, setDisableField] = useState(false);
+
+  const visiableTodos = filterTodos(todos, sortTodosBy);
+  const completedTodos = useCallback(() => {
+    return todos.filter(todo => todo.completed);
+  }, [todos]);
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const hasActiveTodos = activeTodos.length !== 0;
+  const hasCompletedTodos = completedTodos().length !== 0;
+
+  const handleShowError = (text: string) => {
+    setLoadingError(true);
+    setErrorText(text);
+
+    setTimeout(() => {
+      setLoadingError(false);
+    }, 3000);
+  };
+
+  const handleCloseError = () => {
+    setLoadingError(false);
+  };
+
+  const loadTodos = (userId: number) => {
+    setIsLoading(true);
+    setLoadingError(false);
+
+    getTodos(userId)
+      .then(loadedTodos => setTodos(loadedTodos))
+      .catch(() => {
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,24 +1,211 @@
...
-
-const USER_ID = 0;
+import { filterTodos } from './helpers';
+import { SortTodoBy } from './types/SortTodoBy';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const [sortTodosBy, setSortTodosBy] = useState(SortTodoBy.Default);
```
OleksiiBlezniuk, line 25: We filtering todos, not sorting them

src/App.tsx
```diff
@@ -1,24 +1,211 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+  const [loadingError, setLoadingError] = useState(false);
+  const [errorText, setErrorText] = useState('');
+
+  const [disableField, setDisableField] = useState(false);
+
+  const visiableTodos = filterTodos(todos, sortTodosBy);
+  const completedTodos = useCallback(() => {
+    return todos.filter(todo => todo.completed);
+  }, [todos]);
```
OleksiiBlezniuk, line 37:
```suggestion
  const completedTodos = useMemo(() => {
    return todos.filter(todo => todo.completed);
  }, [todos]);
```

OleksiiBlezniuk: Why do you need to save it to callback?

src/App.tsx
```diff
@@ -1,24 +1,211 @@
...
+  const activeTodos = todos.filter(todo => !todo.completed);
```
OleksiiBlezniuk, line 38: useMemo?

src/App.tsx
```diff
@@ -1,24 +1,211 @@
...
+  const hasActiveTodos = activeTodos.length !== 0;
+  const hasCompletedTodos = completedTodos().length !== 0;
+
+  const handleShowError = (text: string) => {
```
OleksiiBlezniuk, line 42:
```suggestion
  const handleShowError = (text: TodosError) => {
```

src/App.tsx
```diff
@@ -1,24 +1,211 @@
...
+    setLoadingError(true);
+    setErrorText(text);
```
OleksiiBlezniuk, line 44: you don't need boolean state for error, you already have state with message - just create TodosError.None and set it by default

src/App.tsx
```diff
@@ -1,24 +1,211 @@
...
+
+      try {
+        setDisableField(true);
+        setTempTodo({
+          ...newTodo,
+          id: 0,
+        });
+
+        await postTodo(newTodo)
+          .then((response) => setTodos([...todos, response]));
```
OleksiiBlezniuk, line 89: no, if you using async/await - use only it, don't combine

src/App.tsx
```diff
@@ -1,24 +1,211 @@
...
+        id,
+      ]);
+
+      handleDeleteTodo(id)
+        .then(() => {
+          setTodos(
+            (prevTodos) => prevTodos.filter(todo => !todo.completed),
+          );
+        })
+        .finally(() => setLoadingTodosIds([]));
```
OleksiiBlezniuk, line 126: use async/await

OleksiiBlezniuk, line 126: I guess it should be set after all todos will be removed, not after first resolver/rejected promise
or remove only id of todo that you removed here

src/App.tsx
```diff
@@ -1,24 +1,211 @@
...
+
+      const updatedTodo = await updateTodo(todoId, property);
+
+      setTodos((prevTodos) => prevTodos
+        .map((prevTodo) => {
+          return prevTodo.id === todoId
+            ? {
+              ...prevTodo,
+              ...updatedTodo,
+            } : prevTodo;
```
OleksiiBlezniuk, line 142:
```suggestion
            ? {
              ...prevTodo,
              ...updatedTodo,
            }
            : prevTodo;
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+          href="#/completed"
+          className={classNames(
+            'filter__link',
+            { selected: sortBy === SortTodoBy.Completed },
+          )}
+          onClick={() => changeSortBy(SortTodoBy.Completed)}
+        >
+          Completed
+        </a>
+      </nav>
```
OleksiiBlezniuk, line 61: Let's map enum instead 😉

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,131 @@
...
+            <input
+              type="text"
+              className="todo__title-field"
+              ref={inputRef}
+              value={newTitle}
+              onChange={(event) => setNewTitle(event.target.value)}
+              onBlur={handleBlur}
+              onKeyUp={handleCancelEditing}
+            />
+          </form>
```
OleksiiBlezniuk, line 99: Would be nice to move it to new component

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,43 @@
...
+  } = props;
+
+  return (
+    <section className="todoapp__main">
+      {todos.map(todo => (
+        <TodoItem
+          key={todo.id}
+          todo={todo}
+          onDelete={onDeleteTodo}
+          loadingTodosIds={loadingTodosIds}
```
OleksiiBlezniuk, line 28: instead of passing all array of loading arrays - pass `isLoading` to define should loader be displayed here

OleksiiBlezniuk (changes requested):

PR #454, opened 2023-04-15
done

Vadivel-Babu:
[DEMO LINK](https://Vadivel-Babu.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,208 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useCallback, useEffect, useMemo, useState } from "react";
+import { UserWarning } from "./UserWarning";
+import { getTodos, postTodo, deleteTodo, patchTodo } from "./api/todos";
+import { Todos } from "./types/todo";
+import { TodoList } from "./Components/TodoList/TodoList";
+import { Error } from "./Components/Error/Error";
+import { FilterTodo } from "./Components/FilterTodo/FilterTodo";
+import { FilterStatus } from "./types/FilterStatus";
+import { Header } from "./Components/Header/Header";
 
-const USER_ID = 0;
+const USER_ID = 6429;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todos[]>([]);
+  const [error, setError] = useState("");
+  const [title, setTitle] = useState("");
+  const [temporaryTodo, setTemporaryTodo] = useState<Todos>();
+  const [filter, setFilter] = useState<FilterStatus>(FilterStatus.all);
+  const activeTodosСount = useMemo(() => {
+    return todos.filter(({ completed }) => !completed).length;
+  }, [todos]);
+  const completedTodos = useMemo(() => {
+    return todos.filter(({ completed }) => completed).length;
+  }, [todos]);
+
+  useEffect(() => {}, [error]);
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((result: React.SetStateAction<Todos[]>) => setTodos(result))
+      .catch(() => setError("Unable to load the todos"));
+  }, []);
+
+  const currentTodos = useMemo(() => {
+    return todos.filter((todo) => {
+      switch (filter) {
+        case FilterStatus.active:
+          return !todo.completed;
+
+        case FilterStatus.completed:
+          return todo.completed;
+
+        default:
+          return todo;
+      }
+    });
+  }, [todos, filter]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+  const addTodo = (todoTitle: string) => {
+    const newTodo = {
+      title: todoTitle,
+      userId: USER_ID,
+      completed: false,
+    };
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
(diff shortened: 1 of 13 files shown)


src/types/todo.ts
```diff
@@ -0,0 +1,6 @@
+export interface Todos {
```
ivanchuk-yurii, line 1:
```suggestion
export interface TodoInterface {
```
it's singular

src/utils/fetchClients.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
ivanchuk-yurii, line 1: don't leave comments or disable eslint

src/Components/Error/Error.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  return (
+    <div
+      className={classNames(
+        "notification is-danger is-light has-text-weight-normal",
+        { hidden: error.length === 0 }
+      )}
+    >
+      <button type="button" className="delete" onClick={onClear} hidden={false}>
+        {"delete "}
```
ivanchuk-yurii, line 26:
```suggestion
        delete
`
is it work like here?

File: src/App.tsx [hunk 4]
```diff
@@ -1,24 +1,208 @@
...
+
+      <div className="todoapp__content">
+        <Header
+          onSubmitForm={handleSubmit}
+          title={title}
+          onChangeTitle={changeHandler}
+          onToggleAllCompleted={toggleAllCompleted}
+        />
+
+        {todos.length > 0 && (
```
ivanchuk-yurii, line 173:
```suggestion
        {todos.length && (
```

ivanchuk-yurii (changes requested):

src/App.tsx
Vadivel-Babu: if I write todos.length its shows  0 when there is no list but in the case  of  todos.length > 0 if there is no list it doesn't show anything

src/utils/fetchClients.ts
Vadivel-Babu: if I remove this commend its warning about Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

src/Components/Error/Error.tsx
Vadivel-Babu: if I remove the text from the button element it shows the error  A control must be associated with a text label  jsx-a11y/control-has-associated-label

src/utils/fetchClients.ts
ivanchuk-yurii: so fix the error)

ivanchuk-yurii: so fix the error)

src/Components/Error/Error.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  return (
+    <div
+      className={classNames(
+        "notification is-danger is-light has-text-weight-normal",
+        { hidden: error.length === 0 }
+      )}
+    >
+      <button type="button" className="delete" onClick={onClear}>
+        {""}
```
ivanchuk-yurii, line 26:
```suggestion
        delete
```
why did you do this?

src/utils/fetchClients.ts
```diff
@@ -0,0 +1,45 @@
...
+  });
+}
+
+// To have autocompletion and avoid mistypes
+type RequestMethod = "GET" | "POST" | "PATCH" | "DELETE";
+
+function request<T>(
+  url: string,
+  method: RequestMethod = "GET",
+  data: any = null // we can send any data to the server
```
ivanchuk-yurii, line 16:
```suggestion
  data: unknown = null // we can send any data to the server
```

ivanchuk-yurii (changes requested):

src/App.tsx
```diff
@@ -1,24 +1,210 @@
...
+                <button
+                  type="button"
+                  className="todoapp__clear-completed"
+                  onClick={onDeleteComplete}
+                >
+                  Clear completed
+                </button>
+              ) : (
+                ''
+              )}
```
Anastasiia-Svintsova, line 201: Why do you need a ternary operator here? 
```suggestion
              {completedTodos && (
                <button
                  type="button"
                  className="todoapp__clear-completed"
                  onClick={onDeleteComplete}
                >
                  Clear completed
                </button>
              )}
```

Anastasiia-Svintsova (changes requested):
- The checkmark in the input should change the style if all todos are checked. And don't show the checkmark if there are no todos in the list

https://user-images.githubusercontent.com/99403198/232446491-2f329b8b-07b3-4e76-86d5-0bb6eea2ac7d.mov

- Look how it works in the working version 

https://user-images.githubusercontent.com/99403198/232446779-e639a58a-113b-40d4-bc18-dc44acbd0e37.mov

- Check my comments to your previous [PR](https://github.com/mate-academy/react_todo-app-add-and-delete/pull/402) as well 

- Don't show the old value after editing 

https://user-images.githubusercontent.com/99403198/232448274-11747774-0679-4803-97ed-605137685809.mov

src/App.tsx
```diff
@@ -1,24 +1,225 @@
...
+    return todos.filter(({ completed }) => completed);
+  }, [todos]);
+
+  const resetError = () => {
+    const timeoutID = setTimeout(() => setError(""), 3000);
+
+    return () => {
+      clearTimeout(timeoutID);
+    };
+  };
```
anstsot, line 44: You can use `useEffect` instead of calling this function after `error` changing
```suggestion
  const resetError = () => {
    const timeoutID = setTimeout(() => setError(""), 3000);

    return () => {
      clearTimeout(timeoutID);
    };
  };
  
  useEffect(() => {
    if (error.length > 0) resetError();
  }, [error]);
```

anstsot (approved):
Great job, approved!

polosanya (commented):
It would be great to fix all unresolved previous comments. Also, make sure that you added all needed files to the commit)

PR #467, opened 2023-04-23
Solution

Viktorianeimesh:
[DEMO LINK](https://Viktorianeimesh.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/types/Sort.ts
+++ b/src/types/Sort.ts
@@ -1,0 +1,5 @@
+export enum Sort {
+  all = 'All',
+  active = 'Active',
+  completed = 'Completed',
+}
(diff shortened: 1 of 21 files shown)


src/types/Sort.ts
```diff
@@ -0,0 +1,5 @@
+export enum Sort {
+  all = 'All',
+  active = 'Active',
+  completed = 'Completed',
```
polosanya, line 4:
```suggestion
export enum Sort {
  All = 'all',
  Active = 'active',
  Completed = 'completed',
```
Enum properties should be named in PascalCase

polosanya (changes requested):

polosanya (approved):
Well done!

PR #476, opened 2023-05-03
Ready

sergik94:
https://sergik94.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,339 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState, useMemo } from 'react';
 import { UserWarning } from './UserWarning';
+import { Header } from './components/Header';
+import { Todos } from './components/Todos';
+import { Footer } from './components/Footer';
+import { Notification } from './components/Notification';
 
-const USER_ID = 0;
+import {
+  getTodos, postTodos, deleteTodo, patchTodos,
+} from './api/todos';
+
+import { Todo } from './types/Todo';
+import { FilteredBy } from './types/FilteredBy';
+
+const USER_ID = 9964;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [filteredBy, setFilteredBy] = useState<FilteredBy>(FilteredBy.All);
+  const [isLoadingError, setLoadingError] = useState(false);
+  const [isAddTodoError, setAddTodoError] = useState(false);
+  const [isTodoDeleteError, setTodoDeleteError] = useState(false);
+  const [isTodoUpdatingError, setTodoUpdatingError] = useState(false);
+  const [isTitleEmpty, setTitleEmpty] = useState(false);
+  const [isTodoAdded, setTodoAdded] = useState(false);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [updatingTitle, setUpdatingTitle] = useState('');
+  const [targetTodosIds, setTargetTodosIds] = useState<number[]>([]);
+  const isActive = todos?.some(todo => !todo.completed);
+  const newTodo = {
+    id: 0,
+    title: newTodoTitle,
+    userId: USER_ID,
+    completed: false,
+  };
+
+  async function getTodoList() {
+    try {
+      const todoList = await getTodos(USER_ID);
+
+      setTodos(todoList);
+    } catch (error) {
+      setLoadingError(true);
+    }
+  }
+
+  const postNewTodo = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    setAddTodoError(false);
+
+    if (newTodoTitle !== '') {
+      setTodoAdded(true);
+      try {
+        const todo = await postTodos(USER_ID, newTodo);
+
+        setTodos((currTodo) => {
+          if (currTodo) {
+            return currTodo.concat(todo);
+          }
+
+          return todo;
+        });
+
+        setNewTodoTitle('');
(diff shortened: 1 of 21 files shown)


src/App.tsx
```diff
@@ -1,24 +1,339 @@
...
+
+import { Todo } from './types/Todo';
+import { FilteredBy } from './types/FilteredBy';
+
+const USER_ID = 9964;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [filteredBy, setFilteredBy] = useState<FilteredBy>(FilteredBy.All);
+  const [isLoadingError, setLoadingError] = useState(false);
```
YegorKorenkov, line 20: Why do you need so many states for an error? If you can leave one boolean state that will indicate the presence of an error and a string state for the error message?

src/App.tsx
```diff
@@ -1,24 +1,339 @@
...
+    if (oldTitle === updatingTitle) {
+      return;
+    }
+
+    setTargetTodosIds(currIds => {
+      return [...currIds, todoId || 0];
+    });
+    if (updatingTitle !== '') {
+      try {
+        setTodos((currTodo) => {
```
YegorKorenkov, line 234: You first update the state, and then make a request, and this is wrong, todo should't update if the request passed with an error

src/components/Todos/Todos.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+            {editedTodoId === todo.id ? (
+              <form>
+                <input
+                  type="text"
+                  className="todo__title-field"
+                  placeholder="Empty todo will be deleted"
+                  value={updatingTitle}
+                  onChange={(event) => setUdatingTitle(event.target.value)}
+                  onKeyDown={(e) => {
```
YegorKorenkov, line 63: Please create separate functions for all events. Firstly, you trash jsx, and secondly, these functions are anonymous and they are recreated on each re-render

src/components/Todos/Todos.tsx
```diff
@@ -0,0 +1,127 @@
...
+  isTodoAdded,
+  removeTodo,
+  targetTodosIds,
+  changeTodoStatus,
+  callEditeMode,
+  updatingTitle,
+  setUdatingTitle,
+  submitTitleUpdating,
+}) => {
+  const [editedTodoId, setEditedTodoId] = useState(0);
```
YegorKorenkov, line 32: Correctly to set null, if you have a todo with id 0, will the editing input be immediately visible?

YegorKorenkov (changes requested):

mykhalenych (changes requested):

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,24 +1,317 @@
...
+  const [updatingTitle, setUpdatingTitle] = useState('');
+  const [targetTodosIds, setTargetTodosIds] = useState<number[]>([]);
+  const isActive = todos?.some(todo => !todo.completed);
+
+  const newTodo = {
+    id: 0,
+    title: newTodoTitle,
+    userId: USER_ID,
+    completed: false,
+  };
```
Viktor-Kost, line 33: move this object inside postNewTodo method

src/App.tsx
```diff
@@ -1,24 +1,317 @@
...
+      try {
+        const todo = await postTodos(USER_ID, newTodo);
+
+        setTodos((currTodo) => {
+          if (currTodo) {
+            return currTodo.concat(todo);
+          }
+
+          return todo;
+        });
```
Viktor-Kost, line 59: you can use spread operator

src/App.tsx
```diff
@@ -1,24 +1,317 @@
...
+          todosForUpdaiting.push(patchTodos(todo.id, { completed: true }));
+          targetTodoIds.push(todo.id);
+        }
+      });
+    } else {
+      todos?.forEach(todo => {
+        todosForUpdaiting.push(patchTodos(todo.id, { completed: false }));
+        targetTodoIds.push(todo.id);
+      });
+    }
```
Viktor-Kost, line 185: you can simplify sth like this 

```
  todos?.forEach(todo => {
    const completed = isActive ? true : false;
    todosForUpdaiting.push(patchTodos(todo.id, { completed }));
    targetTodoIds.push(todo.id);
  });

```

Viktor-Kost (changes requested):
Good job!
Some minor fixes left

src/App.tsx
sergik94: yes, I can. But in this case we will update all todos, not only active. We can try to simplify this part of code, but this simplification will not be very significant.

src/App.tsx
```diff
@@ -1,24 +1,317 @@
...
+    };
+
+    if (newTodoTitle !== '') {
+      setTodoAdded(true);
+      try {
+        const todo = await postTodos(USER_ID, newTodo);
+
+        setTodos((currTodo) => {
+          if (currTodo) {
+            return [...currTodo, ...todo];
```
VitaliyBondarenko1982, line 55:
```suggestion
            return [...currTodo, todo];
```
after adding new todo app is failed
looks like problem here
Screenshot from 2023-05-04 13-15-01

VitaliyBondarenko1982 (changes requested):

sergik94:
image
I have never had such a large list of reviewers))

mykhalenych (changes requested):
fix items calculation, check demo example
https://mate-academy.github.io/react_todo-app-with-api

https://user-images.githubusercontent.com/53228259/236185097-d0fd3b26-38b4-47d4-b918-3320de07b259.mov

mykhalenych (approved):

PR #480, opened 2023-05-07
solution

katerynashylina:
[DEMO LINK](https://katerynashylina.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoItem.tsx
+++ b/src/components/TodoItem.tsx
@@ -1,0 +1,130 @@
+import classNames from 'classnames';
+import { useState } from 'react';
+import { Todo } from '../types/Todo';
+import { patchTodo } from '../api/todos';
+
+type Props = {
+  todo: Todo,
+  onDelete: (todoId: number) => void,
+  updateTodo: (updatedTodo: Todo) => void,
+  removeLoadingTodo: (todoId: number) => void,
+  addLoadingTodo: (todoId: number) => void,
+  loadingTodos: number[],
+  deletingTodoId: number | null,
+  isLoading: boolean,
+  updateTodoOnServer: (todoId: number, data: Partial<Todo>) => Promise<void>,
+};
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  onDelete,
+  updateTodo,
+  removeLoadingTodo,
+  addLoadingTodo,
+  loadingTodos,
+  deletingTodoId,
+  isLoading,
+  updateTodoOnServer,
+}) => {
+  const { id, title, completed } = todo;
+
+  const [checked, setChecked] = useState(completed);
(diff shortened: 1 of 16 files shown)


src/utils/fetchClient.ts
```diff
@@ -0,0 +1,50 @@
+import { Todo } from '../types/Todo';
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
```
OleziO, line 6: Remove all comments from the code.

OleziO (changes requested):

src/components/TodoList.tsx
```diff
@@ -0,0 +1,43 @@
...
+        removeLoadingTodo={removeLoadingTodo}
+        addLoadingTodo={addLoadingTodo}
+        loadingTodos={loadingTodos}
+        deletingTodoId={deletingTodoId}
+        isLoading={isLoading}
+        updateTodoOnServer={updateTodoOnServer}
+      />
+    ))}
+  </div>
+);
```
DarkMistyRoom, line 43: you just pass all this props throughout this component without using them. Consider using `Context`

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,130 @@
...
+  removeLoadingTodo,
+  addLoadingTodo,
+  loadingTodos,
+  deletingTodoId,
+  isLoading,
+  updateTodoOnServer,
+}) => {
+  const { id, title, completed } = todo;
+
+  const [checked, setChecked] = useState(completed);
```
DarkMistyRoom, line 31: this state duplicates the prop. It's a bad practice. We should use only one source of truth

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,130 @@
...
+          <form onSubmit={handleTitleChange}>
+            <input
+              type="text"
+              className="todo__title-field"
+              value={newTitle}
+              onChange={(event) => setNewTitle(event.target.value)}
+              onBlur={() => handleTitleChange()}
+              onKeyUp={handleKeyUp}
+              // eslint-disable-next-line jsx-a11y/no-autofocus
+              autoFocus
```
DarkMistyRoom, line 98: better to use useRef for autofocusing

DarkMistyRoom (changes requested):
Good job! Let's improve it)
there are errors in adding new todo.
image

looks like the reason is here
image
you send todo with id, but the server should set id by itself

the counter should show the non-completed todos amount. It doesn't depend on the filter
image

OleziO (approved):

PR #507, opened 2023-05-16
Develop

Zad0rnik:
[DEMO LINK](https://Zad0rnik.github.io/react_todo-app-with-api/)


jstmpelowycz (changes requested):
one missing piece of the function — it's possible to update a todo with an empty title

Zad0rnik:
> one missing piece of the function — it's possible to update a todo with an empty title

How can I fix this, I don`t understand how can I add empty todo, because it`s unable at my app)

PR #523, opened 2023-05-18
finish

Rage1502:
[DEMO LINK](https://Rage1502.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,155 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import classNames from 'classnames';
+import { useCallback, useEffect } from 'react';
+import { deleteTodo, getTodos, updateTodo } from './api/todos';
+import { Error } from './components/ErrorNotification';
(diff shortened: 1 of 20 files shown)


src/App.tsx
```diff
@@ -1,24 +1,155 @@
...
+      });
+    } catch (error) {
+      setError('Error while updating todos');
+    } finally {
+      setTodoIdsInUpdating([]);
+    }
+  }, [hasActiveTodos, todos, updateTodo]);
+
+  useEffect(() => {
+    getTodos(USER_ID)
```
SerhiiKirik, line 98: if you use `async` function & then, catch- it would be better to use just one of the ways
1. then, catch
2. async / await

src/App.tsx
```diff
@@ -1,24 +1,155 @@
...
+            />
+          )}
+
+          <CreateTodoForm />
+        </header>
+
+        <TodoList />
+
+        {hasTodos && (
+          <footer className="todoapp__footer">
```
SerhiiKirik, line 133: move footer to the separate component

PR #529, opened 2023-05-20
Solution

yana-kotsulym:
https://yana-kotsulym.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,243 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+//* eslint-disable jsx-a11y/control-has-associated-label */
+import React, {
+  useEffect,
+  useState,
+  useMemo,
+  useCallback,
+  FormEvent,
+} from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  getTodos,
+  addTodo,
+  removeTodo,
+  toggleTodo,
+  changeTodoTitle,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Filter, ErrorType } from './utils/enums';
+import { TodoList } from './components/TodoList';
+import { TodoItem } from './components/TodoItem';
 
-const USER_ID = 0;
+const USER_ID = 10380;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<ErrorType | null>(null);
+  const [filterStatus, setFilterStatus] = useState<Filter>(Filter.All);
+  const [query, setQuery] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,24 +1,243 @@
...
+            ...todo,
+            completed,
+          };
+        }));
+        setProcessedTodos(currState => currState.filter(id => id !== todoId));
+      })
+      .catch(() => setError(ErrorType.Patch));
+  };
+
+  const changeDataonServer = (todoId: number, title: string) => {
```
TarasHoliuk, line 101:
```suggestion
  const changeTodoTitle = (todoId: number, title: string) => {
```
as for me `data` is not clear enough
the same for `toggleTodoStatus` - but not critical - up to you

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,39 @@
...
+        hidden: !error,
+      },
+    )}
+    >
+      <button
+        type="button"
+        className="delete"
+        onClick={() => onError(null)}
+      />
+      {error === ErrorType.Get && 'Unable to load a todo'}
```
TarasHoliuk, line 32: create `getErrorMessage` function and use `switch` to get the correct message

src/components/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+      <form
+        onSubmit={onSubmit}
+      >
+        <input
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          value={query}
+          onChange={(e) => onChange(e.target.value)}
```
TarasHoliuk, line 52: it's better to create a separate function instead of passing anonymous

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+        />
+      </label>
+
+      {willUpdate
+        ? (
+          <form
+            onSubmit={() => {
```
TarasHoliuk, line 52: create corresponding functions above for each handler

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+};
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  onClose,
+  processedTodos,
+  onToggle,
+  onChange,
+}) => {
+  const [willUpdate, setWillUpdate] = useState(false);
```
TarasHoliuk, line 22: this name is not clear to me
what about smth like `isEditMode`

TarasHoliuk (changes requested):

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,62 @@
...
+        type="button"
+        className="delete"
+        onClick={() => onError(null)}
+      />
+      {getErrorMessage(error)}
+      {/* {error === ErrorType.Get && 'Unable to load a todo'}
+      {error === ErrorType.Post && 'Unable to add a todo'}
+      {error === ErrorType.Delete && 'Unable to delete a todo'}
+      {error === ErrorType.Patch && 'Unable to update a todo'}
+      {error === ErrorType.isEmpty && 'Title can\'t be empty'} */}
```
SanyaBratashchuk, line 59: Do not leave comments

SanyaBratashchuk (approved):

PR #534, opened 2023-05-21
Todo api v1

FallenMAD:
[DEMO LINK](https://FallenMAD.github.io/react_todo-app-with-api/)
And i have a problem with style in TodoItem component with onBlur

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -47,8 +47,7 @@ export const App: React.FC = () => {
         setTodos(todosData);
       }
     } catch (error) {
-      setErrorMessage("Can't create a todo");
-      throw new Error('Error in loadData');
+      setErrorMessage("Can't load todos");
     }
   };
 
@@ -94,7 +93,6 @@ export const App: React.FC = () => {
       setQuery('');
     } catch {
       setErrorMessage('Unable to add a todo');
-      throw new Error('Error in Header');
     } finally {
       setTempTodo(null);
     }
@@ -109,7 +107,6 @@ export const App: React.FC = () => {
       setTodos(todos.filter(todo => todo.id !== todoId));
     } catch {
       setErrorMessage('Unable to delete a todo');
-      throw new Error('Error in Delete Method');
     } finally {
       setActiveTodoId(0);
     }
@@ -143,8 +140,8 @@ export const App: React.FC = () => {
     todoId: number,
     completed: boolean,
   ) => {
-    setActiveTodoId(todoId);
     try {
+      setActiveTodoId(todoId);
       const newTodo = await patchTodo(todoId, { completed });
 
       setTodos((prev: Todo[]) => {
--- a/src/components/TodoItem/TodoItem.tsx
+++ b/src/components/TodoItem/TodoItem.tsx
@@ -39,6 +39,7 @@ export const TodoItem: React.FC<Props> = ({
     onChangeTodoTitle(id, query);
 
     if (query === title) {
+      setIsEdit(false);
       setQuery(title);
     }
 


src/App.tsx
```diff
@@ -1,24 +1,242 @@
...
+  const [status, setStatus] = useState(StatusTodos.ALL);
+  const [isError, setIsError] = useState(false);
+  const [query, setQuery] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isTodoId, setIsTodoId] = useState<number[] | null>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const isTodoCompleted = todos.every(todo => todo.completed);
+
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
```
witflash, line 29: You should know you can use an implicit return for the such cases. It's not necessarily but your code will be shorter.

src/App.tsx
```diff
@@ -1,24 +1,242 @@
...
+import { Notification } from './components/Notification';
+import { StatusTodos } from './types/StatusTodo';
 
-const USER_ID = 0;
+const USER_ID = 9946;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState(StatusTodos.ALL);
+  const [isError, setIsError] = useState(false);
```
witflash, line 21: You can avoid using this state. Use just the `errorMessage` state instead of these two separate.
So you can check if `errorMessage` is not empty and render your Notification component.

src/App.tsx
```diff
@@ -1,24 +1,242 @@
...
+            visibleTodos={todos}
+            status={status}
+            onStatusChange={setStatus}
+            onChangeTodo={setTodos}
+          />
+        )}
+
+      </div>
+
+      {isError && (
```
witflash, line 232: Just check `errorMessage` statement

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export const Notification: React.FC<Props> = ({
+  error,
+  deleteError,
+  errorMessage,
+}) => {
+  return (
+    <div
+      className={classNames(
+        'notification is-danger is-light has-text-weight-normal', {
```
witflash, line 17: If you use `classnames` lib, you can pass all the classes as separate arguments: `classNames('notification', 'is-danger', ...)`. It's more readable I think.

src/App.tsx
```diff
@@ -1,24 +1,242 @@
...
+        <Notification
+          errorMessage={errorMessage}
+          error={isError}
+          deleteError={setIsError}
```
witflash, line 236: and here you can clear `errorMessage`

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,74 @@
...
+  onChangeTodo: (value: Todo[]) => void;
+};
+
+export const TodoFilter: React.FC<Props> = ({
+  visibleTodos,
+  status,
+  onStatusChange,
+  onChangeTodo,
+}) => {
+  const activeTodos = visibleTodos.filter(todo => !todo.completed).length;
```
witflash, line 18: `const activeTodosLength`

src/App.tsx
```diff
@@ -1,24 +1,242 @@
...
-const USER_ID = 0;
+const USER_ID = 9946;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState(StatusTodos.ALL);
+  const [isError, setIsError] = useState(false);
+  const [query, setQuery] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isTodoId, setIsTodoId] = useState<number[] | null>([]);
```
witflash, line 24: Do you really need an Array type here?
As I see, this state is about just one active Todo, is't it?
If I'am right it may be something like `activeTodoId: number`

witflash (changes requested):
Good job!
Let's just make a little bit improvements.

src/App.tsx
FallenMAD: pls check my first comments👆🏻

src/App.tsx
```diff
@@ -1,24 +1,230 @@
...
+    try {
+      const todosData = await getTodos(USER_ID);
+
+      if ('Error' in todosData) {
+        throw new Error('Error in TodosData');
+      } else {
+        setTodos(todosData);
+      }
+    } catch (error) {
+      setErrorMessage("Can't create a todo");
```
agniya-i, line 50: Better to change the error message to "Can't load todos" to accurately reflect the failed operation.

You can also create an enum for error messages

src/App.tsx
```diff
@@ -1,24 +1,230 @@
...
+      throw new Error('Error in loadData');
```
agniya-i, line 51: Do we need  `throw new Error()` in the catch block?

agniya-i (changes requested):
Looks good to me, if you have any questions better ask in Slack `fe_chat` 🙌

OleziO (approved):

PR #536, opened 2023-05-21
solution

yapelenua:
[DEMO LINK](https://yapelenua.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -1,0 +1,23 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+// Add more methods here
+
+export const addTodo = (todo: Todo) => {
+  return client.post<Todo>('/todos', todo);
+};
+
+export const deleteTodo = (id: number) => {
+  return client.delete(`/todos/${id}`)
+    .catch(() => {
+      throw new Error('Unable to delete a todo');
+    });
+};
+
+export const updateTodo = (id: number, todo: Todo) => {
+  return client.patch<Todo>(`/todos/${id}`, todo);
+};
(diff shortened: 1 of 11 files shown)


src/components/ERROR.tsx
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import { FC } from 'react';
+
+interface Props {
+  errorMessage: string | null;
+  setErrorMessage: React.Dispatch<React.SetStateAction<string | null>>;
+}
+
+export const Eror: FC<Props> = ({ setErrorMessage, errorMessage }) => {
```
jstmpelowycz, line 9:
```suggestion
export const ErrorMessage: FC<Props> = ({ setErrorMessage, errorMessage }) => {
```

src/components/ERROR.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <div className="notification is-danger is-light has-text-weight-normal">
+      <button
+        type="button"
+        className="delete"
+        onClick={() => setErrorMessage(null)}
+      />
+      {`Unable to create a todos, because ${errorMessage}`}
```
jstmpelowycz, line 17: it's only about creating, there may be other errors

src/api/todos.ts
```diff
@@ -0,0 +1,23 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+// Add more methods here
```
jstmpelowycz, line 8: do not keep such kind of comments
```suggestion
```

jstmpelowycz (changes requested):
missing functional:
- gray box-shadow persists on recent elements users interact with, but shouldn't
- nothing happens on blur (it may seem to be OK, but if to reload the page, todos are still not updated)
- you should handle deleting a todo being updated if new title is empty
- `toggleAll` seems not to be implemented

PR #537, opened 2023-05-21
Develop

anastasia-radchencko:
https://anastasia-radchenckogithub.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,7 +7,12 @@ import React, {
   useMemo,
 } from 'react';
 import classNames from 'classnames';
-import { addTodo, deleteTodo, getTodos } from './api/todos';
+import {
+  addTodo,
+  deleteTodo,
+  getTodos,
+  patchTodo,
+} from './api/todos';
 import { Todo } from './types/Todo';
 import { TodoList } from './components/TodoList/TodoList';
 import { TodoFilter } from './components/TodoFilter/TodoFilter';
@@ -52,6 +57,64 @@ export const App: React.FC = () => {
       setErrorMessage('Unable to add todo');
     }
   }, []);
+
+  const updateTodoCompleted = useCallback(
+    async (todoId: number, completed: boolean) => {
+      const todoTemp = [...todos];
+
+      setTodos(prevTodos => prevTodos.map((todo) => {
+        if (todo.id !== todoId) {
+          return todo;
+        }
+
+        return { ...todo, completed };
+      }));
+
+      try {
+        await patchTodo(todoId, { completed });
+      } catch {
+        setErrorMessage('Unable to update todo');
+        setTodos(todoTemp);
+      }
+    }, [todos],
+  );
+
+  const updateTitle = useCallback(
+    async (todoId: number, titles: string) => {
+      const todoTemp = [...todos];
+
+      setTodos(prevTodos => prevTodos.map((todo) => {
+        if (todo.id !== todoId) {
+          return todo;
+        }
+
+        return { ...todo, title: titles };
+      }));
+
+      try {
+        await patchTodo(todoId, { title: titles });
+      } catch {
+        setErrorMessage('Unable to update todo');
+        setTodos(todoTemp);
+      }
+    }, [todos],
+  );
+
+  const handleUpdateAllTodo = useCallback(() => {
+    if (!activeTodo) {
+      todos.forEach(todo => {
+        updateTodoCompleted(todo.id, !todo.completed);
+      });
+
+      return;
+    }
+
+    todos.forEach(todo => {
+      if (!todo.completed) {
+        updateTodoCompleted(todo.id, !todo.completed);
+      }
+    });
+  }, [todos]);
 
   const onDeleteError = useCallback(
     async () => setErrorMessage(''), [errorMessage],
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,24 +1,244 @@
...
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sort, setSort] = useState(SortType.All);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [title, setTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const completedTodo = todos.filter(todo => todo.completed === true);
+  const activeTodo = todos.length - completedTodo.length;
```
DarkMistyRoom, line 32: apply `useMemo` here to avoid recalculation on each render

DarkMistyRoom, line 32:
```suggestion
  const completedTodo = todos.filter(todo => todo.completed === true);
  const activeTodosAmount = todos.length - completedTodo.length;
```

src/App.tsx
```diff
@@ -1,24 +1,244 @@
...
-const USER_ID = 0;
+const USER_ID = 11050;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sort, setSort] = useState(SortType.All);
+  const [errorMessage, setErrorMessage] = useState('');
```
DarkMistyRoom, line 27: better use enum for error messages

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,42 @@
...
+  updateTodo: (todoId: number, completed: boolean) => Promise<void>;
+  handleTitleEdit: (todoId: number, titles: string) => Promise<void>;
+};
+
+export const TodoList: React.FC<Props> = ({
+  todos,
+  onDelete,
+  tempTodo,
+  updateTodo,
+  handleTitleEdit,
```
DarkMistyRoom, line 18: You pass these props down without using them in this component. Better to use Context in this case. It is a suggestion for improvement, not a mistake. So if you're aware to do it, go forward. If not - it's ok

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,118 @@
...
+              placeholder="Empty todo will be deleted"
+              value={newTitle}
+              onBlur={handleFormSubmit}
+              onKeyUp={handleCancel}
+              onChange={event => setNewTitle(event.target.value)}
+            />
+          </form>
+        )
+        : (
+          <>
```
DarkMistyRoom, line 90:
```suggestion
        ) : (
          <>
```

DarkMistyRoom (changes requested):
It's a really good job! I request changes to make you consider some improvements
- fix the link

Screenshot 2023-05-22 at 10 10 54

- use autofocusing on input after double-clicking

- there is no loader after submitting editing. What if the request is rejected? It may be better to handle it like adding and deleting. If you have a different point of view, tell in the chat, or in the comment to the next request

src/App.tsx
loralevitska: not fixed

src/App.tsx
```diff
@@ -1,24 +1,273 @@
...
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sort, setSort] = useState(SortType.All);
+  const [errorMessage, setErrorMessage] = useState(ErrorType.None);
+  const [title, setTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const completedTodo = todos.filter(todo => todo.completed === true);
+  const activeTodosAmount = todos.length - completedTodo.length;
```
loralevitska, line 33: image

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,30 @@
...
+  }, [message]);
+
+  return (
+    <div className={classNames(
+      'notification is-danger is-light has-text-weight-normal', {
+        hidden: message.length === 0,
+      },
+    )}
+    >
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
loralevitska, line 21: Try to fix eslint rules and remove comments

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <button
+        type="button"
+        className="delete"
+        onClick={() => onDelete()}
```
loralevitska, line 25:
```suggestion
        onClick={onDelete}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,42 @@
...
+        />
+      ))}
+
+      {createTodo && (
+        <TodoInfo
+          todo={tempTodo}
+          tempTodoId={tempTodo.id}
+        />
+      )}
+    </section>
```
loralevitska, line 40: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
```
loralevitska, line 4: Remove comments. Try it write obvious code that explain all we need without comments)

loralevitska (approved):

PR #539, opened 2023-05-22
Develop

Stambuliak:
[DEMO LINK](https://Stambuliak.github.io/react_todo-app-with-api/)


jstmpelowycz (changes requested):
missing pieces of the functional: 
- `toggleAll` is not implemented
- cannot update a todo's title on double-click (on-blur sumbisson + deleting if empty title — as well)

PR #549, opened 2023-05-27
Develop

LBondarchuk:
[DEMO LINK](https://LBondarchuk.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -74,7 +74,6 @@ export const App: React.FC = () => {
 
   function editTodo(id: number, itemToEdit: object, changeTodo: string) {
     addTodoToProcesing(id);
-
     updateTodo(id, itemToEdit)
       .then(() => setTodoList(todoList.map(currentTodo => {
         if (changeTodo === 'All') {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,132 @@
-/* eslint-disable max-len */
+/* eslint-disable @typescript-eslint/no-unused-expressions */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import { Header } from './components/header/Header';
+import { List } from './components/list/List';
+import { Footer } from './components/footer/Footer';
 import { UserWarning } from './UserWarning';
+import { Warnings } from './components/warnings/Warnings';
+import { deleteTodo, getTodos, updateTodo } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoFilter } from './types/TodoFilter';
+import { ErrorMessage } from './types/ErrorMessage';
 
-const USER_ID = 0;
+const USER_ID = 10514;
+
+const { All, Active, Completed } = TodoFilter;
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[]>([]);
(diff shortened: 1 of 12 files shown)


LBondarchuk:
[DEMO LINK](https://LBondarchuk.github.io/react_todo-app-with-api/)

src/App.tsx
```diff
@@ -1,24 +1,131 @@
-/* eslint-disable max-len */
+/* eslint-disable @typescript-eslint/no-unused-expressions */
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
Anastasiia-Svintsova, line 2: Don't disable eslint. Try to solve errors

src/App.tsx
```diff
@@ -1,24 +1,131 @@
...
+  const loadTodos = () => {
+    getTodos(USER_ID)
+      .then((todos) => setTodoList(todos))
+      .catch(() => setError(ErrorMessage.Get));
+  };
+
+  useEffect(() => loadTodos(), []);
+
+  useEffect(() => {
+    setTimeout(() => setError(''), 3000);
```
Anastasiia-Svintsova, line 33: Clear the timeout on the component unmount to avoid side effects https://bobbyhadz.com/blog/react-cleartimeout

src/App.tsx
```diff
@@ -1,24 +1,131 @@
...
+  }, [error]);
+
+  const completedTodos = useMemo(() => todoList
+    .filter(todo => todo.completed).length, [todoList]);
+
+  const leftItems = todoList.length - completedTodos;
```
Anastasiia-Svintsova, line 39: Use useMemo here as well

src/App.tsx
```diff
@@ -1,24 +1,131 @@
...
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  function filterList() {
```
Anastasiia-Svintsova, line 45: It's better to use arrow functions. It makes React code more concise and easier to read

src/App.tsx
```diff
@@ -1,24 +1,131 @@
...
+          setError={setError}
+          addTodoToProcesing={(id:number | null) => addTodoToProcesing(id)}
+          editTodo={
+            (id: number, itemToEdit: object) => editTodo(id, itemToEdit, 'All')
+          }
+        />
+
+        <List
+          setTodoList={setTodoList}
+          todoList={filterList()}
```
Anastasiia-Svintsova, line 108: Don't call the function inside JSX. Move this function to a variable and use memorization for it as well

src/components/header/Header.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const todoForProcessing = {
+    title: inputValue,
+    completed: false,
+    userId: 10514,
+    id: 0,
+  };
+
+  useEffect(() => setAllActiveButton(
+    todoList.some(item => !item.completed),
+  ));
```
Anastasiia-Svintsova, line 41: Are you sure you need this useEffect without a dependency array here?

Anastasiia-Svintsova (changes requested):

src/components/header/Header.tsx
LBondarchuk: logically, it is not needed at all. But its absence gives an error

src/App.tsx
LBondarchuk: thanks

mykhalenych (approved):

PR #555, opened 2023-05-31
solution

ghostargg:
[DEMO LINK](https://ghostargg.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -37,10 +37,8 @@ export const App: React.FC = () => {
   }, []);
 
   useEffect(() => {
-    if (editedTodoId) {
-      if (inputRef.current) {
-        inputRef.current.focus();
-      }
+    if (editedTodoId && inputRef.current) {
+      inputRef.current.focus();
     }
   }, [editedTodoId]);
 
@@ -242,7 +240,6 @@ export const App: React.FC = () => {
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
-
       <div className="todoapp__content">
         <Header
           hasActiveTodos={hasActiveTodos}
@@ -276,9 +273,7 @@ export const App: React.FC = () => {
           onRemoveCompleted={handleRemoveCompleted}
         />
       </div>
-
-      {errorMessage
-      && (
+      {errorMessage && (
         <Notification
           onCleanErrorMessage={handleCleanErrorMessage}
           errorMessage={errorMessage}
--- a/src/components/Notification.tsx
+++ b/src/components/Notification.tsx
@@ -16,15 +16,12 @@ export const Notification: React.FC<NotificationProps> = ({
         hidden: !errorMessage,
       })}
     >
-
       <button
         type="button"
         className="delete"
         onClick={onCleanErrorMessage}
       />
-
       {errorMessage}
-
     </div>
   );
 };
--- a/src/components/Todo.tsx
+++ b/src/components/Todo.tsx
@@ -31,7 +31,7 @@ export const Todo: React.FC<TodoProps> = React.memo(({
 }) => {
   const { completed, id, title } = todo;
   const handleKeyUp = (event: React.KeyboardEvent<HTMLInputElement>) => {
-    if (event.key === 'Esc') {
+    if (event.key === 'Escape') {
       setEditedTodoId(null);
     }
   };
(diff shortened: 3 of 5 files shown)


src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
+      setErrorMessage('Failed to load todos');
+      setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    }
+  }, []);
+
+  useEffect(() => {
+    if (editedTodoId) {
+      if (inputRef.current) {
```
kasiakubisiak, line 41:
```suggestion
    if (editedTodoId && inputRef.current) {
```

src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
+
+    try {
+      setIsInputDisabled(true);
+      setTempTodo(newTodo);
+
+      const createdTodo = await addNewTodo(newTodo);
+
+      setTodos(prevTodos => [...prevTodos, createdTodo]);
+    } catch (error) {
+      setErrorMessage('Unable to add a todo');
```
kasiakubisiak, line 119: it would be better to set error in finally block

src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+  const [editedTodoId, setEditedTodoId] = useState<number | null>(null);
+  const [editedTodoText, setEditedTodoText] = useState('');
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const handleLoadTodos = useCallback(async () => {
+    try {
+      setTodos(await getTodos(USER_ID));
+    } catch (error) {
+      setErrorMessage('Failed to load todos');
```
kasiakubisiak, line 32: it would be better to set error in finally block

src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
+        try {
+          setLoadingTodoIds(prevIds => [...prevIds, editedTodo.id]);
+          const updatedTodo = { ...editedTodo, title: editedTodoText };
+
+          await updateTodo(editedTodo.id, updatedTodo);
+
+          setTodos(prevTodos => [...prevTodos
+            .map(todo => (todo.id === editedTodo.id ? updatedTodo : todo))]);
+        } catch (error) {
+          setErrorMessage('Unable to update a todo');
```
kasiakubisiak, line 230: same as in the above example, it would be better to set error in finally block

src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
+          filter={filter}
+          onFilterChange={handleFilterChange}
+          hasCompletedTodos={hasCompletedTodos}
+          todosLength={todos.length}
+          onRemoveCompleted={handleRemoveCompleted}
+        />
+      </div>
+
+      {errorMessage
+      && (
```
kasiakubisiak, line 281:
```suggestion
      {errorMessage && (
```

src/components/Notification.tsx
```diff
@@ -0,0 +1,30 @@
...
+  onCleanErrorMessage,
+  errorMessage,
+}) => {
+  return (
+    <div
+      className={cn('notification is-danger is-light has-text-weight-normal', {
+        hidden: !errorMessage,
+      })}
+    >
+
```
kasiakubisiak, line 19:
```suggestion
```

src/components/Todo.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+                </span>
+
+                <button
+                  type="button"
+                  className="todo__remove"
+                  onClick={() => onTodoRemove(id)}
+                >
+                  ×
+
```
kasiakubisiak, line 73:
```suggestion
```
lets clean code a bit and remove no needed line breaks

src/components/Todo.tsx
```diff
@@ -0,0 +1,127 @@
...
+                <div className="modal overlay">
+                  <div className="modal-background has-background-white-ter" />
+                  <div className="loader" />
+                </div>
+              </>
+            )}
+
+        </div>
+      )
+      : (
```
kasiakubisiak, line 103:
```suggestion
      ) : (
```

kasiakubisiak (changes requested):

src/App.tsx
ghostargg: Right now its implemented in a way that if errorMessage is empty string then if there is error i change it to adequate error message and if i move it to finally block i will have to create new state to hold information if there is error or not (set true of false in catch block) then i have to pass down two states as a props to children components. I can see it being worth doing for code clarity if working in a team but in this scenario ? If yes i will do it.

kasiakubisiak (approved):

PR #560, opened 2023-06-05
Solution

Irina-Kulish:
https://Irina-Kulish.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,207 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, {
+  useState,
+  useEffect,
+  useMemo,
+  useCallback,
+} from 'react';
+import { TodoList } from './component/TodoList/todoList';
+import { Footer } from './component/Footer/footer';
+import { Header } from './component/Header/header';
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,24 +1,207 @@
...
+  const [errorMessage, setErrorMessage] = React.useState('');
+  const [querySearch, setQuerySearch] = useState('');
+  const [filterBy, setFilterBy] = useState<GetFilter>(GetFilter.ALL);
+
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
+      switch (filterBy) {
+        case GetFilter.ACTIVE:
```
VitaliyBondarenko1982, line 31: filter each case separately. it will avoid redundant filter case GetFilterAll. in this or default case just return todos

VitaliyBondarenko1982 (changes requested):
- Screenshot from 2023-06-05 18-44-57
do not show this button when no todos at all
- when removing a few todos in one time should be loaders on each todo which in removing process
todo-with-api
- check counting left items. should always be 2 in this case
Screenshot from 2023-06-05 18-45-28
Screenshot from 2023-06-05 18-45-47
Screenshot from 2023-06-05 18-45-39

PR #577, opened 2023-06-08
Develop

natalia-klonowska:
- [DEMO LINK](https://natalia-klonowska.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -1,0 +1,103 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import { FormEvent, useState } from 'react';
+import { Todo } from '../types/Todo';
+import { addTodo, updateTodo } from '../api/todos';
+
+interface HeaderProps {
+  todos: Todo[],
+  userId: number,
+  handleTempTodo: (todo: Todo | null) => void,
+  handleError: (error: string) => void;
+  handleIsUpdating: (status: boolean) => void,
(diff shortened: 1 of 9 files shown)


kasiakubisiak:
there's this weird behaviour when you add todo - it dissapears and then appears again 

https://github.com/mate-academy/react_todo-app-with-api/assets/129735139/64760b5f-9f3c-4455-bb1f-641909e744e0

kasiakubisiak:
you should be able to cancel edition on escape and keyup, now escape doesnt cancel it and keyup saves changes

src/components/Header.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const handleCleaner = () => {
+    handleIsUpdating(false);
+    setIsDisabled(false);
+    setTodoTitle('');
+    handleTempTodo(null);
+  };
+
+  const handleUpdateAllCompleted = () => {
+    handleIsUpdating(true);
+    let isAllCompleted = true;
```
kasiakubisiak, line 32: maybe lets use useState to store and update this value

kasiakubisiak (changes requested):

src/components/Header.tsx
natalia-klonowska: this value is used only in this method so I won't use useState to store it

kasiakubisiak (approved):

PR #580, opened 2023-06-09
solution

Nikoramo:
[DEMO LINK](https://Nikoramo.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,0 +1,59 @@
+import React from 'react';
+import { Todo } from '../../types';
+import { TodoItem } from '../TodoItem';
+
+interface Props {
+  todos: Todo[];
+  tempTodo: Todo | null;
+  selectedTodoId: number | null;
+  completedTodosId: number[];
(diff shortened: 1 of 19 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,59 @@
...
+        todo={todo}
+        key={todo.id}
+        onDeleteTodo={onDeleteTodo}
+        isLoading={isLoading && (
+          !todo.id
+          || selectedTodoId === todo.id
+          || (completedTodosId?.includes(todo.id) && isClearCompleted)
+          || ((!completedTodosId?.includes(todo.id)
+              || completedTodosId.length === todos.length)
+            && isToggleAllTodos)
```
Volodymyr-Melnyk-Fron10der, line 42: move into variable

Volodymyr-Melnyk-Fron10der (changes requested):
1. filters block should be always centered 
image

2. there are a lot of functions in your App component, I would split it into some hooks, if you are able to do it)

PR #582, opened 2023-06-11
Solution

SuaBy514:
[DEMO LINK](https://suaby514.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,139 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
+import { deleteTodo, getTodos, updateTodo } from './api/todos';
+import { NewTodo } from './components/NewTodo/NewTodo';
+import { TodoList } from './components/TodoList/TodoList';
+import { Notification } from './components/Notification/Notification';
+import { Todo } from './types/Todo';
+import { FilterTypes } from './types/FilterTypes';
+import { Footer } from './components/Footer/Footer';
 
-const USER_ID = 0;
+const USER_ID = 10596;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [currentTodos, setCurrentTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [inputText, setInputText] = useState('');
+  const [filterType, setFilterType] = useState<FilterTypes>(FilterTypes.ALL);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [needUpdate, setNeedUpdate] = useState(false);
+  const [allComplete, setAllComplete] = useState(false);
+  const [allLoading, setAllLoading] = useState(false);
+
+  const handleErrorMessage = (type: string) => {
+    setErrorMessage(`Unable to ${type} a Todo`);
+  };
+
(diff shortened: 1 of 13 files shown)


kasiakubisiak:
when you add todo, its name should be cleared from input:

https://github.com/mate-academy/react_todo-app-with-api/assets/129735139/43a412cb-7dc0-4471-88a8-14d5995bf9d3

kasiakubisiak:
there's also '0' in the corner when no todos are completed, lets remove it

src/App.tsx
```diff
@@ -1,24 +1,158 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [inputText, setInputText] = useState('');
+  const [filterType, setFilterType] = useState<FilterTypes>(FilterTypes.ALL);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [needUpdate, setNeedUpdate] = useState(false);
+  const [allComplete, setAllComplete] = useState(false);
+  const [allLoading, setAllLoading] = useState(false);
+  const [batchOperation, setBatchOperation] = useState<BatchOperations | null>(
+    null,
+  ); // ugly line break, but linter wouldn't let the length go...
```
kasiakubisiak, line 27: its good, you can remove comment 😄

src/App.tsx
```diff
@@ -1,24 +1,158 @@
...
+      setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    }
+  }, [errorMessage]);
+
+  useEffect(() => {
+    switch (filterType) {
+      case FilterTypes.COMPLETED: {
+        setCurrentTodos(todos.filter((todo) => todo.completed));
```
kasiakubisiak, line 84: you use `todos.filter((todo) => todo.completed` few times, lets create function for that

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const handleDelete = () => {
+    setIsLoading(true);
+    onDeleteTodo(id);
+  };
+
+  const handleUpdateTitle = () => {
+    setIsEditingEnabled(false);
+
+    if (newTitle !== title) {
+      if (newTitle.trim().length > 0) {
```
kasiakubisiak, line 46: maybe you can do it this way?

```suggestion
    if (newTitle !== title && newTitle.trim().length > 0) {
```

kasiakubisiak (changes requested):

src/components/TodoItem/TodoItem.tsx
SuaBy514: I structured it that way because both conditions inside require the new title to be different from the old one; to remove if nesting, I would need to write a not-so-pretty-looking thing like this:
image

SuaBy514: The point I'm trying to make is that there's a repeated condition this way

SuaBy514: Though I suppose I could just check the newtitle against the title first and return if they are equal; will apply

kasiakubisiak: okay i get it, you can leave it how it was 👍

kasiakubisiak (changes requested):
last thing, try to fix failing tests

SuaBy514:
The test is failing due to missing animation library on github's side; a side effect of doing the optional instruction

SuaBy514:
The library in question is react-transition-group and its types

SuaBy514:
In order to fix this issue I would need to know how to add those to the list of things that get loaded with npm install, and we were never taught that; how should I go about doing that?

SuaBy514:
I will attempt committing the package.json

SuaBy514:
That seems to have worked

kasiakubisiak (approved):
good job!

SuaBy514:
Thank you!

PR #589, opened 2023-06-15
Add React Todo App with API

ghost:
[DEMO LINK](https://ruslanpashkov.github.io/react_todo-app-with-api/)

since [the working example](https://mate-academy.github.io/react_todo-app-with-api/) doesn't work correctly with some cases, i included them in my solution

https://github.com/mate-academy/react_todo-app-with-api/assets/129327901/3076f6e4-f0cd-4c87-b171-899300bbb0e0

### DIFF
--- a/src/components/Layout/Layout.tsx
+++ b/src/components/Layout/Layout.tsx
@@ -19,7 +19,7 @@ const USER_ID = 10684;
 
 export const Layout: FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [loadingTodos, setLoadingTodos] = useState<number[]>([0]);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([0]);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [error, setError] = useState<string | null>(null);
 
@@ -95,7 +95,7 @@ export const Layout: FC = () => {
 
   const removeTodo = useCallback(async (todoId: number) => {
     try {
-      setLoadingTodos((currentTodos) => [...currentTodos, todoId]);
+      setLoadingTodoIds((currentTodos) => [...currentTodos, todoId]);
 
       await deleteTodo(todoId);
 
@@ -105,7 +105,7 @@ export const Layout: FC = () => {
     } catch {
       setError('Unable to delete a todo');
     } finally {
-      setLoadingTodos((currentTodos) => (
+      setLoadingTodoIds((currentTodos) => (
         currentTodos.filter((id) => id !== todoId)
       ));
     }
@@ -122,7 +122,7 @@ export const Layout: FC = () => {
     const data = { ...values };
 
     try {
-      setLoadingTodos((currentTodos) => [...currentTodos, todoId]);
+      setLoadingTodoIds((currentTodos) => [...currentTodos, todoId]);
 
       if (data.title) {
         validateTitle(data.title);
@@ -140,7 +140,7 @@ export const Layout: FC = () => {
     } catch {
       setError('Unable to update a todo');
     } finally {
-      setLoadingTodos((currentTodos) => (
+      setLoadingTodoIds((currentTodos) => (
         currentTodos.filter((id) => id !== todoId)
       ));
     }
@@ -170,7 +170,7 @@ export const Layout: FC = () => {
           <>
             <TodoList
               todos={visibleTodos}
-              loadingTodos={loadingTodos}
+              loadingTodoIds={loadingTodoIds}
               tempTodo={tempTodo}
               removeTodo={removeTodo}
               editTodo={editTodo}
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -9,14 +9,14 @@ import { Todo } from '../../types/Todo';
 
 interface Props {
   todo: Todo;
-  loadingTodos: number[];
+  loadingTodoIds: number[];
   removeTodo: (todoId: number) => void;
   editTodo: (todoId: number, data: Partial<Todo>) => void;
 }
 
 export const TodoInfo: FC<Props> = ({
   todo,
-  loadingTodos,
+  loadingTodoIds,
   removeTodo,
   editTodo,
 }) => {
@@ -126,7 +126,7 @@ export const TodoInfo: FC<Props> = ({
 
       <div
         className={cn('modal overlay', {
-          'is-active': loadingTodos.includes(id),
+          'is-active': loadingTodoIds.includes(id),
         })}
       >
         <div className="modal-background has-background-white-ter" />
--- a/src/utils/helpers.ts
+++ b/src/utils/helpers.ts
@@ -15,23 +15,17 @@ export function getActiveTodos(todos: Todo[]) {
   return todos.filter((todo) => !todo.completed);
 }
 
-export function getVisibleTodos(todos: Todo[], filterStatus: TodoStatusFilter) {
+export function getVisibleTodos(todos: Todo[], statusFilter: TodoStatusFilter) {
   return todos.filter((todo) => {
-    let status: boolean;
-
-    switch (filterStatus) {
+    switch (statusFilter) {
       case TodoStatusFilter.Active:
-        status = !todo.completed;
-        break;
+        return !todo.completed;
 
       case TodoStatusFilter.Completed:
-        status = todo.completed;
-        break;
+        return todo.completed;
 
       default:
-        status = true;
+        return true;
     }
-
-    return status;
   });
 }
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -5,7 +5,7 @@ import { Todo } from '../../types/Todo';
 
 interface Props {
   todos: Todo[];
-  loadingTodos: number[];
+  loadingTodoIds: number[];
   tempTodo: Todo | null;
   removeTodo: (todoId: number) => void;
   editTodo: (todoId: number, data: Partial<Todo>) => void;
@@ -13,7 +13,7 @@ interface Props {
 
 export const TodoList: FC<Props> = ({
   todos,
-  loadingTodos,
+  loadingTodoIds,
   tempTodo,
   removeTodo,
   editTodo,
@@ -28,7 +28,7 @@ export const TodoList: FC<Props> = ({
         >
           <TodoInfo
             todo={todo}
-            loadingTodos={loadingTodos}
+            loadingTodoIds={loadingTodoIds}
             removeTodo={removeTodo}
             editTodo={editTodo}
           />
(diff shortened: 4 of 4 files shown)


silchyn (changes requested):
Todos can't be added, just like in the previous task

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
 
-const USER_ID = 0;
+const USER_ID = 10684;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterStatus, setFilterStatus] = useState(TodoFilterStatus.All);
```
silchyn, line 30:
```suggestion
  const [statusFilter, setStatusFilter] = useState(TodoFilterStatus.All);
```

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodos, setLoadingTodos] = useState<number[]>([0]);
```
silchyn, line 32: What is `0` here?

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+  const [error, setError] = useState<string | null>(null);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((todosFromServer) => {
+        setTodos(todosFromServer);
+      })
```
silchyn, line 39:
```suggestion
      .then(setTodos)
```

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+        setError(null);
+      }, 3000);
+    }
+
+    return () => clearTimeout(timerId);
+  }, [error]);
+
+  const completedTodos = useMemo(() => {
+    return getCompletedTodos(todos);
+  }, [todos, filterStatus]);
```
silchyn, line 59:
```suggestion
  const completedTodos = useMemo(() => (
    getCompletedTodos(todos);
  ), [todos, filterStatus]);
```

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+
+  const visibleTodos = useMemo(() => {
+    return getVisibleTodos(todos, filterStatus);
+  }, [todos, filterStatus]);
+
+  const activeTodosLeft = activeTodos.length;
+
+  const hasTodos = todos.length > 0;
+  const isTodosVisible = todos.length > 0 || Boolean(tempTodo);
```
silchyn, line 72:
```suggestion
  const hasVisibleTodos = todos.length > 0 || Boolean(tempTodo);
```

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+  const isAllTodosCompleted = !activeTodos.length;
+  const isClearCompletedVisible = completedTodos.length > 0;
```
silchyn, line 74:
```suggestion
  const canClearCompleted = completedTodos.length > 0;
```

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+  }, [createTodo]);
+
+  const removeTodo = useCallback(async (todoId: number) => {
+    try {
+      setLoadingTodos((currentTodos) => [...currentTodos, todoId]);
+      await deleteTodo(todoId);
+      setTodos((currentTodos) => (
+        currentTodos.filter((todo) => todo.id !== todoId)
+      ));
+    } catch {
```
silchyn, line 118:
```suggestion
    try {
      setLoadingTodos((currentTodos) => [...currentTodos, todoId]);
      
      await deleteTodo(todoId);
      
      setTodos((currentTodos) => (
        currentTodos.filter((todo) => todo.id !== todoId)
      ));
    } catch {
```

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+
+  const clearCompleted = useCallback(() => {
+    completedTodos.forEach(async (todo) => {
+      await removeTodo(todo.id);
+    });
+  }, [completedTodos, removeTodo]);
+
+  const editTodo = useCallback(async (
+    todoId: number,
+    todoUpdate: Partial<Todo>,
```
silchyn, line 135:
```suggestion
    values: Partial<Todo>,
```

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+      }
+
+      await updateTodo(todoId, data);
+      setTodos((currentTodos) => currentTodos.map((currentTodo) => {
+        if (currentTodo.id === todoId) {
+          return { ...currentTodo, ...data };
+        }
+
+        return currentTodo;
+      }));
```
silchyn, line 155:
```suggestion
      setTodos((currentTodos) => currentTodos.map((currentTodo) => (
        currentTodo.id === todoId
          ? { ...currentTodo, ...data }
          : currentTodo
      )));
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,47 @@
...
+interface Props {
+  filterStatus: TodoFilterStatus;
+  changeFilterStatus: (status: TodoFilterStatus) => void;
+}
+
+export const TodoFilter: React.FC<Props> = ({
+  filterStatus,
+  changeFilterStatus,
+}) => {
+  return (
```
silchyn, line 14:
```suggestion
}) => (
```

silchyn (changes requested):

src/App.tsx
ruslanpashkov: @silchyn it's the id of the temporarily selected todo, if we are waiting for a response from the server, this todo will hang as loading at first until the server responds (because we don't have the id instantly)

ruslanpashkov: @silchyn the message above only makes sense to add a todo, we need a temporary todo to show the effect of loading it while we are waiting for a response from the server, in case there is no todo on the server, it does not cause any problems, because tempTodo is only created when you are interactive with the user|

ruslanpashkov: in other words, we set a temporary todo id 0 (which will always be with the loading effect), until we assign our todo id from the server

IrynaBalandiukh (approved):
Great job! 🚀🚀🚀

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,30 @@
...
+}) => {
+  const handleCloseButtonOnClick = () => closeError();
+
+  return (
+    <div
+      className={cn('notification is-danger is-light has-text-weight-normal', {
+        hidden: !error,
+      })}
+    >
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
dffUqp, line 21: let's fix this eslint issue:smiley:
try to give `aria-label` to button, i think it can help

src/App.tsx
```diff
@@ -1,24 +1,211 @@
...
+      setLoadingTodos((currentTodos) => (
+        currentTodos.filter((id) => id !== todoId)
+      ));
+    }
+  }, [deleteTodo]);
+
+  const clearCompleted = useCallback(() => {
+    completedTodos.forEach(async (todo) => {
+      await removeTodo(todo.id);
+    });
```
dffUqp, line 128: I think in this case we can use `Promise.all`

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const { id, title, completed } = todo;
+
+  const [isEditing, setIsEditing] = useState(false);
+  const todoTitleRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    todoTitleRef.current?.focus();
+  }, [isEditing]);
+
+  const submitTodo = useCallback(() => {
```
dffUqp, line 32: unnecessary `useCallback`

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,137 @@
...
+    submitTodo();
+  };
+
+  const handleTodoOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (todoTitleRef.current) {
+      todoTitleRef.current.value = event.target.value;
+    }
+  };
+
+  const handleTodoOnBlur = () => submitTodo();
```
dffUqp, line 66: do we really need this function?

dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

src/components/Layout/Layout.tsx
```diff
@@ -0,0 +1,194 @@
...
+} from '../../api/todos';
+import {
+  getActiveTodos, getCompletedTodos, getVisibleTodos, validateTitle,
+} from '../../utils/helpers';
+
+const USER_ID = 10684;
+
+export const Layout: FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loadingTodos, setLoadingTodos] = useState<number[]>([0]);
```
silchyn, line 22:
```suggestion
  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([0]);
```

silchyn, line 22: The problem with `0` here is that it's a wholly hardcoded value and shouldn't be used like that because whoever will read this code would need context to understand what's happening here. Think about a workaround

src/components/Layout/Layout.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  const visibleTodos = useMemo(() => (
+    getVisibleTodos(todos, status as TodoStatusFilter)
+  ), [todos, status]);
+
+  const activeTodosLeft = activeTodos.length;
+
+  const hasTodos = todos.length > 0;
+  const hasVisibleTodos = todos.length > 0 || Boolean(tempTodo);
+  const isAllTodosCompleted = !activeTodos.length;
```
silchyn, line 62:
```suggestion
  const hasActiveTodos = !activeTodos.length;
```

src/utils/helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+        status = todo.completed;
+        break;
+
+      default:
+        status = true;
+    }
+
+    return status;
+  });
+}
```
silchyn, line 37:
```suggestion
export function getVisibleTodos(todos: Todo[], statusFilter: TodoStatusFilter) {
  return todos.filter((todo) => {
    switch (filterStatus) {
      case TodoStatusFilter.Active:
        return !todo.completed;

      case TodoStatusFilter.Completed:
        return todo.completed;

      default:
        return true;
    }
  });
}
```

silchyn (approved):
🎉🎉🎉

src/components/Layout/Layout.tsx
ruslanpashkov: @silchyn it's a great solution if our statement will be activeTodos.length > 0, but we can't even use `hasCompletedTodos`, because it means that we have at least one completed todo, and we need to make sure that all are completed

PR #590, opened 2023-06-16
React todo app with api was done

SamirMamedov1010:
[DEMO LINK](https://SamirMamedov1010.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,24 +1,244 @@
...
+            ...todo,
+            completed: !todo.completed,
+          });
+        }
+
+        return todo;
+      });
+    };
+
+    const updateCompleted = () => {
```
sTorba24, line 171:
```suggestion
    const handleUpdateCompleted = () => {
```

sTorba24 (changes requested):
Almost done, lets fix some bugs

1. When i change todo status, it should just change icon view, now it's removing from `all` list
2. When i change title to '' it disappears, shouldn't work so

src/App.tsx
SamirMamedov1010: Thank you  for improves, changed naming convention in this case. In addition to this i fix problem with todo status change, function filtered todos was with error(in case all - showed not completed todos) Finally, problem in changing title to '' - is not a problem, cause the task said to do this in this way, even standart placeholder say - 'Empty todo will be deleted'.

loralevitska (approved):
Good job!
Check my comments for improvement

Add right condition for 1 item left
image

src/App.tsx
```diff
@@ -1,24 +1,238 @@
...
+import { UserWarning } from './components/UserWarning';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [errorType, setErrorType] = useState('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [selectedFilter, setSelectedFilter] = useState(FilterValues.ALL);
+  const [removingTodoIds, setRemovingTodoIds] = useState<number[]>([]);
+  const [updatingTodoIds, setUpdatingTodoIds] = useState<number[]>([]);
```
DarkMistyRoom, line 21: Do we need to separate these states? Consider using one state for both cases

DarkMistyRoom (approved):

PR #595, opened 2023-06-26
react_todo-app-with-api

YelyzavetaMolochynska:
[DEMO LINK](https://YelyzavetaMolochynska.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,304 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import {
+  getTodos,
+  postTodo,
+  removeTodo,
+  updateTodo,
+} from './api/todos';
+import { FilterType } from './types/FilterType';
 
-const USER_ID = 0;
+const USER_ID = 10824;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState('');
+  const [errorType, setErrorType] = useState('');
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,24 +1,304 @@
...
+
+  const visibleTodos = getVisibleTodo(filterType, todos);
+
+  const hasCompleted = visibleTodos.some(todo => todo.completed);
+
+  (function handleTempTodo() {
+    if (tempTodo) {
+      visibleTodos.splice(indexUpdatedTodo, 1, tempTodo);
+    }
+  }());
```
VitaliyBondarenko1982, line 74: looks like redundant function. splice mutate visibleTodos array.
anyway it is better to make in useEffect.
just render tempTodo after visible todos
```suggestion
```
it can be a reason of this bug (blinking)
chrome-capture-2023-5-26 (1)

src/App.tsx
```diff
@@ -1,24 +1,304 @@
...
+const USER_ID = 10824;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState('');
+  const [errorType, setErrorType] = useState('');
+  const [filterType, setFilterType] = useState(FilterType.ALL);
+  const [loading, setLoading] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [indexUpdatedTodo, setIndexUpdatedTodo] = useState(0);
```
VitaliyBondarenko1982, line 26:
```suggestion
  const [indexUpdatedTodo, setIndexUpdatedTodo] = useState(0);
```
for what do you need this state

VitaliyBondarenko1982 (changes requested):
toogle button in header shouldn't be visible if there is no any todos
Screenshot from 2023-06-26 12-45-15

src/App.tsx
YelyzavetaMolochynska: to insert tempTodo render in current place in the list when you delete it

IvanFesenko (changes requested):
You shouldn't update stae in case when request is down
https://github.com/mate-academy/react_todo-app-with-api/assets/61904995/25710702-9296-4185-bc82-20886c421232

PR #604, opened 2023-07-04
Solution

bohdan-mykhailenko:
[DEMO LINK](https://bohdan-mykhailenko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -17,7 +17,7 @@ import './styles/todoapp.scss';
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
-  const [statusFilter, setFilterStatus] = useState(StatusFilter.all);
+  const [statusFilter, setFilterStatus] = useState(StatusFilter.ALL);
   const [error, setError] = useState<string | null>(null);
   const [isLoading, setIsLoading] = useState(false);
   const [title, setTitle] = useState('');
@@ -26,7 +26,9 @@ export const App: React.FC = () => {
   useEffect(() => {
     getTodos()
       .then(setTodos)
-      .catch(errorFromServer => setError(errorFromServer.message));
+      .catch(errorFromServer => {
+        setError(`Unable to load todos: ${errorFromServer.message}`);
+      });
   }, []);
 
   useEffect(() => {
@@ -56,8 +58,8 @@ export const App: React.FC = () => {
     todos.filter((todo) => !todo.completed)
   ), [todos, statusFilter]);
 
-  const hasSomeActiveTodos = activeTodos.length > 0;
   const hasSomeCompletedTodos = completedTodos.length > 0;
+  const hasOnlyCompletedTodos = completedTodos.length === todosCount;
 
   const changeFilterStatus = useCallback((filterStaus: StatusFilter) => {
     setFilterStatus(filterStaus);
@@ -94,8 +96,8 @@ export const App: React.FC = () => {
           todoFromResponse,
         ]);
       })
-      .catch(() => {
-        setError('Unable to add a todo');
+      .catch((errorFromServer) => {
+        setError(`Unable to add a todo: ${errorFromServer.message}`);
       })
       .finally(() => {
         setIsLoading(false);
@@ -116,7 +118,9 @@ export const App: React.FC = () => {
           currentTodos.filter((todo) => todo.id !== todoId)
         ));
       })
-      .catch(() => setError('Unable to delete a todo'))
+      .catch((errorFromServer) => {
+        setError(`Unable to delete a todo: ${errorFromServer.message}`);
+      })
       .finally(() => {
         setLoadingTodoIds((currentIds) => (
           currentIds.filter((id) => id !== todoId)
@@ -139,20 +143,13 @@ export const App: React.FC = () => {
       todoId,
     ]);
 
-    const updatedTodo: Partial<Todo> = {};
+    const updatedField = typeof updatedValue === 'boolean'
+      ? 'completed'
+      : 'title';
 
-    switch (typeof updatedValue) {
-      case 'string':
-        updatedTodo.title = updatedValue;
-        break;
-
-      case 'boolean':
-        updatedTodo.completed = updatedValue;
-        break;
-
-      default:
-        break;
-    }
+    const updatedTodo: Partial<Todo> = {
+      [updatedField]: updatedValue,
+    };
 
     patchTodo(todoId, updatedTodo)
       .then((todoFromServer) => {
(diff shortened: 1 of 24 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <button
+          type="button"
+          className={cn(
+            'todoapp__toggle-all', {
+              active: isSomeActiveTodos,
+            },
+          )}
+          onClick={handleToggleTodos}
+        />
+      )}
```
serentinos, line 56: you can use aria-label in button to avoid this eslint warning

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,33 @@
...
+      )}
+    >
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+      <button
+        type="button"
+        className="delete"
+        onClick={onCloseNotifications}
+      />
+      {error}
+    </div>
```
serentinos, line 31: same here

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,44 @@
...
+          <a
+            href={status !== 'all' ? `#/${status}` : '#/'}
+            className={cn('filter__link', {
+              selected: status === filterStatus,
+            })}
+            onClick={event => handleFilterClick(event, FilterStatus[status])}
+          >
+            {status}
+          </a>
+        </Fragment>
```
serentinos, line 40: :fire:

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const TodoList: React.FC<TodoListProps> = memo(({
+  tempTodo,
+  todos,
+  onRemoveTodo,
+  onUpdateTodo,
+  loadingTodoIds,
+}) => {
+  const loadedTodos = tempTodo
+    ? [...todos, tempTodo]
+    : todos;
```
serentinos, line 22: Cool solution, because I created an another TodoItem which shows only when tempTodo is present, nice xD

serentinos (approved):
Good job!

silchyn (changes requested):
image

bohdan-mykhailenko:
> image

What is the problem here? Because of the error when trying to update? But as I see it's  not the problem with the code because no one can't update

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,33 @@
...
+        'notification',
+        'is-danger',
+        'is-light',
+        'has-text-weight-normal', {
+          hidden: !error,
+        },
+      )}
+    >
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+      <button
```
roman-mirzoian, line 25: This line has been disabled for ESLint check 'jsx-a11y/control-has-associated-label'. It's better to add an accessible label to the button element for better accessibility.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,70 @@
...
+      return;
+    }
+
+    onAddTodo(title);
+  };
+
+  return (
+    <header className="todoapp__header">
+      {true && (
+        // eslint-disable-next-line jsx-a11y/control-has-associated-label
```
roman-mirzoian, line 46: The `eslint-disable-next-line jsx-a11y/control-has-associated-label` is used to bypass accessibility requirements for controls. Consider providing an accessible label for the button.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+      return;
+    }
+
+    onAddTodo(title);
+  };
+
+  return (
+    <header className="todoapp__header">
+      {true && (
```
roman-mirzoian, line 45: The boolean expression in the JSX always evaluates to `true`, which means the `button` will always be rendered. If this is the intended behaviour, remove the redundant conditional. If you want to conditionally render the `button`, replace `true` with the appropriate condition.

src/types/StatusFilter.ts
```diff
@@ -0,0 +1,5 @@
+export enum StatusFilter {
+  all = 'all',
```
roman-mirzoian, line 2: This is not a mistake, but it is customary to write enam names in capital letters ( `ALL = 'all'` ).

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,159 @@
...
+  };
+
+  const handleDoubleClick = () => {
+    const updatedStatus = !isActiveEditForm;
+
+    setIsActiveEditForm(updatedStatus);
+  };
+
+  const handleKeyTouch = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.code === 'Escape') {
```
roman-mirzoian, line 80: It would be very cool to put such strings into separate variables to avoid using magic data.
For example:
```
const ESCAPE= 'Escape';
const NOT_SET_VALUE = '-';
```
This applies to all code.

src/App.tsx
```diff
@@ -1,24 +1,225 @@
...
+
+    patchTodo(todoId, updatedTodo)
+      .then((todoFromServer) => {
+        const updatedTodos = todos.map((todo) => (
+          todo.id === todoFromServer.id ? todoFromServer : todo
+        ));
+
+        setTodos(updatedTodos);
+      })
+      .catch(() => {
```
roman-mirzoian, line 165: In the `catch` block, you should also return an `error` from the server and show it to the user or output it to the console in case of unexpected errors:
```
.catch((error) => {
    setError(`Unable to update a todo: ${error.message}`);
})
```

src/App.tsx
```diff
@@ -1,24 +1,225 @@
...
+      todoId,
+    ]);
+
+    deleteTodo(todoId)
+      .then(() => {
+        setTodos((currentTodos) => (
+          currentTodos.filter((todo) => todo.id !== todoId)
+        ));
+      })
+      .catch(() => setError('Unable to delete a todo'))
```
roman-mirzoian, line 119: In the `catch` block, you should also return an `error` from the server and show it to the user or output it to the console in case of unexpected errors:
```
.catch((error) => {
    setError(`Unable to delete a todo: ${error.message}`);
})
```

src/App.tsx
```diff
@@ -1,24 +1,225 @@
...
+    switch (typeof updatedValue) {
+      case 'string':
+        updatedTodo.title = updatedValue;
+        break;
+
+      case 'boolean':
+        updatedTodo.completed = updatedValue;
+        break;
+
+      default:
```
roman-mirzoian, line 153: If you need to check only two conditions and not return the default value, you can use a short construction:
```
const updatedFieldName = typeof updatedValue === 'boolean' ? 'completed' : 'title';
updatedTodo[updatedFieldName] = updatedValue;
```

roman-mirzoian (changes requested):
The code is written well, but a few points need to be improved a little. 
Pay attention to the use of async/await, if the chain of promises takes up a lot of space, this use will make your code more readable. 
Try to break large chunks of code into smaller chunks of code.
Also, now the update created todo does not work and the application does not give any information about why this happened, so it is worth returning more detailed information about the error:
https://gyazo.com/60ca1b911b4aed9593f89997fc6ad8d7

dffUqp (approved):
🚀

PR #608, opened 2023-07-04
Develop

serentinos:
[DEMO LINK](https://serentinos.github.io/react_todo-app-with-api/)


ivankhrsh:
If title is not changed but I press Enter - there is network request. You should check if tittle changed before doing request

src/components/EditForm.tsx
```diff
@@ -0,0 +1,73 @@
...
+    if (textInput.current) {
+      textInput.current.focus();
+    }
+
+    window.addEventListener('keyup', escKeyUpHandler);
+
+    return () => {
+      window.removeEventListener('keyup', escKeyUpHandler);
+    };
+  }, [escKeyUpHandler]);
```
ivankhrsh, line 56: you can use onKeyDown React event and checking if pressed button is Escape instead of it

src/components/Header.tsx
```diff
@@ -0,0 +1,71 @@
...
+        const { id, completed } = currentTodo;
+
+        return completed && !isAllTodosCompleted
+          ? editTodoByID(id, { completed })
+          : editTodoByID(id, { completed: !completed });
+      }),
+    );
+
+    setCurrentlyLoadingTodos([]);
+  };
```
ivankhrsh, line 50: is it good idea to have this logic in header instead of app?

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,101 @@
...
+              className="todo__remove"
+              onClick={onHandleRemoveTodo}
+            >
+              ×
+            </button>
+          </>
+        )}
+
+      <div className={cn('modal overlay', {
+        'is-active': isLoading || isLoadingNow,
```
ivankhrsh, line 93: can't get sense of this variables from names, they are look the same
if it's two different actions - better to have more meaningful names

src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,33 @@
...
+      {Object.entries(FilterType).map(([key, value]) => (
+        <a
+          key={key}
+          href={`#/${value}`}
+          className={cn('filter__link', {
+            selected: value === currentFilterType,
+          })}
+          onClick={() => handleSelectFilterType(value)}
+        >
+          {value[0].toUpperCase() + value.slice(1)}
```
ivankhrsh, line 28: You can just change values in src/Enums/FilterType.ts to have first capitalized letter and avoid this manipulation.

ivankhrsh (changes requested):

src/components/TodosFilter.tsx
serentinos: Yep I agree xD

src/components/Header.tsx
serentinos: I guess yes, because in app i have just basic manipulation with API

src/components/EditForm.tsx
serentinos: Thank you!

serentinos:
@ivankhrsh Thank you for these suggestions! They helped me a lot! :)

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,43 @@
...
+    }));
+  }, [setError]);
+
+  const debouncedDisableError = useMemo(() => (
+    debounce(disableError, 3000)
+  ), [disableError]);
+
+  if (loadError.status) {
+    debouncedDisableError();
+  }
```
33Orange, line 25: You can realize it by `useEffect`.
Debounce intended for slightly different tasks. 
Create `useEffect`  which will create timeOut when an error update and will unmount this timer after 3 seconds

src/components/Footer.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+      <TodosFilter
+        setFilterType={setFilterType}
+        currentFilterType={filterType}
+      />
+
+      <button
+        type="button"
+        className={cn('todoapp__clear-completed', {
+          hidden: completedTodosIds.length === 0,
```
33Orange, line 54: Create a variable for this condition

33Orange (changes requested):
Almost done

A few moments to improve and fix your code

src/components/NewTodoForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+      return;
+    }
+
+    setIsInputDisabled(true);
+    const isSucsess = await addNewTodo(newTodoQuery);
+
+    if (isSucsess) {
+      setNewTodoQuery('');
+    }
```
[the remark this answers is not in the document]
serentinos: addNewTodo(newTodoQuery) function returns a boolean that indicates that the todo item was successfully added. If it returns false, it suggests that an error occurred during the process. Is this a good practice or not?

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  const onHandleRemoveTodo = useCallback(async () => {
+    setIsLoading(true);
+    await removeTodoByID(id);
+    setIsLoading(false);
+  }, [id, removeTodoByID]);
+
+  const editCurrentTodo = useCallback(async (data: Partial<Todo>) => {
+    setIsLoading(true);
+    await editTodoByID(id, data);
```
roman-mirzoian, line 31: Any asynchronous requests should be wrapped in a `try-catch` block to handle unexpected errors and prevent application crashes.

src/components/ErrorNotification.tsx
roman-mirzoian: I support this remark, it is important to make similar parts using `useEffect`.

src/components/EditForm.tsx
```diff
@@ -0,0 +1,79 @@
...
+    event: KeyboardEvent<HTMLInputElement>,
+  ) => {
+    if (event.key === 'Escape') {
+      cancelChange();
+    }
+  };
+
+  const onFormSubmit = () => {
+    if (query !== title) {
+      editFormSubmitHandler(query.trim());
```
roman-mirzoian, line 47: It is good that there is a reduction of extra spaces here (`.trim()`), but it is better to transfer it to a separate variable and use it also in the `if` condition.

roman-mirzoian (changes requested):
A job well done and very few fixes left.

PR #609, opened 2023-07-04
solution

gabriella135:
[DEMO LINK](https://gabriella135.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,194 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useMemo, useState } from 'react';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { Todo } from './types/Todo';
+import { Filter } from './enums/filter';
+import { ErrorMessage } from './enums/error';
+import { Error } from './components/Error';
+import { TodoList } from './components/TodoList';
+import {
+  getTodos, addTodos, deleteTodo, updateTodos,
+} from './api/todos';
 
-const USER_ID = 0;
+const USER_ID = 10567;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterValue, setFilterValue] = useState(Filter.All);
+  const [errorMessage, setErrorMessage]
+  = useState<ErrorMessage>(ErrorMessage.NONE);
+  const [value, setValue] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletedTodosId, setDeletedTodosId] = useState<number[] | []>([]);
+
+  const fetchData = async () => {
+    try {
+      const response = await getTodos(USER_ID);
+      const todosData = response as Todo[];
+
+      setTodos(todosData);
+    } catch (error) {
+      setErrorMessage(ErrorMessage.LOAD);
+    }
+  };
+
+  useEffect(() => {
+    fetchData();
+  }, []);
+
+  useEffect(() => {
+    setTimeout(() => (
+      setErrorMessage(ErrorMessage.NONE)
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,24 +1,194 @@
...
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { Todo } from './types/Todo';
+import { Filter } from './enums/filter';
+import { ErrorMessage } from './enums/error';
+import { Error } from './components/Error';
+import { TodoList } from './components/TodoList';
+import {
+  getTodos, addTodos, deleteTodo, updateTodos,
+} from './api/todos';
```
loralevitska, line 13:
```suggestion
import {
  getTodos, 
  addTodos,
   deleteTodo, 
   updateTodos,
} from './api/todos';
```

src/components/TodoInfo.tsx
```diff
@@ -0,0 +1,116 @@
...
+          <form onSubmit={handleTitleEdit}>
+            <input
+              type="text"
+              className="todo__title-field"
+              value={newTitle}
+              onChange={(event) => setNewTitle(event.target.value)}
+              onBlur={(event) => handleTitleEdit(event)}
+              onKeyUp={(event) => (
+                event.key === 'Escape' && handleTitleEdit(event)
+              )}
```
loralevitska, line 82: Move to separate function  (transfer the function to a variable, and here only write the name of the function)

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
```
loralevitska, line 4: Remove comments everywhere. Try it write obvious code that explain all we need without comments)

loralevitska (changes requested):

- remove outlines everywhere, it breaks design
- this buttons don't work, please fix it
image 
image

image
- don't show `Clear completed` button if you don't have a completed todo
- fix error message 
image
- only one field can be in focus at a time
image
- I can add empty `todo`. Fix it with [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)
image
- add lines between todos according [working version](https://mate-academy.github.io/react_todo-app-with-api/)
image

lerastarynets (changes requested):
Pls fix all the prev comments. Toggle all button is still not working, items can be toggled on select, user can add a todo consisting of whitespaces only and there are no lines between todos in list

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+  };
+
+  useEffect(() => {
+    fetchData();
+  }, []);
+
+  useEffect(() => {
+    setTimeout(() => (
+      setErrorMessage(ErrorMessage.NONE)
+    ), 3000);
```
Anastasiia-Svintsova, line 59: It's a good practice to clear all timers on the component unmount to avoid unpredictable side effects https://bobbyhadz.com/blog/react-cleartimeout

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+  }, [errorMessage]);
+
+  const handleDeleteTodo = (id: number) => {
+    setDeletedTodosId([id]);
+
+    deleteTodo(id)
+      .then(() => {
```
Anastasiia-Svintsova, line 66: I noticed that you use 2 different approaches in the project `async/await` and `then` for working with async functions. It's  generally recommended to be consistent and use a single approach throughout a project

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+      .catch(() => setErrorMessage(ErrorMessage.DELETE))
+      .finally(() => setDeletedTodosId([]));
+  };
+
+  const handleUpdateTodo = (id: number) => {
+    setDeletedTodosId([id]);
+
+    const updateTodo = todos.find(todo => todo.id === id);
+
+    if (updateTodo) {
```
Anastasiia-Svintsova, line 80: Just return the function if there is no updateTodo to avoid nesting

Anastasiia-Svintsova (approved):
The App component is huge in my opinion, so just general recommendation is to separate the logic from the component itself to improve readability and maintainability.

PR #610, opened 2023-07-04
Solution v1.0

skinnyjazz:
[DEMO LINK](https://skinnyjazz.github.io/react_todo-app-with-api/)


amashutikov (changes requested):
Looks good. But your App is not deleting empty todo on edit submit. And also there shold be placeholder.
image
Also i cant submit the same title for todo.

Your filters is jumping to the right when there is no completed todos. You can fix that adding conditional styles to the "Clear completed" button.
image

Escape button is not working too, to exit editing without saving changes.

Toggle all button should be active only when all todos are completed.

PR #611, opened 2023-07-04
completed toggle button

EugeneSherepa:
[DEMO LINK](https://EugeneSherepa.github.io/react_todo-app-with-api/)
Task is not completed yet

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,182 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useEffect,
+  useCallback,
+  useState,
+  FormEvent,
+} from 'react';
+import {
+  getTodos,
+  addTodos,
+  deleteTodo,
+  updateTodo,
+} from './api/todos';
 import { UserWarning } from './UserWarning';
+import { Header } from './components/TodoHeader';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/todoFooter';
+import { Message } from './components/ErrorMessege';
+import { Todo } from './types/Todo';
+import { client } from './utils/fetchClient';
+import { TodoStatus } from './types/TodoStatus';
+import { FilterTodos } from './utils/todoFilter';
 
-const USER_ID = 0;
+export const USER_ID = 10883;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todoFilter, setTodoFilter] = useState<TodoStatus>(TodoStatus.All);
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        setVisibleError('Unable to load todos');
+      });
+  }, []);
+
+  const handleFormSubmit = useCallback((e: FormEvent<HTMLFormElement>) => {
```
cephy77, line 42: Maybe it is better to name e as event

src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+  };
+
+  const removeCompletedTodos = useCallback(() => {
+    const completedTodoIds = todos
+      .filter(todo => todo.completed)
+      .map(todo => todo.id);
+
+    Promise.all(
+      completedTodoIds.map(id => (
+        client.delete(`/todos/${id}`)
```
cephy77, line 129: It is better to use separate function in api/todos.ts

src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+      <h1 className="todoapp__title">todos</h1>
+
+      <Header
+        todos={todos}
+        handleFormSubmit={handleFormSubmit}
+        todoTitle={todoTitle}
+        setTodoTitle={setTodoTitle}
+        isTodoLoading={isTodoLoading}
+        handleToggleButton={handleToggleButton}
+      />
```
cephy77, line 159: In my opinion, you can pass only todos and two functions, and then write all the logic inside the component

src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+
+      <div className="todoapp__content">
+
+        <TodoList
+          tempTodo={tempTodo}
+          todos={FilterTodos(todos, todoFilter)}
```
cephy77, line 165: It seems that the mentor will ask you to create a separate variable (visibleTodos) for that

src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+          removeTodo={removeTodo}
+          deletedTodoId={deletedTodoId}
+          handleCheck={handleCheck}
+          todoTitle={todoTitle}
+          setTodos={setTodos}
+          setTodoTitle={setTodoTitle}
+        />
+
+        {todos.length > 0 && (
```
cephy77, line 174: Or Boolean(todos.lenght)

cephy77 (changes requested):

silchyn (changes requested):
The review shouldn't be requested if the task is not completed yet

PR #612, opened 2023-07-04
solution

MaksimKad:
[DEMO LINK](https://MaksimKad.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,215 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { FormEvent, useEffect, useState } from 'react';
+import { Todo } from './components/Todo';
+import { Todo as TodoType } from './types/Todo';
+import { Notification } from './components/Notification';
+import { Filter } from './components/Filter';
+import {
+  getTodos,
+  addTodo,
+  removeTodo,
+  updateTodo,
+} from './api/todos';
 
-const USER_ID = 0;
+const USER_ID = 10915;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<TodoType[]>([]);
+  const [error, setError] = useState('');
+  const [filteredTodos, setFilteredTodos] = useState<TodoType[]>([]);
+  const [filter, setFilter] = useState('');
+  const [addValue, setAddValue] = useState('');
+  const [isInputDisable, setIsInputDisable] = useState(false);
+  const [tempTodo, setTempTodo] = useState<TodoType | null>(null);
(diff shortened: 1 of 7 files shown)


src/App.tsx
niaYaro, line 1: all todos being created/edited/deleted should be covered with loader (use className 'is-hidden')

230707-sr-1

niaYaro: Filter 'All' should be default selected.

230707-sr-2

niaYaro: Looks like editing logic does not match the one implemented in [the working example](https://mate-academy.github.io/react_todo-app-with-api/)

Try to use autoFocus attribute for input or do it with [useRef](https://www.youtube.com/watch?v=42BkpGe8oxg)

230707-sr-3

niaYaro: Items left number should increase if todo is added

230707-sr-4

niaYaro: `toggleAll` button should have `active` class only if all the todos are completed

230707-sr-6

Also pay attention that `toggleAll` button disappears if there are no completed todos. Such behavior does not match the one implemented in the working example

230707-sr-5

niaYaro (approved):
Good enough! Few points to pay your attention.

src/App.tsx
```diff
@@ -1,24 +1,215 @@
...
+      .catch(() => setError('Unable to Load todos'));
+  }, [deleteIds, isUpdating]);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(response => {
+        setFilteredTodos(response);
+      })
+      .catch(() => setError('Unable to Load todos'));
+  }, [todos, tempTodo]);
```
AlexKherson, line 41: I am not sure Why do you use useEffect for this case You should get filteredTodos by filtering todos and for optimizing you can use useMemo.

src/App.tsx
```diff
@@ -1,24 +1,215 @@
...
+      case 'Completed': {
+        setFilteredTodos(todos.filter(todo => todo.completed === true));
+        break;
+      }
+
+      default: {
+        setFilteredTodos(todos);
+      }
+    }
+  }, [filter]);
```
AlexKherson, line 65: the same Why do use useEffect in this case??? we should useEffect for loading data from server If want to optimize You should useMemo.

src/App.tsx
```diff
@@ -1,24 +1,215 @@
...
+      .then(response => {
+        setTodos(response);
+      })
+      .catch(() => setError('Unable to Load todos'));
+  }, [deleteIds, isUpdating]);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(response => {
+        setFilteredTodos(response);
```
AlexKherson, line 38: in this line, you set to setFilteredTodos all todos, are you??

AlexKherson (changes requested):
Good job. A few moments to change.

PR #617, opened 2023-07-05
made react task todo-app-with-api

vladyslav-lapshyn:
[DEMO LINK](https://vladyslav-lapshyn.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,220 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import {
+  FC,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import { UserWarning } from './UserWarning';
+import { Todo, UpdateTodoArgs } from './types/Todo';
+import { TodoList } from './Components/TodoList';
+import {
+  createTodo,
+  getTodos,
+  deleteTodo,
+  updateTodo,
+} from './api/api.todos';
+import { FilterOption } from './enums/FilterOption';
+import { Footer } from './Components/Footer';
+import { TodoNotification } from './Components/TodoNotification';
+import { Header } from './Components/Header';
 
-const USER_ID = 0;
+const USER_ID = 10898;
 
-export const App: React.FC = () => {
+export const App: FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterOption, setFilterOption] = useState(FilterOption.All);
+  const [error, setError] = useState<string | null>(null);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingItems, setLoadingItems] = useState<number[]>([]);
+  const [isLoadingAddTodo, setIsLoadingAddTodo] = useState(false);
+
+  const isAllCompleted = todos.every(todo => todo.completed);
+
+  const visibleTodos = useMemo(() => {
+    return todos.filter((todo) => {
+      switch (filterOption) {
+        case FilterOption.Active:
+          return !todo.completed;
+
+        case FilterOption.Completed:
+          return todo.completed;
+
+        default:
+          return FilterOption.All;
+      }
+    });
+  }, [filterOption, todos]);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((todosFromServer) => {
+        setError(null);
+        setTodos(todosFromServer);
+      })
+      .catch(() => {
+        setError('Unable to upload todos');
+      });
+  }, []);
+
+  useEffect(() => {
+    let errorTimer: number;
+
+    if (error) {
+      errorTimer = window.setTimeout(() => {
+        setError(null);
+      }, 3000);
+    }
+
+    return () => {
+      clearTimeout(errorTimer);
+    };
+  }, [error]);
+
+  const addTodo = useCallback(async (title: string) => {
+    setIsLoadingAddTodo(true);
+
+    const newTodo = {
+      id: 0,
+      completed: false,
+      title,
+      userId: USER_ID,
+    };
+
+    setTempTodo({ ...newTodo });
+
+    try {
+      const createdTodo = await createTodo(newTodo);
+
+      setTodos(prevTodos => [...prevTodos, createdTodo]);
+    } catch (err) {
+      setError('Unable to add a todo');
+    } finally {
+      setTempTodo(null);
+      setIsLoadingAddTodo(false);
+    }
+  }, []);
+
+  const removeTodo = useCallback((todoId: number) => {
+    setLoadingItems(prevItems => [...prevItems, todoId]);
+
+    deleteTodo(todoId)
+      .then(() => {
+        setTodos(prevTodos => prevTodos.filter(
+          todo => todo.id !== todoId,
+        ));
+      })
+      .catch(() => {
+        setError('Unable to delete a todo');
+      })
+      .finally(() => {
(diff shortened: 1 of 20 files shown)


src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+        setError(null);
+      }, 3000);
+    }
+
+    return () => {
+      clearTimeout(errorTimer);
+    };
+  }, [error]);
+
+  const addTodo = useCallback(async (title: string) => {
```
tiserett, line 75: Please, do not use useCallback / useMemo hooks until you see real problems with optimisation in your application. This is definitely not the case to use them. You pass this function to Header component, which is not even wrapped in memo, please, make sure to read more about useCallback  / useMemo, when and how to appropriately use them

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+    try {
+      const createdTodo = await createTodo(newTodo);
+
+      setTodos(prevTodos => [...prevTodos, createdTodo]);
+    } catch (err) {
+      setError('Unable to add a todo');
+    } finally {
+      setTempTodo(null);
+      setIsLoadingAddTodo(false);
+    }
```
tiserett, line 96: here you use try/catch/finally with async/await syntax, above you use chaining with then + catch, why don't you use the same approach everywhere? I suggest rewriting evrerything with modern async/await syntax

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+    completedTodos.forEach(todo => removeTodo(todo.id));
+  }, [removeTodo, todos]);
+
+  const editTodo = useCallback(async (
+    todoId: number,
+    args: UpdateTodoArgs,
+  ) => {
+    setLoadingItems(prevItems => [...prevItems, todoId]);
+
+    const updatedTodo: UpdateTodoArgs = { ...args };
```
tiserett, line 130: not sure if you shold set type here, since it should inherit it from args variable

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+
+    if (isAllCompleted) {
+      todos.forEach(todo => {
+        editTodo(
+          todo.id,
+          { completed: !isAllCompleted },
+        );
+      });
+    }
+  }, [editTodo, isAllCompleted, todos]);
```
tiserett, line 173:
```suggestion
  const toggleTodosCompleted = useCallback(() => {
    todos.forEach(todo => {
        if (!todo.completed) {
          editTodo(
            todo.id,
            { completed: !isAllCompleted },
          );
        }
      });
  }, [editTodo, isAllCompleted, todos]);
```

shouldn't it work just fine if you write it like this?

src/Components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+      <a
+        href="#/completed"
+        className={cn('filter__link', {
+          selected: filterOption === FilterOption.Completed,
+        })}
+        onClick={() => setFilterOption(FilterOption.Completed)}
+      >
+        Completed
+      </a>
```
tiserett, line 44: you can simplify this code by iterating through your enum, use google to learn how to do it ;)

tiserett (changes requested):
looks good, but please, get rid of all of these useCallback / useMemo hooks, and make sure to read, when it's really necessary to use them

src/App.tsx
vladyslav-lapshyn: It doesn't work, in your way I can change todos only when them uncompleted, but can't change when they all completed. But I've tryed do something similar

src/Components/TodoFilter/TodoFilter.tsx
vladyslav-lapshyn: @tiserett amazing idea. thanks 🔥

src/App.tsx
vladyslav-lapshyn: yeap, you are right it my mis

src/App.tsx
vladyslav-lapshyn: 😢

src/Components/Header/Header.tsx
```diff
@@ -0,0 +1,84 @@
...
+  isAllCompleted: boolean;
+}
+
+export const Header: FC<Props> = ({
+  addTodo,
+  setErrorMessage,
+  toggleTodosCompleted,
+  isAllCompleted,
+}) => {
+  const [value, setValue] = useState<string>('');
```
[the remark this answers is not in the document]
vladyslav-lapshyn: @tiserett yeap, it's good idea, but I in my opinion maybe something like `todoTitle` would be better. `todo` can confuse

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+
+      if (!isAllCompleted) {
+        if (!todo.completed) {
+          editTodo(
+            todo.id,
+            { completed: !isAllCompleted },
+          );
+        }
+      }
+    });
```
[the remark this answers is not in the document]
vladyslav-lapshyn: @tiserett yeap, thanks a lot)

src/Components/Header/Header.tsx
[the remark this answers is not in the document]
vladyslav-lapshyn: @tiserett I change than it but I name it `todoTitle`. In my opinion it sounds better, name like `todo` can confuse. I hope it's not a misstake

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,148 @@
...
+            className="todo__title"
+            onDoubleClick={handleDoubleClick}
+          >
+            {todo.title}
+          </span>
+
+          <button
+            type="button"
+            className="todo__remove"
+            onClick={() => removeTodo(todo.id)}
```
33Orange, line 131: Create a separate handler function

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,148 @@
...
+      { title: newTitle },
+    );
+
+    setIsEditing(false);
+  };
+
+  const handleKeyUp = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.code !== 'Escape') {
+      return;
+    }
```
33Orange, line 80: When I click `ESC` edit should be canceled and return to the previous value
image
image

src/Components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,37 @@
...
+interface Props {
+  filterOption: FilterOption;
+  setFilterOption: (filterOption: FilterOption) => void;
+}
+
+export const TodoFilter: FC<Props> = ({
+  filterOption,
+  setFilterOption,
+}) => {
+  const filterOptions = Object.keys(FilterOption);
```
33Orange, line 14: Let's use entries to avoid `const filterOptionKey = option as keyof typeof FilterOption;`

```suggestion
  const filterOptions = Object.Entries(FilterOption);
```

33Orange (changes requested):
Almost done

src/Components/TodoItem/TodoItem.tsx
vladyslav-lapshyn: @33Orange yeap, it's my mistake, I thought that changes should be saved

src/Components/Header/Header.tsx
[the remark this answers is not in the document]
tiserett: todoTitle sound amazing, even better then 'todo' 🚀🔥

tiserett (approved):
LGTM 🔥🔥🔥

PR #619, opened 2023-07-05
solution

annholovko:
[DEMO LINK](https://annholovko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,220 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+/* eslint-disable react-hooks/rules-of-hooks */
+import React, {
+  useMemo,
+  useEffect,
+  useState,
+  useCallback,
+} from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo, UpdateTodoArgs } from './types/Todo';
+import {
+  createTodo,
+  deleteTodo,
+  getTodos,
+  patchTodo,
+} from './api/todos';
+import { TodoForm } from './components/TodoForm';
+import { Error } from './components/Error';
+import { Filters } from './types/Filters';
+import { Filter } from './components/Filter';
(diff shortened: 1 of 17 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,138 @@
...
+          <input
+            type="text"
+            className="todo__title-field"
+            placeholder="Empty todo will be deleted"
+            value={newTitle}
+            onChange={handleTitleChange}
+            onKeyUp={handleKeyUp}
+            ref={inputRef}
+          />
+        </form>
```
dffUqp, line 111: you use `handleSubmitTitle` in your `handleKeyUp` function, why do we need `handleSubmitTitle` in `onSubmit`?

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+
+      setTodos((prevTodos) => [...prevTodos, createdTodo]);
+    } catch {
+      setError('Unable to add a todo');
+    } finally {
+      setTempTodo(null);
+    }
+  }, []);
+
+  const removeTodo = useCallback(async (todoId: number) => {
```
dffUqp, line 99: unnecessary `useCallback` because your `TodoList` component is not memoized

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+      default:
+        return todos;
+    }
+  }, [todos, filter]);
+
+  const handleCloseError = () => {
+    setError(null);
+  };
+
+  const addTodo = useCallback(async (title: string) => {
```
dffUqp, line 76: unnecessary useCallback because your `TodoForm` component is not memoized

dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

dffUqp (changes requested):
needs a few fixes:upside_down_face:

I can see, that the filter buttons slide to the right of the footer when there are no completed todos. Check [the working example](https://mate-academy.github.io/react_todo-app-with-api/).

image

PR #622, opened 2023-07-05
solution

snuber-w:
- [DEMO LINK](https://snuber-w.github.io/react_todo-app-with-api/)


silchyn (changes requested):
image

Also, `clear completed` button shouldn't be visible if there are no completed todos

PR #624, opened 2023-07-05
undercooked solution

ulositska:
[DEMO LINK](https://ulositska.github.io/react_todo-app-with-api/)


ulositska:
Well, autofocus attribute helped, now double click works as expected. 
But I think, It can't be the best solution, if I needed to turn off eslint for it. And as I understand, usage of autofocus leads to problems with accessibility((

silchyn (changes requested):
Why just putting the `onDblClick` event doesn't work? (it should)

image

PR #634, opened 2023-07-06
Todo-app-with-api

Mykola-Maik:
[DEMO LINK](https://Mykola-Maik.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,191 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback,
+  useEffect,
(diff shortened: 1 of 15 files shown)


silchyn (changes requested):
> cancel editing on Esс key keyup event

image

Moroz-Dmytro (changes requested):
1. With no Internet, I tried to add todo and I see the next bug
image

2. When I tried to update todo without the Internet I got notification that I can't do it, that's good
image
But when I open todo in edit mode again, the name was changed:
image

3. From task description "if new title is the same as the old one just cancel editing", but you send API call
image
4. From task description
image
5. Todo is displaying without spaces, but when I open it in edit mode spaces are visible

PR #635, opened 2023-07-06
add functionality to todo-app-with-api

cephy77:
[DEMO LINK](https://cephy77.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,156 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { ErrorMessage } from './components/ErrorMessage/ErrorMessage';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/Todo';
+import {
+  addTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+} from './api/todos';
+import { TodoFilter } from './components/TodoFilter/TodoFilter';
+import { FilterBy } from './utils/enums';
+import { TodoForm } from './components/TodoForm/TodoForm';
+import { TodoPatch } from './types/TodoPatch';
 
-const USER_ID = 0;
+const USER_ID = 10897;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<null | Todo>(null);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+  const [filterBy, setFilterBy] = useState(FilterBy.All);
+  const [errorText, setErrorText] = useState<null | string>(null);
+
+  const resetError = () => {
+    setErrorText(null);
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch((error) => {
+        setErrorText(error.message);
+        setTimeout(resetError, 3000);
+      });
+  }, []);
+
+  const onFilterChange = useCallback((filter: FilterBy) => {
+    setFilterBy(filter);
+  }, []);
+
+  const onAddTodo = async (title: string) => {
+    if (!/\S/g.test(title)) {
+      setErrorText('Title can\'t be empty');
+
+      return false;
+    }
+
+    const newTodo = {
+      title,
+      userId: USER_ID,
+      completed: false,
+    };
+
+    setTempTodo({ ...newTodo, id: 0 });
+
+    try {
+      const addedTodo = await addTodo(newTodo);
+
+      setTodos(prevTodos => [...prevTodos, addedTodo]);
+    } catch {
+      setErrorText('Unable to add a todo');
+    } finally {
+      setTempTodo(null);
+    }
+
+    return true;
+  };
+
+  const onDeleteTodo = useCallback(async (todoId: number) => {
+    setLoadingTodoIds(prevIds => [...prevIds, todoId]);
+
+    try {
+      await deleteTodo(todoId);
+
+      setTodos(prevTodos => prevTodos.filter(todo => todo.id !== todoId));
+    } catch {
+      setErrorText('Unable to delete a todo');
+    } finally {
+      setLoadingTodoIds(prevIds => prevIds.filter(id => id !== todoId));
+    }
+  }, []);
+
+  const onUpdateTodo = useCallback(async (todoId: number, data: TodoPatch) => {
+    setLoadingTodoIds(prevIds => [...prevIds, todoId]);
+
+    try {
+      await updateTodo(todoId, data);
+
+      setTodos(prevTodos => prevTodos.map(todo => {
+        if (todo.id === todoId) {
+          return {
+            ...todo,
+            ...data,
+          };
+        }
+
+        return todo;
+      }));
+    } catch {
+      setErrorText('Unable to update a todo');
+    } finally {
+      setLoadingTodoIds(prevIds => prevIds.filter(id => id !== todoId));
+    }
+  }, []);
+
+  const visibleTodos = filterBy === FilterBy.All
+    ? todos
+    : todos.filter(todo => (
+      filterBy === FilterBy.Active
+        ? !todo.completed
+        : todo.completed
+    ));
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <TodoForm
+          todos={todos}
+          onAddTodo={onAddTodo}
+          onUpdateTodo={onUpdateTodo}
+        />
+
+        {Boolean(todos.length) && (
+          <>
+            <TodoList
+              todos={visibleTodos}
+              tempTodo={tempTodo}
+              onDelete={onDeleteTodo}
+              onUpdateTodo={onUpdateTodo}
+              loadingTodoIds={loadingTodoIds}
+            />
+            <TodoFilter
+              todos={todos}
+              statusFilter={filterBy}
+              onFilterChange={onFilterChange}
+              onDeleteTodo={onDeleteTodo}
+            />
+          </>
+        )}
+      </div>
+
+      <ErrorMessage
+        errorText={errorText}
+        clearError={resetError}
+      />
+    </div>
   );
 };
--- a/src/components/ErrorMessage/ErrorMessage.tsx
+++ b/src/components/ErrorMessage/ErrorMessage.tsx
@@ -1,0 +1,23 @@
+import React from 'react';
+import cN from 'classnames';
+
+type Props = {
+  errorText: null | string,
+  clearError: () => void,
+};
+
+export const ErrorMessage: React.FC<Props> = ({ errorText, clearError }) => (
+  <div
+    className={cN('notification is-danger is-light has-text-weight-normal', {
+      hidden: !errorText,
+    })}
+  >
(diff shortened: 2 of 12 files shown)


silchyn (changes requested):
> cancel editing on Esс key keyup event

silchyn (changes requested):
image

Moroz-Dmytro (changes requested):
“Toggle all” is not active when all items selected
image
image

src/components/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import cN from 'classnames';
+
+type Props = {
+  errorText: null | string,
```
roman-mirzoian, line 5: The `errorText` prop can be marked as optional, because a `null` value can be passed. 
Update the type to `errorText?: string` for cleaner code.

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const hasTodos = Boolean(todos.length);
+  const hasActiveTodos = todos.filter(todo => !todo.completed);
+  const hasAllCompleted = todos.every(todo => todo.completed);
+
+  const handleSubmit = async (event: React.FormEvent) => {
+    event.preventDefault();
+
+    setIsLoading(true);
+
+    await onAddTodo(todoTitle);
```
roman-mirzoian, line 29: Asynchronous requests should be placed in the middle of a `try-catch` block to catch errors.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+  onUpdateTodo?: (todoId: number, data: TodoPatch) => void,
+  isLoading?: boolean
+};
+
+export const TodoItem: React.FC<Props> = memo(
+  ({
+    todo,
+    onDelete,
+    onUpdateTodo = () => { },
+    isLoading = true,
```
roman-mirzoian, line 18: Default value for `isLoading` prop is set to `true`. It is better to set default to `false`, as usually you don't want to show loading state until a request hasn't started.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+        setNewTitle(title);
+      }
+    };
+
+    return (
+      <div
+        className={cN('todo', {
+          completed,
+        })}
+        key={todo.id}
```
roman-mirzoian, line 71: `Key` prop is unnecessary in this component as it is not part of an array or iterator.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+    } = todo;
+
+    const [isEditing, setIsEditing] = useState(false);
+    const [newTitle, setNewTitle] = useState(title);
+
+    const handleDeleteTodo = () => onDelete(id);
+
+    const handleToggleTodo = () => onUpdateTodo(id, { completed: !completed });
+
+    const handleToggleEgiting = () => setIsEditing(state => !state);
```
roman-mirzoian, line 33: Typo in `handleToggleEgiting`. It should be `handleToggleEditing`.

src/App.tsx
```diff
@@ -1,24 +1,160 @@
...
+
+    const newTodo = {
+      title,
+      userId: USER_ID,
+      completed: false,
+    };
+
+    setTempTodo({ ...newTodo, id: 0 });
+
+    try {
```
roman-mirzoian, line 62: The `catch` block is missing an `error` parameter. 
This could be useful for logging or displaying a more specific error message. 
It should look like this: `catch (error)`.

src/App.tsx
```diff
@@ -1,24 +1,160 @@
...
+
+  const visibleTodos = filterBy === FilterBy.All
+    ? todos
+    : todos.filter(todo => (
+      filterBy === FilterBy.Active
+        ? !todo.completed
+        : todo.completed
+    ));
+
   if (!USER_ID) {
```
roman-mirzoian, line 120: This `if` statement will always be false, because `USER_ID` is set to 10897. Consider removing this check.

src/App.tsx
```diff
@@ -1,24 +1,160 @@
...
+  };
+
+  const onDeleteTodo = useCallback(async (todoId: number) => {
+    setLoadingTodoIds(prevIds => [...prevIds, todoId]);
+
+    try {
+      await deleteTodo(todoId);
+
+      setTodos(prevTodos => prevTodos.filter(todo => todo.id !== todoId));
+    } catch {
```
roman-mirzoian, line 82: The `catch` block is missing an `error` parameter. 
This could be useful for logging or displaying a more specific error message. 
It should look like this: `catch (error)`.

src/App.tsx
```diff
@@ -1,24 +1,160 @@
...
+        if (todo.id === todoId) {
+          return {
+            ...todo,
+            ...data,
+          };
+        }
+
+        return todo;
+      }));
+    } catch {
```
roman-mirzoian, line 105: The `catch` block is missing an `error` parameter. 
This could be useful for logging or displaying a more specific error message. 
It should look like this: `catch (error)`.

src/App.tsx
```diff
@@ -1,24 +1,160 @@
...
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <TodoForm
+          todos={todos}
+          onAddTodo={onAddTodo}
+          onUpdateTodo={onUpdateTodo}
+        />
+
+        {Boolean(todos.length) && (
```
roman-mirzoian, line 135: It is not necessary to cast it directly to the `boolean` type, since dynamic type casting will work here.

src/App.tsx
```diff
@@ -1,24 +1,160 @@
...
+    }
+  }, []);
+
+  const visibleTodos = filterBy === FilterBy.All
+    ? todos
+    : todos.filter(todo => (
+      filterBy === FilterBy.Active
+        ? !todo.completed
+        : todo.completed
+    ));
```
roman-mirzoian, line 118: Now everything is written correctly, but such code is difficult to read and maintain. It's better to break large parts of code into smaller parts:
```
function filterTodos(todos, filterBy) {
    return todos.filter(({ completed }) => (
      filterBy === FilterBy.Active ? !completed : completed
    ));
}
    ....
let visibleTodos = todos;
if (filterBy !== FilterBy.All) {
    visibleTodos = filterTodos(todos, filterBy);
}
```

roman-mirzoian (changes requested):
The task is well done, but there are several points to improve the code.
There are also several comments on the functionality itself:
1. The error text now disappears before the entire error component disappears:
https://gyazo.com/21f9209083c2f8358810aaf2dae34b24
2. When updating the name there, after pressing enter, we do not exit edit mode:
https://gyazo.com/afe46c92f27f3824fe65f28ac5eb0b80

src/components/TodoForm/TodoForm.tsx
cephy77: @roman-mirzoian I have a try-catch construction inside this function at App.tsx

src/App.tsx
cephy77: @roman-mirzoian I don't quite understand this. Should I leave here only todos.length or !!todos.length?

Moroz-Dmytro (changes requested):
Well done 🔥 
Only small fixes

src/App.tsx
roman-mirzoian: @cephy77 you can leave just `{ todos.length && ( ...`
You can read more about types conversions here:
https://www.freecodecamp.org/news/js-type-coercion-explained-27ba3d9a2839

src/components/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  isVisible: boolean,
+  errorText?: null | string,
```
roman-mirzoian, line 6: The `errorText` prop should be required. As it stands, it's possible to show an error message without any text. To improve the code, the `errorText` prop shouldn't be optional and shouldn't allow null values.

src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <TodoForm
+          todos={todos}
+          onAddTodo={onAddTodo}
+          onUpdateTodo={onUpdateTodo}
+        />
+
+        {Boolean(todos.length) && (
```
roman-mirzoian, line 136: It's unnecessary to use Boolean in Boolean(todos.length). You can use todos.length directly to check if an array is empty or not.

src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
+        return todo;
+      }));
+    } catch (error) {
+      setError('Unable to update a todo');
+    } finally {
+      setLoadingTodoIds(prevIds => prevIds.filter(id => id !== todoId));
+    }
+  }, []);
+
+  const filterTodos = (todosArray: Todo[], filter: FilterBy) => {
```
roman-mirzoian, line 113: This function `filterTodos` could be moved outside of the component. This would improve performance as it won't be recreated each time the component rerenders. Remember to pass all dependencies as parameters to the function.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+    };
+
+    const handleBlur = (event: React.FormEvent) => {
+      handleSubmit(event);
+    };
+
+    const handleCancelEditing = (
+      event: React.KeyboardEvent<HTMLInputElement>,
+    ) => {
+      if (event.key === 'Escape') {
```
roman-mirzoian, line 61: It would be cool not to use magic values and put such strings into constants and reuse them:
```
const ESCAPE = 'Escape';
```
This will increase the readability of the code and save us from possible errors.
This applies to development in general.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+      title,
+      completed,
+    } = todo;
+
+    const [isEditing, setIsEditing] = useState(false);
+    const [newTitle, setNewTitle] = useState(title);
+
+    const handleDeleteTodo = () => onDelete(id);
+
+    const handleToggleTodo = () => onUpdateTodo(id, { completed: !completed });
```
roman-mirzoian, line 31: If your `onUpdateTodo` is not a mandatory `prop` (so it is written in your type `Props`), then it is worth checking first whether `onUpdateTodo` has now been passed, and only then use it.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+  ({
+    todos,
+    statusFilter,
+    onFilterChange,
+    onDeleteTodo,
+  }) => {
+    const activeTodosNumber = todos.filter(todo => !todo.completed).length;
+    const hasCompletedTodos = todos.length !== activeTodosNumber;
+
+    const clearCompleted = () => {
```
roman-mirzoian, line 23: `clearCompleted` function can be taken outside the component and used as a helper.

src/App.tsx
roman-mirzoian: Now you catch the `error` parameter, but you don't save it in any way. It should be returned along with the error text:
```
setError(`Unable to add a todo: ${error.message}`);
```

src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
+    return todosArray.filter(({ completed }) => (
+      filter === FilterBy.Active ? !completed : completed
+    ));
+  };
+
+  let visibleTodos = todos;
```
roman-mirzoian, line 119: My example was for clarity, and to work with arrays, you should make copies of them, and not change them directly:
```
let visibleTodos = [...todos];
```

roman-mirzoian (changes requested):
The code is well written but needs some corrections.
I also see now that after adding a new todo and opening it for editing, empty spaces are displayed in different ways:
https://gyazo.com/d2b364da190e038ce34d90d893a4848a

src/App.tsx
roman-mirzoian: yes, you are right, then let's use `!!todos.length` or `todos.length > 0`

Moroz-Dmytro (approved):
LGTM 👍 
Only better will be move `constants` file from `utils` to `constants` folder
And `utils` to `types` folder

PR #636, opened 2023-07-06
Solution

RoksolanaTylipska:
[DEMO LINK](https://RoksolanaTylipska.github.io/react_todo-app-with-api/)


silchyn (changes requested):
- No loader after toggling todo's status
- I can't enter space when editing todo's title

PR #637, opened 2023-07-06
add solution

PanyaPrimal:
[DEMO LINK](https://PanyaPrimal.github.io/react_todo-app-with-api/)


silchyn (changes requested):
Todo with a title consisting of spaces can be added

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setNewTitle(event.target.value);
+  };
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Escape') {
+      handleCancelEditing();
+    } else if (event.key === 'Enter') {
+      handleSaveChanges();
+    }
```
roman-mirzoian, line 60: Before starting to work with functions and components, it would be very cool to define the variables that we use further into separate constants, so that we do not use magic values:
```
const ESCAPE= 'Escape';
const ENTER = 'Enter';
```
This will provide us with better readability of the code itself, conditions in the code and understanding that we will not accidentally make a mistake in a line.

Also, for better readability, it is not necessary to use the `if-else-if` construction:
```
if (event.key === ESCAPE) { ... }
if (event.key === ENTER ) { ... }
```

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export const TodoForm: React.FC<Props> = ({ handleAddTodo, setError }) => {
+  const [todoTitle, setTodoTitle] = useState('');
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
+
+  const handleFormSubmit = async (e:React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (!todoTitle.trim()) {
```
roman-mirzoian, line 15: Now, regardless of this check, you can add an empty todo (but an error will appear at the bottom) and add a todo whose name will have only spaces, so the check needs to be improved a little.

src/App.tsx
```diff
@@ -1,24 +1,212 @@
...
+
+      setTempTodo({
+        id: 0,
+        ...newTodo,
+      });
+
+      const createdTodo = await createTodo(newTodo);
+
+      setTodos(currentTodos => [...currentTodos, createdTodo]);
+    } catch {
```
roman-mirzoian, line 82: Adding a `catch` block without catching the error is a bad practice, because now we don't give the user an opportunity to understand what's wrong and only give a general answer ( `catch (error)` ) .
This applies to all code.

src/App.tsx
```diff
@@ -1,24 +1,212 @@
...
+      } as Todo;
+
+      await updateTodoTitle(updatedTodo);
+
+      setTodos(prevTodos => (
+        prevTodos.map(todo => (
+          todo.id !== todoToUpdate.id
+            ? todo
+            : updatedTodo
+        ))
```
roman-mirzoian, line 151: A similar piece of code is on lines 113-119, so you could put this logic in a separate handler and reuse it.

src/App.tsx
```diff
@@ -1,24 +1,212 @@
...
+      ));
+    } catch {
+      setError('Unable to update a todo');
+    } finally {
+      setLoadingTodos(currentTodosId => currentTodosId
+        .filter(todoId => todoId !== todoToUpdate.id));
```
roman-mirzoian, line 157: A similar piece of code is on lines 123-124, so you could put this logic in a separate handler and reuse it.

roman-mirzoian (changes requested):
The work is done well, but some suggestions to optimize the code and correct the logic of the application itself.

PR #640, opened 2023-07-08
Develop

vchervinko:
[DEMO LINK](https://vchervinko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,10 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useMemo, useState } from 'react';
+import React, {
+  useEffect,
+  useMemo,
+  useState,
+  useCallback,
+} from 'react';
 import { UserWarning } from './UserWarning';
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
@@ -60,7 +65,7 @@ export const App: React.FC = () => {
     }
   }, [todos, filter]);
 
-  const addTodo = async (title: string) => {
+  const addTodo = useCallback(async (title: string) => {
     try {
       const newTodo = {
         title,
@@ -81,9 +86,9 @@ export const App: React.FC = () => {
     } finally {
       setTempTodo(null);
     }
-  };
+  }, [todos]);
 
-  const deleteTodo = async (todoId: number) => {
+  const deleteTodo = useCallback(async (todoId: number) => {
     try {
       setLoadingTodos(prevIds => [...prevIds, todoId]);
       await removeTodo(todoId);
@@ -93,7 +98,7 @@ export const App: React.FC = () => {
     } finally {
       setLoadingTodos(prevIds => prevIds.filter(id => id !== todoId));
     }
-  };
+  }, [loadingTodos]);
 
   const handleDeleteCompletedButton = async () => {
     const deletePromises = completedTodos.map(todo => deleteTodo(todo.id));
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,210 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { TodoList } from './components/TodoList';
+import { TodoFilter } from './components/TodoFilter';
+import { Header } from './components/Header';
+import {
+  getTodos,
+  createTodo,
+  removeTodo,
+  updateTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Filters } from './types/filters';
 
-const USER_ID = 0;
+const USER_ID = 10953;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [err, setErr] = useState<string | null>(null);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodos, setLoadingTodos] = useState([0]);
+  const [filter, setFilter] = useState(Filters.ALL);
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+      })
+      .catch(() => setErr('Error: cannot upload todos'));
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,24 +1,215 @@
...
+import { Todo } from './types/Todo';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Filters } from './types/filters';
 
-const USER_ID = 0;
+const USER_ID = 10953;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [err, setErr] = useState<string | null>(null);
```
naziamoff, line 26: let's create enum for errors. It is difficult and not reliable to work with strings

src/App.tsx
```diff
@@ -1,24 +1,215 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodos, setLoadingTodos] = useState([0]);
```
naziamoff, line 28: why do we have array containing zero as a default state here?

src/App.tsx
```diff
@@ -1,24 +1,215 @@
...
+  const [filter, setFilter] = useState(Filters.ALL);
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
```
naziamoff, line 31: you now create these variables on each render, even if you don't need them
Let's just calculate it if you need it

src/App.tsx
naziamoff: also, why `err`, but not `error`? Looks better this way

src/App.tsx
```diff
@@ -1,24 +1,215 @@
...
+      await removeTodo(todoId);
+      setTodos(prevTodos => prevTodos.filter(todo => todo.id !== todoId));
+    } catch {
+      setErr('Unable to delete a todo');
+    } finally {
+      setLoadingTodos(prevIds => prevIds.filter(id => id !== todoId));
+    }
+  }, [loadingTodos]);
+
+  const handleDeleteCompletedButton = async () => {
```
naziamoff, line 103: do we handle buttons?

src/App.tsx
```diff
@@ -1,24 +1,215 @@
...
+    const deletePromises = completedTodos.map(todo => deleteTodo(todo.id));
```
naziamoff, line 104: by this name it seems to me we're dealing with a function

src/App.tsx
```diff
@@ -1,24 +1,215 @@
...
+
+  const handleToggleComplete = async (todoId: number) => {
+    setLoadingTodos([todoId]);
+
+    try {
+      const todoToUpdate = todos.find((todo) => todo.id === todoId);
+
+      if (todoToUpdate) {
+        const updatedTodo = await updateTodo(todoId,
+          { completed: !todoToUpdate.completed });
```
naziamoff, line 121:
```suggestion
         const updatedTodo = await updateTodo(
            todoId, 
            { completed: !todoToUpdate.completed }
         );
```

src/App.tsx
```diff
@@ -1,24 +1,215 @@
...
+            todos={visibleTodos}
+            filter={filter}
+            onChangeFilter={setFilter}
+            handleDeleteCompletedButton={handleDeleteCompletedButton}
+          />
+        )}
+      </div>
+
+      {/* Notification is shown in case of any error */}
+      {/* Add the 'hidden' class to hide the message smoothly */}
```
naziamoff, line 208: never leave comments unless you really have to
```suggestion
```

src/api/todos.ts
```diff
@@ -0,0 +1,22 @@
...
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const updateTodo = (
+  todoId: number,
+  todo: Partial<Pick<Todo, 'title' | 'completed'>>,
+) => {
+  return client.patch<Todo>(`/todos/${todoId}`, todo);
+};
+// Add more methods here
```
naziamoff, line 22:
```suggestion
```

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,41 @@
...
+  setErr: (err: string | null) => void;
+};
+
+export const ErrorNotification: React.FC<Props> = ({
+  err,
+  setErr,
+}) => {
+  const removeNotofication = () => {
+    setErr(null);
+  };
```
naziamoff, line 15: You have the handling stuff in the higher component - so let's be consistent

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,41 @@
...
+      />
+
+      {err}
+
+      {/* show only one message at a time
+      Unable to add a todo
+      <br />
+      Unable to delete a todo
+      <br />
+      Unable to update a todo */}
```
naziamoff, line 38: ?

src/components/Header.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <form
+        onSubmit={handleSubmit}
+      >
+        <input
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          value={newTodoTitle}
+          onChange={handleInputChange}
+          disabled={tempTodo !== null}
```
naziamoff, line 58: use shortcuts for booleans

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,74 @@
...
+  handleDeleteCompletedButton: () => void;
+};
+
+export const TodoFilter: React.FC<Props> = ({
+  todos,
+  filter,
+  onChangeFilter,
+  handleDeleteCompletedButton,
+}) => {
+  const count = () => {
```
naziamoff, line 19: count what?

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,74 @@
...
+    return incompleteTodos.length;
+  };
+
+  return (
+    <footer className="todoapp__footer">
+      <span className="todo-count">
+        {`${count()} items left`}
+      </span>
+
+      {/* Active filter should have a 'selected' class */}
```
naziamoff, line 31: 👎

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,74 @@
...
+          className={cn('filter__link ', {
+            selected: filter === Filters.COMPLETED,
+          })}
+          onClick={() => onChangeFilter(Filters.COMPLETED)}
+        >
+          Completed
+        </a>
+      </nav>
+
+      {/* don't show this button if there are no completed todos */}
```
naziamoff, line 64: 👎

naziamoff (changes requested):

PR #642, opened 2023-07-09
Develop

Zibi95:
[DEMO LINK](https://Zibi95.github.io/react_todo-app-with-api/)


src/components/TodoError/TodoError.tsx
```diff
@@ -0,0 +1,36 @@
...
+};
+
+export const TodoError = ({ errorMsg }: TodoErrorProps) => {
+  const [hidden, setHidden] = useState(true);
+
+  useEffect(() => {
+    if (errorMsg) {
+      setHidden(false);
+    }
+  }, []);
```
darokrk, line 15: The dependency array in useEffect is empty, which means the effect will only run once after the initial render, and not on updates. It should include the errorMsg prop to ensure the error message is displayed whenever it changes.

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,42 @@
...
+            <CSSTransition
+              key={todo.id}
+              timeout={300}
+              classNames={classNames('item', {
+                'temp-item': todo.id === 0,
+              })}
+            >
+              <Todo
+                isProcessed={processedTodos.some(({ id }) => id === todo.id)}
+                key={todo.id}
```
darokrk, line 33: You have already defined a 'key' prop on the CSSTransition component. Defining another 'key' prop on the Todo component is redundant and will not be used.

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export const TodoList = ({ todos }: TodoListProps) => {
+  const { tempTodo, processedTodos } = useTodos();
+
+  return (
+    <section className="todoapp__main">
+      <TransitionGroup>
+        {[...todos, tempTodo].map((todo) => {
+          if (!todo) {
```
darokrk, line 19: You should avoid returning undefined in a map function. Instead, you can use filter method to exclude undefined todos before mapping over the todos array.

src/styles/itemTransition.scss
```diff
@@ -0,0 +1,50 @@
...
+    max-height: 58px;
+  }
+
+  &-enter-active {
+    overflow: hidden;
+    max-height: 58px;
+    transition: max-height 0.3s ease-in-out;
+  }
+
+  &-item-exit {
```
darokrk, line 34: The class name '.&-item-exit' seems incorrect. It should probably be '.&-exit' to be consistent with the rest of your code.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // we wait for testing purpose to see loaders
+  return wait(300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
+      if (!response.ok) {
+        throw new Error();
```
darokrk, line 34: Consider giving a more meaningful message in the Error object to help with debugging. Example: throw new Error('Failed to fetch data')

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+      {editing ? (
+        <form
+          onSubmit={() => {
+            onTodoTitleChange(inputValue);
+          }}
+        >
+          <input
+            // eslint-disable-next-line jsx-a11y/no-autofocus
+            autoFocus
```
darokrk, line 96: Consider avoid using 'autoFocus' on interactive elements. This can be a problem for users who rely on a keyboard and can cause unexpected behavior.

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,129 @@
...
+  return (
+    <div
+      onDoubleClick={() => setEditing(true)}
+      className={classNames('todo', { completed })}
+    >
+      <label className="todo__status-label">
+        <input
+          type="checkbox"
+          className="todo__status"
+          defaultChecked={completed}
```
darokrk, line 80: You should use the 'checked' property instead of 'defaultChecked' to make this input controlled.

darokrk (approved):
Nice PR 🏆 ! Left some suggestions, and give approval so as not to hold up your work, you can fix them before moving next but they are not necessary, good job 👍🏽

PR #643, opened 2023-07-09
update todos

AlexKherson:
[DEMO LINK](https://alexKherson.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,46 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect } from 'react';
 import { UserWarning } from './UserWarning';
+import { TodoContent } from './components/TodoContent';
+import { Notifications } from './components/Notifications';
+import { todosApi } from './api/todos-api';
+import { useTodoContext } from './context/todoContext/useTodoContext';
+import { useErrorContext } from './context/errorContext/useErrorContext';
 
-const USER_ID = 0;
+const USER_ID = 10875;
 
 export const App: React.FC = () => {
+  const { setTodos } = useTodoContext();
+  const { errorMessage, notifyAboutError } = useErrorContext();
+
+  useEffect(() => {
+    todosApi.getByUser(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        notifyAboutError('Something went wrong');
+      });
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+  const closeNotification = () => {
+    notifyAboutError('');
+  };
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+  return (
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <TodoContent />
+      {errorMessage && (
+        <Notifications
+          onClose={closeNotification}
+          errorMessage={errorMessage}
+        />
+      )}
+    </div>
   );
 };
--- a/src/components/Notifications.tsx
+++ b/src/components/Notifications.tsx
@@ -1,0 +1,28 @@
+import { FC, useEffect } from 'react';
+
+interface NotificationsProps {
+  onClose: () => void,
+  errorMessage: string,
+}
+
+export const Notifications: FC<NotificationsProps> = (props) => {
+  const { onClose, errorMessage } = props;
+
+  useEffect(() => {
+    setTimeout(() => {
+      onClose();
+    }, 3000);
+  }, []);
+
+  return (
+    <div className="notification is-danger is-light has-text-weight-normal">
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+      <button
+        type="button"
+        className="delete"
+        onClick={onClose}
+      />
+      <p>{errorMessage}</p>
+    </div>
+  );
+};
--- a/src/components/TodoContentFooter.tsx
+++ b/src/components/TodoContentFooter.tsx
@@ -1,0 +1,77 @@
+import { FC } from 'react';
+import classNames from 'classnames';
+import { getEnumKeys } from '../helpers/getEnumKeys';
+import { TodoStatus } from '../types/TodoStatus';
+import { useTodoContext } from '../context/todoContext/useTodoContext';
+import { todosApi } from '../api/todos-api';
+import { useErrorContext } from '../context/errorContext/useErrorContext';
+
+interface TodoContentFooterProps {
+  selectedStatusTodo: TodoStatus,
+  onSelectStatusTodo: (status: TodoStatus) => void
+}
+
+export const TodoContentFooter: FC<TodoContentFooterProps> = (props) => {
+  const {
+    selectedStatusTodo,
+    onSelectStatusTodo,
+  } = props;
+
+  const { todos, setHandlingTodoIds } = useTodoContext();
+  const { notifyAboutError } = useErrorContext();
+  const { size, countCompleted, removeTodos } = useTodoContext();
+  const statusKeys = getEnumKeys(TodoStatus);
+
+  const onRemoveCompletedTodos = async () => {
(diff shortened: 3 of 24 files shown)


src/components/TodoContentHeader.tsx
```diff
@@ -0,0 +1,111 @@
...
+      notifyAboutError('Unable to update todos');
+    } finally {
+      setHandlingTodoIds([]);
+    }
+  };
+
+  return (
+    <header className="todoapp__header">
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+      <button
```
roman-mirzoian, line 92: Using eslint-disable-line to ignore accessibility rules should be avoided. Instead, add appropriate labels to your controls.

src/components/TodoContentMain.tsx
```diff
@@ -0,0 +1,213 @@
...
+    } finally {
+      setHandlingTodoIds([]);
+    }
+  };
+
+  const onChangeStatus = (id: number, isChecked: boolean) => {
+    onEdit(id, { completed: isChecked });
+  };
+
+  const handlerDoubleClick = (countClick: number, id: number) => {
```
roman-mirzoian, line 71: Instead of using a counter to track double clicks, try to use `onDoubleClick` event.

src/components/TodoContentMain.tsx
```diff
@@ -0,0 +1,213 @@
...
+    const trimmedTitle = todoTitleRef.current?.value.trim();
+
+    if (!trimmedTitle) {
+      onRemove(id);
+
+      return;
+    }
+
+    if (todoTitleRef.current) {
+      await onEdit(id, { title: todoTitleRef.current.value });
```
roman-mirzoian, line 103: You are using `todoTitleRef.current.value` directly to update the Todo. It's better to use `trimmedTitle` that you created at line 94, in order to ensure that the updated title is the trimmed version.

src/components/Notifications.tsx
```diff
@@ -0,0 +1,28 @@
...
+  errorMessage: string,
+}
+
+export const Notifications: FC<NotificationsProps> = (props) => {
+  const { onClose, errorMessage } = props;
+
+  useEffect(() => {
+    setTimeout(() => {
+      onClose();
+    }, 3000);
```
roman-mirzoian, line 14: There is no `clearTimeout` in the return function of the `useEffect` hook. This can potentially lead to memory leaks or unexpected behavior if the component unmounts before the timeout has finished.

src/components/Notifications.tsx
```diff
@@ -0,0 +1,28 @@
...
+  }, []);
```
roman-mirzoian, line 15: The `useEffect` hook is set to run only once (empty dependency array). This means that the timeout will be set only when the component is first mounted and not whenever the `errorMessage` prop changes. This may not be the desired behavior if you want the notification to automatically close 3 seconds after each error message is received.

src/context/errorContext/errorContextProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+  useState,
+} from 'react';
+import { ErrorContext, ErrorContextProps } from './errorContext';
+
+interface Props {
+  children: ReactNode,
+}
+
+export const ErrorContextProvider: FC<Props> = memo(({ children }) => {
+  const [errorMessage, setErrorMessage] = useState('');
```
roman-mirzoian, line 15: The default value of errorMessage is an empty string. It might be more clear to make the default value `null` to indicate that there is no error.

src/components/TodoContentFooter.tsx
```diff
@@ -0,0 +1,77 @@
...
+  onSelectStatusTodo: (status: TodoStatus) => void
+}
+
+export const TodoContentFooter: FC<TodoContentFooterProps> = (props) => {
+  const {
+    selectedStatusTodo,
+    onSelectStatusTodo,
+  } = props;
+
+  const { todos, setHandlingTodoIds } = useTodoContext();
```
roman-mirzoian, line 20: The `useTodoContext` hook is called twice. Instead of calling it twice, call it only once and destructure all necessary variables at once.

src/App.tsx
```diff
@@ -1,24 +1,46 @@
...
+
+  useEffect(() => {
+    todosApi.getByUser(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        notifyAboutError('Something went wrong');
+      });
+  }, []);
+
   if (!USER_ID) {
```
roman-mirzoian, line 25: The condition `!USER_ID` will never be true because `USER_ID` is a constant and it's not zero. Either make `USER_ID` dynamic or remove this condition if it's not required.

src/App.tsx
```diff
@@ -1,24 +1,46 @@
...
 
 export const App: React.FC = () => {
+  const { setTodos } = useTodoContext();
+  const { errorMessage, notifyAboutError } = useErrorContext();
+
+  useEffect(() => {
+    todosApi.getByUser(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        notifyAboutError('Something went wrong');
```
roman-mirzoian, line 21: In case of an `error`, you should provide more descriptive messages or use the error message from the API response.

src/index.tsx
```diff
@@ -5,6 +5,18 @@ import '@fortawesome/fontawesome-free/css/all.css';
...
 import { App } from './App';
+import { TodoContextProvider } from './context/todoContext/todoContextProvider';
+import {
+  ErrorContextProvider,
+} from './context/errorContext/errorContextProvider';
 
 createRoot(document.getElementById('root') as HTMLDivElement)
-  .render(<App />);
+  .render(
+    <>
```
roman-mirzoian, line 15: In this case, you don't need to use the React wrapper fragment, because the rendering result wraps the `TodoContextProvider`.

src/context/todoContext/todoContext.tsx
```diff
@@ -0,0 +1,26 @@
+import { createContext } from 'react';
+import { Todo } from '../../types/Todo';
+
+export interface TodoContextProps {
+  todos: Todo[],
+  size: number,
+  countCompleted: number,
+  setTodos: (todos: Todo[]) => void,
+  addTodo: (todo: Todo) => void;
```
roman-mirzoian, line 9: Be careful, because after defining the type, there should not be `;`.

src/components/TodoContentMain.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+  const onRemove = async (id: number) => {
+    setHandlingTodoIds([id]);
+    try {
+      const result = await todosApi.remove([id]);
+
+      if (result) {
+        removeTodos([id]);
+      }
+    } catch {
```
roman-mirzoian, line 44: In the `catch` block, you should also catch a possible `error`, because now the error does not look informative, for example, you can do this:
```
catch(error) { notifyAboutError(`Unable to delete a todo: ${error.message}`)}
```

src/components/TodoContentMain.tsx
```diff
@@ -0,0 +1,213 @@
...
+  const onEdit = async (id: number, data: UpdateTodoArgs) => {
+    try {
+      setHandlingTodoIds([id]);
+      const updatedTodos = await todosApi.update([{
+        id,
+        data,
+      }]);
+
+      updateTodos(updatedTodos);
+    } catch {
```
roman-mirzoian, line 60: In the `catch` block, you should also catch a possible `error`, because now the error does not look informative, for example, you can do this:
```
catch(error) { notifyAboutError(`Unable to update a todo: ${error.message}`)}
```

src/components/TodoContentHeader.tsx
```diff
@@ -0,0 +1,111 @@
...
+      });
+      setIsHandleRequest(true);
+      const createdTodo = await todosApi.create({
+        title,
+        completed: false,
+        userId: 10875,
+      });
+
+      addTodo(createdTodo);
+    } catch {
```
roman-mirzoian, line 50: In the `catch` block, you should also catch a possible `error`, because now the error does not look informative, for example, you can do this:
```
catch(error) { notifyAboutError(`Unable to add a todo: ${error.message}`)}
```
This applies to all code.

roman-mirzoian (changes requested):
The work is done well and only a few points need to be refined.
It is also necessary to update the functionality of the application itself, because now you can create a todo with an empty name using only spaces.

image

src/context/todoContext/todoContext.tsx
tiserett: fix Roman's comment

tiserett (changes requested):
good work, but make sure to fix all my comments, and those which Roman left, but you haven't fixed ;)

PR #647, opened 2023-07-14
solution

vitvadym:
[DEMO LINK](https:/vitvadym.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,188 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  addTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { getFilteredTodos } from './heplers/getFilteredTodos';
+import { SortType } from './enum/SortType';
+import { TodoHeader } from './components/TodoHeader';
+import { TodoList } from './components/TodoList';
+import { TodoFooter } from './components/TodoFooter';
 
-const USER_ID = 0;
+const USER_ID = 10922;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortBy, setSortBy] = useState(SortType.All);
+  const [error, setError] = useState<string | null>(null);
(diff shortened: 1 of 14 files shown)


src/components/TodoEdit.tsx
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    setIsEditing(false);
+  };
+
+  const handleBlur = (event:React.FormEvent<HTMLFormElement>
+  | React.FocusEvent<HTMLInputElement>) => {
+    if (isTitleChanged) {
+      return;
+    }
```
dinaminahi, line 85: You have this check in the handleSubmit function, so it is executed twice

src/App.tsx
```diff
@@ -1,24 +1,204 @@
...
+          isVisible={isVisibleClear}
+          onDeleteCompleted={handleDeleteAllCompleted}
+        />
+      </div>
+
+      {error && (
+        <div
+          className={`notification is-danger is-light
+          has-text-weight-normal ${!error ? 'hidden' : ''}`}
+        >
```
dinaminahi, line 193: use `classnames`

src/App.tsx
```diff
@@ -1,24 +1,204 @@
...
+    } catch {
+      setError('Unable to update a todo');
+    }
+  };
+
+  const filters: string[] = [
+    SortType.All,
+    SortType.Active,
+    SortType.Completed,
+  ];
```
dinaminahi, line 147: You don't have to create an array of filters here and pass it with props, instead use Object.values(Enum) directly in your footer

src/App.tsx
```diff
@@ -1,24 +1,204 @@
...
+      completed: false,
+      userId: USER_ID,
+    };
+
+    setTempTodo({
+      id: 0,
+      title,
+      completed: false,
+      userId: USER_ID,
+    });
```
dinaminahi, line 57:
```suggestion
    setTempTodo({
      id: 0,
     ...newTodo
    });
```

src/App.tsx
```diff
@@ -1,24 +1,204 @@
...
+    SortType.All,
+    SortType.Active,
+    SortType.Completed,
+  ];
+
+  const visibleTodos = getFilteredTodos(todos, sortBy);
+
+  const amountCompletedTodos = visibleTodos.filter(
+    todo => !todo.completed,
+  ).length;
```
dinaminahi, line 153: And here, you can use your helper function

dinaminahi (changes requested):
Great job 👍 
A few moments to fix :
In working example we have a placeholder, that says that empty todo will be deleted
image

PR #648, opened 2023-07-17
add task solution

ollavka:
[DEMO LINK](https://ollavka.github.io/react_todo-app-with-api/)


src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,39 @@
...
+import classNames from 'classnames';
+import { TodoContext } from './TodoContext';
+
+export const ErrorNotification: FC = () => {
+  const {
+    error,
+    onErrorChange,
+  } = useContext(TodoContext);
+
+  const errorMessage = useRef(error);
```
anstsot, line 14: try to use `error` only. no need to add ref

anstsot (changes requested):
Great job, just add one improvement and fix small bug:
it's possible to edit todo item and leave empty (few spaces: `  `)
image

src/components/ErrorNotification.tsx
ollavka: without using ref I have the text of the error disappears before the error itself and therefore there is a sharp jump when the text disappears, as it disappears smoothly, and with the help of ref I solve this problem)

anstsot: you can leave like this, but try also to add conditional rendering of `ErrorNotification` component in `TodoApp` (render only if there is error message)

anstsot (commented):
Good work, approving:sparkles:

anstsot (approved):

PR #650, opened 2023-07-18
solution

IvanFront-end:
[DEMO LINK](https://IvanFront-end.github.io/react_todo-app-with-api/)


sTorba24 (changes requested):
Almost good! Lets fix one thing

When i add new todo it's added to top of todos, but when reloaded it's at the bottom(Example todo added) So to be consistent add it to the bottom on front-side)
image
image

sTorba24 (approved):
Well done!

PR #654, opened 2023-07-21
solution

DanyloPasko:
[DEMO LINK](https://DanyloPasko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useEffect, useMemo, useState } from 'react';
 import cn from 'classnames';
 
@@ -16,7 +15,7 @@ const USER_ID = 10;
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [hasError, setHasError] = useState(false);
+  const [hasError, setHasError] = useState<boolean>(false);
   const [filterStatus, setFilterStatus] = useState(Status.ALL);
 
   const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
@@ -25,12 +24,12 @@ export const App: React.FC = () => {
     setEditingTodoId((prevTodoId) => (prevTodoId === todoId ? null : todoId));
   };
 
-  const handleSaveTodo = (editedTodo: { id: number; title: any; }) => {
+  const handleSaveTodo = (editedTodo: { id: number; title: string; }) => {
     updateTodoStatus(editedTodo.id, { title: editedTodo.title })
       .then(() => {
-        // eslint-disable-next-line max-len
-        setTodos((prevTodos) => prevTodos.map((todo) => (todo.id === editedTodo.id
-          ? { ...todo, title: editedTodo.title } : todo)));
+        setTodos(
+          (prevTodos) => prevTodos.map((todo) => (todo.id === editedTodo.id
+            ? { ...todo, title: editedTodo.title } : todo)));
       })
       .catch((error) => {
         setHasError(true);
@@ -79,7 +78,14 @@ export const App: React.FC = () => {
     const todoToUpdate = updatedTodos.find((todo) => todo.id === todoId);
 
     if (todoToUpdate) {
-      client.patch(`/todos/${todoId}`, { completed: todoToUpdate.completed });
+      client.patch(`/todos/${todoId}`, { completed: todoToUpdate.completed })
+        .catch((error) => {
+          // eslint-disable-next-line no-console
+          console.error('Failed to update todo status:', error);
+          setTodos(prevTodos => prevTodos.map(
+            todo => (todo.id === todoId
+              ? { ...todo, completed: !todo.completed } : todo)));
+        });
     }
   };
 
@@ -166,13 +172,13 @@ export const App: React.FC = () => {
       <div className="todoapp__content">
         <header className="todoapp__header">
           {todoIsActive && (
+            // eslint-disable-next-line jsx-a11y/control-has-associated-label
             <button type="button" className="todoapp__toggle-all active" />
           )}
           <form
             onSubmit={(event: React.FormEvent<HTMLFormElement>) => {
               event.preventDefault();
               const form = event.target as HTMLFormElement;
-              // eslint-disable-next-line max-len
               const newTodoTitle = form.newTodoTitle.value.trim();
 
               if (newTodoTitle) {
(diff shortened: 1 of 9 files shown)


src/api/todos.ts
```diff
@@ -0,0 +1,29 @@
...
+export const deleteTodo = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
+
+// export const updateTodoStatus = (todoId: number, completed: boolean) => {
+//   return client.put<Todo>(`/todos/${todoId}`, { completed });
+// };
+
+export const updateTodoStatus = (todoId: any, updatedData: any) => {
+  return client.patch(`/todos/${todoId}`, updatedData);
```
anastasiiavorobiova, line 28: Avoid using 'any' as a type declaration. Instead, use specific types to ensure type safety. For instance, todoId should be a number and updatedData could be a Partial<Todo> object.

src/Components/TodoEdit.tsx
```diff
@@ -0,0 +1,35 @@
+import { SetStateAction, useState } from 'react';
+import { Todo } from '../types/Todo';
+
+export const TodoEditForm = ({ todo, onEdit }
+: { todo: Todo; onEdit: (editedTodo: Todo) => void }) => {
+  const [title, setTitle] = useState(todo.title);
+
+  const handleTitleChange
+  = (event: { target: { value: SetStateAction<string>; }; }) => {
```
anastasiiavorobiova, line 9: The type of 'event.target.value' is not 'SetStateAction<string>'. It should be 'string'. SetStateAction is used with the setState function, not for event handling.

src/styles/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
anastasiiavorobiova, line 1: Avoid using 'eslint-disable'. It is generally a better practice to follow the linter's rules as it helps in maintaining the code quality and standards set by the community. In this case, rather than using 'any' for 'data', consider defining an interface that closely matches the structure of the data or use 'unknown' if the structure is not known.

src/styles/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // we wait for testing purpose to see loaders
+  return wait(300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
+      if (!response.ok) {
+        throw new Error();
```
anastasiiavorobiova, line 34: When throwing an Error, it is a good practice to provide a message describing what went wrong. This can help in debugging if this error ever gets thrown. You can make use of 'response.statusText' to get more information about the error.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,47 @@
...
+      return response.json();
+    });
+}
+
+export const client = {
+  get: <T>(url: string) => request<T>(url),
+  post: <T>(url: string, data: any) => request<T>(url, 'POST', data),
+  patch: <T>(url: string, data: any) => request<T>(url, 'PATCH', data),
+  delete: (url: string) => request(url, 'DELETE'),
+  put: <T>(url: string, data: any) => request<T>(url, 'PATCH', data),
```
anastasiiavorobiova, line 46: It appears that the 'put' method is using the 'PATCH' HTTP verb. This is likely a mistake, as 'put' should typically use the 'PUT' HTTP verb to replace a resource on the server.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
anastasiiavorobiova, line 1: Avoid using 'eslint-disable' if possible. It's better to follow the ESLint conventions and adjust the code accordingly. This specific rule '@typescript-eslint/no-explicit-any' is important because it encourages the use of specific types instead of 'any', which can lead to type errors.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,47 @@
...
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // we wait for testing purpose to see loaders
+  return wait(300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
+      if (!response.ok) {
+        throw new Error();
```
anastasiiavorobiova, line 34: When throwing an error, it is helpful to include a message that describes what went wrong. This can make debugging easier in the future.

src/App.tsx
```diff
@@ -1,24 +1,311 @@
...
+      return todo;
+    });
+
+    setTodos(updatedTodos);
+
+    const todoToUpdate = updatedTodos.find((todo) => todo.id === todoId);
+
+    if (todoToUpdate) {
+      client.patch(`/todos/${todoId}`, { completed: todoToUpdate.completed });
+    }
```
anastasiiavorobiova, line 83: The PATCH request to update the status of a todo item doesn't handle any potential errors. It would be good to add a .catch() block to handle any errors that may occur.

src/App.tsx
```diff
@@ -1,24 +1,311 @@
...
+  const [filterStatus, setFilterStatus] = useState(Status.ALL);
+
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
+
+  const handleEditTodo = (todoId: number) => {
+    setEditingTodoId((prevTodoId) => (prevTodoId === todoId ? null : todoId));
+  };
+
+  const handleSaveTodo = (editedTodo: { id: number; title: any; }) => {
+    updateTodoStatus(editedTodo.id, { title: editedTodo.title })
```
anastasiiavorobiova, line 29: The 'title' property of 'editedTodo' is of type 'any'. It's best to avoid using 'any' in TypeScript. Instead, use a more specific type, in this case probably 'string'.

src/App.tsx
```diff
@@ -1,24 +1,306 @@
...
+
 import { UserWarning } from './UserWarning';
+import { TodoEditForm } from './Components/TodoEdit';
 
-const USER_ID = 0;
+const USER_ID = 10;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState(false);
```
anastasiiavorobiova, line 19: Consider adding a generic type here

src/App.tsx
```diff
@@ -1,24 +1,306 @@
...
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
+
+  const handleEditTodo = (todoId: number) => {
+    setEditingTodoId((prevTodoId) => (prevTodoId === todoId ? null : todoId));
+  };
+
+  const handleSaveTodo = (editedTodo: { id: number; title: any; }) => {
+    updateTodoStatus(editedTodo.id, { title: editedTodo.title })
+      .then(() => {
+        // eslint-disable-next-line max-len
```
anastasiiavorobiova, line 31: Consider fixing issues instead of disabling eslint

anastasiiavorobiova (changes requested):
Great work! Added comments to fix

volodymyr-soltys97 (approved):

PR #655, opened 2023-07-21
React_todo-ap-with-api

ArtemKravcov85:
[DEMO LINK](https://ArtemKravcov85.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,243 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 
-const USER_ID = 0;
+import * as todoService from './api/todos';
+
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
+
+import { UserWarning } from './UserWarning';
+import { ErrorMessages } from './types/ErrorMessage';
+import { TodoList } from './components/TodoList/TodoList';
+import { Header } from './components/Header/Header';
+import { Footer } from './components/Footer/Footer';
+import { ErrorMessage } from './components/ErrorMessage/ErrorMessage';
+
+const USER_ID = 11074;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<ErrorMessages | null>(null);
+  const [filterStatus, setFilterStatus] = useState(Status.ALL);
+  const [todoTitle, setTodoTitle] = useState<string>('');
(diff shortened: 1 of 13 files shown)


witflash (changes requested):
Good work!

But I see a problem.
I can add an empty Todo (I also get an error, but the Todo has been created).
Could you, please check this trouble and try to fix it?

image

src/App.tsx
```diff
@@ -1,24 +1,245 @@
...
+      }
+    },
+    [],
+  );
+
+  const handleToggleAllTodosCompleted = () => {
+    setIsLoading(true);
+
+    try {
+      todos.map((todo) => handleToggleCompleted(todo.id, todo.completed));
```
lerastarynets, line 146: Same issue as above. You are using .map method which returns an array, it would be better to use .forEach method because you don't need to return an array.

src/App.tsx
```diff
@@ -1,24 +1,245 @@
...
+        <TodoList
+          visibleTodos={visibleTodos}
+          handleToggleCompleted={handleToggleCompleted}
+          handleRemoveTodo={handleRemoveTodo}
+          tempTodo={tempTodo}
+          isLoading={isLoading}
+          handleChangeTitle={handleChangeTitle}
+        />
+
+        {!!todos?.length && (
```
lerastarynets, line 228:
```suggestion
        {!!todos.length && (
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,51 @@
...
+  allCopmletedCheck,
+  handleToggleAllTodosCompleted,
+  handleSubmit,
+  todoTitle,
+  tempTodo,
+  setTodoTitle,
+}) => {
+  return (
+    <header className="todoapp__header">
+      {todos.length > 0 && (
```
lerastarynets, line 27:
```suggestion
      {!!todos.length && (
```

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit more

src/components/TempTodo/TempTodo.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  tempTodo: Todo;
+  isLoading: boolean;
+};
+
+export const TempTodo: React.FC<Props> = ({ tempTodo, isLoading }) => {
```
witflash, line 10: You can just name the prop `todo`. It's definitely a component of the Todo, so there is no need to name `tempTodo` this prop. It's about naming only. It's not an error.

witflash (approved):
Good work!

Pay attention to my (and Lera's) comments ([here](https://github.com/mate-academy/react_todo-app-with-api/pull/655/files)) for your future work.
You can fix some uncritical points to get additional practice. It's not necessary to make a request for the review again.

PR #662, opened 2023-07-26
todo with api

Ir-ra:
[DEMO LINK](https://Ir-ra.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,162 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+import * as todoServise from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
 
-const USER_ID = 0;
+const USER_ID = 11092;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isError, setIsError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todoStatus, setTodoStatus] = useState<Status>(Status.ALL);
+  const [loadingTodoIds, setLoadingTodoIds] = useState([0]);
+
+  useEffect(() => {
+    todoServise.getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        setIsError(true);
+        setErrorMessage('Unable to load todos');
+      });
+  }, []);
+
+  const addTodo = useCallback(async (title: string) => {
+    try {
+      const newTodo = {
+        title,
+        userId: USER_ID,
+        completed: false,
+      };
+
+      setTempTodo({ ...newTodo, id: 0 });
+
+      const createdTodo = await todoServise.createTodo(newTodo);
+
+      setTodos(currTodos => [...currTodos, createdTodo]);
+    } catch {
+      setIsError(true);
+      setErrorMessage('Unable to add a todo');
+    } finally {
+      setTempTodo(null);
+    }
+  }, [USER_ID, todos]);
+
+  const deleteTodo = async (todoId: number) => {
+    setLoadingTodoIds(prevIds => [...prevIds, todoId]);
+    await todoServise.delTodos(todoId)
+      .then(() => {
+        setTodos(currTodo => currTodo.filter(todo => todo.id !== todoId));
+      })
+      .catch((err) => {
+        setTodos(todos);
+        setErrorMessage('Unable to delete a todo');
+        throw err;
+      })
+      .finally(() => {
+        setLoadingTodoIds([0]);
+      });
+  };
+
+  const updateTodo = async (updatedTodo: Todo) => {
+    try {
+      setLoadingTodoIds(prevIds => [...prevIds, updatedTodo.id]);
+
+      await todoServise.updateTodo(updatedTodo);
+      setTodos(currTodos => {
+        const updTodos = [...currTodos];
+        const index = updTodos.findIndex(todo => todo.id === updatedTodo.id);
+
+        updTodos.splice(index, 1, updatedTodo);
+
+        return updTodos;
+      });
+    } catch {
+      setIsError(true);
+      setErrorMessage('Unable to update a todo');
+    } finally {
+      setLoadingTodoIds([0]);
+    }
+  };
+
+  const preparedTodos = useMemo(() => {
+    return todos.filter(todo => {
+      switch (todoStatus) {
+        case Status.ACTIVE:
+          return !todo.completed;
+
+        case Status.COMPLETED:
+          return todo.completed;
+
+        case Status.ALL:
+        default:
+          return todos;
+      }
+    });
+  }, [todoStatus, todos]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <Header
+          todos={todos}
+          addTodo={addTodo}
+          updateTodo={updateTodo}
+        />
+
+        <section className="todoapp__main">
+          <TodoList
+            preparedTodos={preparedTodos}
+            onDelete={deleteTodo}
+            tempTodo={tempTodo}
+            updateTodo={updateTodo}
+            loadingTodoIds={loadingTodoIds}
+          />
+        </section>
+
+        {!!todos.length && (
+          <Footer
+            todos={todos}
+            selectItem={todoStatus}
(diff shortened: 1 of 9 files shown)


src/api/todos.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function createTodo({ userId, title, completed }: Omit<Todo, 'id'>) {
+  return client.post<Todo>('/todos', { userId, title, completed });
+}
+
+export function updateTodo(todo: Todo) {
+  return client.patch<Todo>(`/todos/${todo.id}`, todo);
+}
+
+// export const updateTodo = (id: number, title: string) => {
```
maxim2310: Avoid leaving commented out code in your code base. If the code is not needed, it should be removed.

src/components/TodoList.tsx
```diff
@@ -0,0 +1,45 @@
...
+  tempTodo: Todo | null,
+  updateTodo: (uTodo: Todo) => void,
+  loadingTodoIds: number[]
+};
+
+export const TodoList: React.FC<Props> = ({
+  preparedTodos,
+  onDelete = () => { },
+  tempTodo,
+  updateTodo = () => { },
```
maxim2310, line 17: Default function for updateTodo prop is provided in the component definition. Similar to the onDelete prop, this is unnecessary and should be passed from the parent. Remove the default function definition.

src/components/Footer.tsx
```diff
@@ -0,0 +1,78 @@
...
+        {`${isActiveTodos.length} items left`}
+      </span>
+
+      <nav className="filter">
+        <a
+          href="#/"
+          className={classNames('filter__link', {
+            selected: selectItem === Status.ALL,
+          })}
+          onClick={() => setSelectItem(Status.ALL)}
```
maxim2310, line 41: You are not preventing the default click event which might cause the page to reload when the anchor tags are clicked. Call event.preventDefault() in the click handler.

src/components/Header.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const [value, setValue] = useState('');
+
+  const handleFormSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    addTodo(value);
+    setValue('');
+  };
+
+  const handleCompleteAll = (e: React.MouseEvent) => {
+    e.preventDefault();
```
maxim2310, line 23: You don’t need to call 'preventDefault' on a click event unless it’s a submit button or a link. It's redundant in this case.

src/components/Header.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const activeTodos = todos.filter(todo => !todo.completed);
+
+    if (isSomeActive) {
```
maxim2310, line 26: The part 'isSomeActive' is calculated at the beginning of the function and not updated later, so its value may not reflect the current state of todos when the 'handleCompleteAll' function is called.

src/components/Header.tsx
```diff
@@ -0,0 +1,64 @@
...
+      activeTodos.forEach(todo => {
+        const updatedTodo = { ...todo, completed: !todo.completed };
+
+        updateTodo(updatedTodo);
+      });
+    } else {
+      todos.forEach(todo => {
+        const updatedTodo = { ...todo, completed: !todo.completed };
```
maxim2310, line 34: Same issue as above, the status should be set to 'false' instead of '!todo.completed' because we are unchecking all todos.

src/components/Header.tsx
```diff
@@ -0,0 +1,64 @@
...
+    setValue('');
+  };
+
+  const handleCompleteAll = (e: React.MouseEvent) => {
+    e.preventDefault();
+    const activeTodos = todos.filter(todo => !todo.completed);
+
+    if (isSomeActive) {
+      activeTodos.forEach(todo => {
+        const updatedTodo = { ...todo, completed: !todo.completed };
```
maxim2310, line 28: You're updating the todo status to '!todo.completed' which is inverse of current status. But in 'handleCompleteAll', we want to mark all todos as completed, so it should be set to 'true'.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
maxim2310, line 1: Avoid using 'any' type unless it's absolutely necessary. Especially avoid disabling typescript rules for extensive use of 'any'. It may lead to runtime errors which are hard to debug. Try to replace 'any' with a more specific type.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // we wait for testing purpose to see loaders
+  return wait(300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
+      if (!response.ok) {
+        throw new Error();
```
maxim2310, line 34: It would be helpful to provide a meaningful message when throwing an error so it will be easier to debug. For instance, 'throw new Error('Failed to fetch data')'.

src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
+      const createdTodo = await todoServise.createTodo(newTodo);
+
+      setTodos(currTodos => [...currTodos, createdTodo]);
+    } catch {
+      setIsError(true);
+      setErrorMessage('Unable to add a todo');
+    } finally {
+      setTempTodo(null);
+    }
+  }, [USER_ID, todos]);
```
maxim2310, line 53: The 'todos' dependency is not necessary in the 'addTodo' useCallback hook, since it's not used within the function.

src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
+      .then(() => {
+        setTodos(currTodo => currTodo.filter(todo => todo.id !== todoId));
+      })
+      .catch((err) => {
+        setTodos(todos);
+        setErrorMessage('Unable to delete a todo');
+        throw err;
+      })
+      .finally(() => {
+        setLoadingTodoIds([0]);
```
maxim2310, line 67: LoadingTodoIds is reset to [0] after a todo is deleted, which may confuse users and is not a good practice. Consider setting it to an boolean

src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
+      switch (todoStatus) {
+        case Status.ACTIVE:
+          return !todo.completed;
+
+        case Status.COMPLETED:
+          return todo.completed;
+
+        case Status.ALL:
+        default:
+          return todos;
```
maxim2310, line 103: Returning 'todos' for default case in the filter is not correct. It will return multiple same lists in an array. Instead you should return 'true' to include all todos.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+        completed,
+      })}
+      key={id}
+    >
+      <label className="todo__status-label">
+        <input
+          type="checkbox"
+          className="todo__status"
+          defaultChecked={completed}
+          onChange={handlerChangeStatus}
```
maxim2310, line 67: The checkbox input field should use 'checked' instead of 'defaultChecked' to reflect the state changes correctly.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+  // const handleNewTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+  //   setNewTitle(event.target.value);
+  // };
+
+  return (
+    <div
+      className={classNames('todo', {
+        completed,
+      })}
+      key={id}
```
maxim2310, line 60: The 'key' prop is not necessary here, as it's only needed when rendering a list of children components, not within the component itself.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  const handleKeyUp = (event: React.KeyboardEvent) => {
+    if (event.key === 'Escape') {
+      setIsEdited(false);
+      setNewTitle(title);
+    }
+  };
+
+  // const handleNewTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
```
maxim2310: There is commented out code. If it's not used, it should be removed to keep the code clean.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+    } else if (newTitle === title) {
+      setIsEdited(false);
+    } else {
+      updateTodo(updatedTodo);
+      setIsEdited(false);
+    }
+  };
+
+  const handlerChangeStatus = (e: React.ChangeEvent) => {
+    e.preventDefault();
```
maxim2310, line 42: You don't need to preventDefault in the checkbox's onChange handler. It doesn't trigger a page reload.

maxim2310 (changes requested):
Good job, just make a small fixes

maxim2310 (changes requested):
If you will have questions you may ask it in slack or here in comment

src/utils/fetchClient.ts
Ir-ra: But this line was in start template. And  we can send any data to the server. I don't understand on what type it should be changed

src/App.tsx
```diff
@@ -1,24 +1,159 @@
...
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+import * as todoServise from './api/todos';
```
agniya-i, line 10:
```suggestion
import * as todoService from './api/todos';
```
Typo

src/api/todos.ts
```diff
@@ -0,0 +1,18 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+export function delTodos(todoId: number) {
```
agniya-i, line 8:
```suggestion
export function deleteTodos(todoId: number) {
```

agniya-i (approved):

PR #665, opened 2023-07-27
Develop

read-my-code:
[DEMO LINK](https://read-my-code.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,7 +2,7 @@
 import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { TodoList } from './components/TodoList';
-import { Todo } from './types/Todo';
+import { Todo, TodoData } from './types/Todo';
 import * as todosService from './api/todos';
 import { NotificationText, SortCondition } from './types/enums';
 import { Footer } from './components/Footer';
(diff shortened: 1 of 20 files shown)


src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+
+        return newTodos;
+      }
+
+      return currentTodos;
+    });
+
+    if (updateTodo.completed) {
+      if (todosFromServer
+        && completedItemsCount === todosFromServer?.length - 1
```
loralevitska, line 93: Long expression move to variable

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+
+    if (!todosFromServer) {
+      return;
+    }
+
+    if (activeItemsCount) {
+      itemsToToggle = todosFromServer?.filter(todo => !todo.completed);
+    } else {
+      itemsToToggle = todosFromServer?.filter(todo => todo.completed);
+    }
```
loralevitska, line 130:
```suggestion
    itemsToToggle = todosFromServer?.filter(todo => activeItemsCount ? !todo.completed : todo.completed);
```

loralevitska (changes requested):
Almost done) Let's improve

- remove outline, it breaks design
image

SanyaBratashchuk (changes requested):
Remove outline 
image

DarkMistyRoom (changes requested):
- there is a possibility to add empty todo with white spaces only. Trim value on validation

image

- consider making the "Clear complete" button invisible instead of removing it to keep filters in the center
image

Everything else is good for me

loralevitska (approved):
Good job!

PR #672, opened 2023-07-31
solution

Kinderwagen:
[DEMO LINK](https://Kinderwagen.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,391 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState, useMemo } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import * as postService from './api/todos';
+import { Todo } from './types/Todo';
 
-const USER_ID = 0;
+const USER_ID = 10876;
+
+enum Filter {
+  ALL = 'all',
+  ACTIVE = 'active',
+  COMPLETED = 'completed',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState<Filter>(Filter.ALL);
+  const [isShowError, setIsShowError] = useState<boolean>(false);
+  const [error, setError] = useState<string>('');
+  const [query, setQuery] = useState<string>('');
+  const [isTodoLoaded, setIsTodoLoaded] = useState<boolean>(false);
+
+  const [deleteTodoIds, setDeleteTodoIds] = useState([0]);
+  const [updatedTodoIds, setUpdatedTodoIds] = useState([0]);
+
+  const [isTodoEditing, setIsTodoEditing] = useState<boolean>(false);
+  const [todoEditing, setTodoEditing] = useState<Todo | null>(null);
+  const [newTitle, setNewTitle] = useState<string>('');
+
+  const handleErrorMessage = (message: string) => {
+    setIsShowError(true);
+    setError(message);
+
+    setTimeout(() => {
+      setIsShowError(false);
+    }, 3000);
+  };
+
+  useEffect(() => {
+    postService.getTodos(USER_ID)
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,24 +1,391 @@
...
+  ALL = 'all',
+  ACTIVE = 'active',
+  COMPLETED = 'completed',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState<Filter>(Filter.ALL);
+  const [isShowError, setIsShowError] = useState<boolean>(false);
+  const [error, setError] = useState<string>('');
```
VitaliyBondarenko1982, line 20: looks like redundant isShowError state
Enought error state and you can  reset this state when error shouldn't render

src/App.tsx
```diff
@@ -1,24 +1,391 @@
...
+  const [query, setQuery] = useState<string>('');
+  const [isTodoLoaded, setIsTodoLoaded] = useState<boolean>(false);
+
+  const [deleteTodoIds, setDeleteTodoIds] = useState([0]);
+  const [updatedTodoIds, setUpdatedTodoIds] = useState([0]);
```
VitaliyBondarenko1982, line 25: it can be one state. something like `processingTodoIds`

src/App.tsx
```diff
@@ -1,24 +1,391 @@
...
+        <header className="todoapp__header">
+          <button
+            type="button"
+            className={classNames('todoapp__toggle-all', {
+              active: activeTodos.length === 0,
+            })}
+            onClick={handleToggleAll}
+          />
+
+          <form onSubmit={(event) => addTodo(event, query)}>
```
VitaliyBondarenko1982, line 208: make handler in component and pass reference here

src/App.tsx
```diff
@@ -1,24 +1,391 @@
...
+                      type="checkbox"
+                      className="todo__status"
+                      checked={completed}
+                      onClick={() => checkTodo(todo)}
+                    />
+                  </label>
+
+                  {(isTodoEditing && todoEditing?.id === id) ? (
+                    <form onSubmit={
+                      (event) => changeTodoTitle(event, newTitle)
```
VitaliyBondarenko1982, line 243: same for anonymous arrow function in jsx
make handler and pass referense here
check all places with anonymous arrow functions in jsx

src/App.tsx
```diff
@@ -1,24 +1,391 @@
...
+            <nav className="filter">
+              <a
+                role="button"
+                href="#/"
+                className={classNames(
+                  'filter__link',
+                  'filter__link__all',
+                  { selected: filterType === Filter.ALL },
+                )}
+                onClick={() => handleFilterChange(Filter.ALL)}
```
VitaliyBondarenko1982, line 332: make one handler factory for all sort fields and just call in jsx wity proper value
```
const makeHandleFilterChange = (sortField: Filter) => () => handleFilterChange(sortField);
```
```suggestion
                onClick={(makeHandleFilterChange(Filter.ALL)}
```

VitaliyBondarenko1982 (changes requested):
Looks good.
Needs some improvements.
Check [working example](https://mate-academy.github.io/react_todo-app-with-api/)
- do not render toggle button in header, when no any todos
Screenshot from 2023-07-31 13-07-47
- during editing todo after press escape button should finish editing and todo should be without editing
- focus on input after start edit todo.

src/App.tsx
```diff
@@ -1,24 +1,359 @@
...
+        !allTodosCompleted && !todo.completed)) {
+        checkTodo(todo, !todo.completed);
+      }
+    });
+  };
+
+  const handleDoubleClick = (selectedTodo: Todo) => {
+    setIsTodoEditing(true);
+    setTodoEditing(selectedTodo);
+    setNewTitle(selectedTodo.title);
```
IvanFesenko, line 161: You should focus on input here

IvanFesenko (approved):
Good job.
If you wanna add this project to your portfolio - you will need to fix issue with focus, add authorization

PR #677, opened 2023-07-31
task done

pos1bl:
https://pos1bl.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -32,7 +32,7 @@ export const App: React.FC = () => {
   const completedTodos = todos.filter(todo => todo.completed);
 
   const clearError = useCallback(
-    debounce(setError, 3000),
+    debounce(() => setError(''), 3000),
     [],
   );
 
@@ -41,7 +41,7 @@ export const App: React.FC = () => {
       .then(setTodos)
       .catch(() => {
         setError('Unable to get todos');
-        clearError('');
+        clearError();
       });
   }, []);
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,24 +1,180 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import debounce from 'lodash.debounce';
```
mykhalenych, line 7: Screenshot 2023-07-31 at 17 54 06
you are using debounce but didn't install this package

src/App.tsx
```diff
@@ -1,24 +1,180 @@
...
+  );
+
+  useEffect(() => {
+    postService.getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        setError('Unable to get todos');
+        clearError('');
+      });
+  }, []);
```
mykhalenych, line 46: missed dependancy clearError

src/App.tsx
```diff
@@ -1,24 +1,180 @@
...
+    debounce(setError, 3000),
+    [],
+  );
+
+  useEffect(() => {
+    postService.getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        setError('Unable to get todos');
+        clearError('');
```
mykhalenych, line 44: and why do you need empty string as argument

mykhalenych (changes requested):
your demo link isn't working

src/App.tsx
pos1bl: i installed this package, but it doesnt work

PR #690, opened 2023-08-01
Task done

AntonBaranskyi:
https://antonbaranskyi.github.io/react_todo-app-with-api/

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,5 +8,6 @@ module.exports = {
     'jsx-a11y/label-has-associated-control': ["error", {
       assert: "either",
     }],
+    'import/no-cycle': 'off', 
   },
 };
--- a/src/components/AppBody/AppBody.tsx
+++ b/src/components/AppBody/AppBody.tsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import React, { useContext, useMemo } from 'react';
 import { TodosContext } from '../../context/TodosContext';
 import { TodoList } from '../TodoList';
 import { Footer } from '../Footer';
@@ -10,7 +10,9 @@ import { TempTodoItem } from '../TempTodoItem';
 export const AppBody: React.FC = () => {
   const { todos, sortField, tempTodo } = useContext(TodosContext);
 
-  const visibleTodos = filterTodo(todos, sortField);
+  const visibleTodos = useMemo(() => {
+    return filterTodo(todos, sortField);
+  }, [todos, sortField]);
 
   return (
     <div className="todoapp">
--- a/src/components/AppBody/AppBody.tsx
+++ b/src/components/AppBody/AppBody.tsx
@@ -1,0 +1,35 @@
+import React, { useContext } from 'react';
+import { TodosContext } from '../../context/TodosContext';
+import { TodoList } from '../TodoList';
+import { Footer } from '../Footer';
+import { ErrorComponent } from '../ErrorComponent';
+import { Header } from '../Header';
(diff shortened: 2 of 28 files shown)


.eslintrc.js
```diff
@@ -8,5 +8,6 @@ module.exports = {
     'jsx-a11y/label-has-associated-control': ["error", {
       assert: "either",
     }],
+    'import/no-cycle': 'off',
```
vlopatko, line 11: I think these changes are not good, cause you import files by cycle, and do not need to disable these eslint rules.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,49 @@
...
+    [todos],
+  );
+
+  const countFinished = useCallback(
+    (listTodo: Todo[]) => {
+      return listTodo.map((todo) => todo.completed).filter((todo) => todo)
+        .length;
+    },
+    [todos],
+  );
```
vlopatko, line 23: You can move these functions to a utils file or something like utils, or just out of the component, and useCallback will be do not need

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,142 @@
...
+        <div className="modal overlay">
+          <div className="modal-background has-background-white-ter" />
+          <div className="loader" />
+        </div>
+      )}
+
+      <div className="modal overlay">
+        <div className="modal-background has-background-white-ter" />
+        <div className="loader" />
+      </div>
```
vlopatko, line 139: You can use ClassNames library for editing classname , and you will not be needed to create 3 same elements.
Follow DRY rules

vlopatko (approved):
Good job! 👍🏻 Looks nice 💯 
But I have some advices for you, please check my comments.

src/components/AppBody/AppBody.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Footer } from '../Footer';
+import { ErrorComponent } from '../ErrorComponent';
+import { Header } from '../Header';
+import { filterTodo } from '../../helpers/filterTodo';
+import { TempTodoItem } from '../TempTodoItem';
+
+export const AppBody: React.FC = () => {
+  const { todos, sortField, tempTodo } = useContext(TodosContext);
+
+  const visibleTodos = filterTodo(todos, sortField);
```
anstsot, line 13: use `useMemo` hook

src/components/Footer/Footer.tsx
anstsot, line 23: better to use `useMemo`

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const Header: React.FC = () => {
+  const { toggleAll } = useContext(TodosContext);
+
+  return (
+    <header className="todoapp__header">
+      <button
+        type="button"
+        className="todoapp__toggle-all active"
+        onClick={() => toggleAll()}
```
anstsot, line 14:
```suggestion
        onClick={toggleAll}
```

anstsot (changes requested):
There are few `eslint` issues (fix them) and let's fix few bugs:
1. It's possible to change the todo to empty 
image

2. By clicking on toggle the loaders have all todos (even completed)
image

3. there is loader and error on remove (the loades shouldn't be visible on error)
image

4. There is an issue with focus on editing

anstsot (changes requested):
2 first bugs are still reproducible. check them, please
Fix `eslint` issue
image

PR #697, opened 2023-08-01
solution

ovynarchuk:
[DEMO LINK](https://ovynarchuk.github.io/react_todo-app-with-api/)


anstsot (approved):
Good work, approving:sparkles:
But try to fix one small issue:
- remove temp todo if there is an error by creating/deleting/
image

PR #702, opened 2023-08-01
Task solution

AnastasiiaKhemych:
[DEMO LINK](https://AnastasiiaKhemych.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoHeader.tsx
+++ b/src/components/TodoHeader.tsx
@@ -9,7 +9,8 @@ type Props = {
   tempTodo: Todo | null,
   setTempTodo: (value: null | Todo) => void,
   setHasError: (value: Error) => void,
-  setLoading: (value: boolean) => void,
+  loadingIds: number[],
+  setLoadingIds: React.Dispatch<React.SetStateAction<number[]>>,
 };
 
 export const TodoHeader: React.FC<Props> = ({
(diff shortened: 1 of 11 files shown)


src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,118 @@
...
+      try {
+        await todosService.updateTodo(todo.id, { completed: neededStatus });
+      } catch (err) {
+        setHasError(Error.UPDATE);
+      }
+    }));
+
+    setTodos(todos.map(t => ({
+      ...t,
+      completed: neededStatus,
```
ihorhladkov, line 51: Noticed an issue that the todos list is being modified both in case of an error and without it, but it should only change if the operation is successful. I think this case should be moved inside the try block.

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,118 @@
...
+    <header className="todoapp__header">
+      {todos.length > 0 && (
+        /* eslint-disable-next-line */
+        <button
+          type="button"
+          className={cn('todoapp__toggle-all', {
+            active: !neededStatus,
+          })}
+          onClick={toggleAllTodos}
+        />
```
ihorhladkov, line 103: I think it would be better to disable the button while the request is being processed, as it can be clicked multiple times in quick succession, resulting in multiple requests sent to the server simultaneously.

ihorhladkov (changes requested):

src/components/TodoHeader.tsx
ihorhladkov: Excellent work, but there are a few suggestions for improvement))

ihorhladkov:
Excellent work, but there are a few suggestions for improvement))

ihorhladkov (approved):
Great job. Approved)

KandaskalovOleg (approved):
Good job

And678 (approved):
Awesome :+1:

PR #722, opened 2023-08-06
solution

IzzetYusufov:
[DEMO LINK](https://IzzetYusufov.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoBody/TodoBody.tsx
+++ b/src/components/TodoBody/TodoBody.tsx
@@ -1,0 +1,154 @@
+import classNames from 'classnames';
+import { useEffect, useRef, useState } from 'react';
+import { Todo } from '../../types/Todo';
+import { Errors } from '../../types/Errors';
+
+type Props = {
+  filteringBy: Todo[],
+  newTodoId: number[],
+  tempTodo: Omit<Todo, 'id'> | null,
+  deleteTodo: (todoId: number) => void,
+  updateTodo: (updatedTodo: Todo) => Promise<void>,
+};
+
+export const TodoBody: React.FC<Props> = ({
+  filteringBy,
+  newTodoId,
+  tempTodo,
+  deleteTodo,
+  updateTodo,
+}) => {
+  const [isEditing, setIsEditing] = useState<number | null>(null);
+  const [editValue, setEditValue] = useState('');
+  const editFocus = useRef<HTMLInputElement | null>(null);
+
+  // Activate input OnDoubleclick to edit the todo
+  useEffect(() => {
+    if (editFocus.current) {
+      editFocus.current.focus();
+    }
+  }, [editValue]);
+
+  // Submit the edited todo
+  const handleEditingTodo = (
+    event: React.FormEvent<HTMLFormElement> | null,
+    todo: Todo,
+  ) => {
+    if (event) {
+      event.preventDefault();
+    }
+
+    switch (editValue) {
+      case Errors.NULL:
+        return deleteTodo(todo.id);
+      case todo.title:
+        return setIsEditing(null);
+      default:
+        return updateTodo({ ...todo, title: editValue })
+          .then(() => {
+            setIsEditing(null);
+            setEditValue('');
+          });
+    }
+  };
+
+  // Reset the edited todo on KeyUp "Escape"
+  const resetEditing = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.code === 'Escape') {
+      setIsEditing(null);
+      setEditValue('');
+    }
+  };
+
+  return (
+    <section className="todoapp__main">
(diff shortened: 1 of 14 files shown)


src/types/Errors.ts
```diff
@@ -0,0 +1,8 @@
+export enum Errors {
+  NO_TITLE = 'Title can\'t be empty',
+  ADD = 'Unable to add a todo',
+  DELETE = 'Unable to delete a todo',
+  UPDATE = 'Unable to update a todo',
+  LOAD = 'Unable to load a todos',
```
volodymyr-soltys97, line 6: Typo in the error message. It should be 'Unable to load todos' instead of 'Unable to load a todos'.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
volodymyr-soltys97, line 1: Avoid using `eslint-disable` for `@typescript-eslint/no-explicit-any`. Instead, try to avoid `any` type. It's better to use a more specific type. You can define a type or interface for the data.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // we wait for testing purpose to see loaders
+  return wait(300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
+      if (!response.ok) {
+        throw new Error();
```
volodymyr-soltys97, line 34: In case of error, provide an error message to give more context about the error. For instance, you can use `response.statusText` to get more detail about the error.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (data) {
+    // We add body and Content-Type only for the requests with data
+    options.body = JSON.stringify(data);
+    options.headers = {
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // we wait for testing purpose to see loaders
+  return wait(300)
```
volodymyr-soltys97, line 30: Consider removing the hardcoded delay used for testing purposes. This could negatively impact the performance of your application in production.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+}
+
+// To have autocompletion and avoid mistypes
+type RequestMethod = 'GET' | 'POST' | 'PATCH' | 'DELETE';
+
+function request<T>(
+  url: string,
+  method: RequestMethod = 'GET',
+  data: any = null, // we can send any data to the server
```
volodymyr-soltys97, line 17: Instead of using `any` for the `data` parameter, it would be better to use a generic type.

src/components/TodoBody/TodoBody.tsx
```diff
@@ -0,0 +1,154 @@
...
+          .then(() => {
+            setIsEditing(null);
+            setEditValue('');
+          });
+    }
+  };
+
+  // Reset the edited todo on KeyUp "Escape"
+  const resetEditing = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.code === 'Escape') {
```
volodymyr-soltys97, line 57: Consider using event.key === 'Escape' instead of event.code === 'Escape' for better browser compatibility.

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,96 @@
...
+    if (value === filterBy) {
+      return;
+    }
+
+    setFilterBy(value);
+  };
+
+  // delete all complied todos
+  const deleteComplitedTodos = () => {
+    todos.map(todo => todo.completed && deleteTodo(todo.id));
```
volodymyr-soltys97, line 36: Instead of using map() for side effects, it is better to use forEach(). map() is generally used for transforming data, not for side effects like deleting todos.

src/components/TodoBody/TodoBody.tsx
```diff
@@ -0,0 +1,154 @@
...
+      setIsEditing(null);
+      setEditValue('');
+    }
+  };
+
+  return (
+    <section className="todoapp__main">
+      {/* Todos from server response */}
+      {filteringBy.map(todo => (
+        <div
```
volodymyr-soltys97, line 67: Use a destructuring for todo

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):

volodymyr-soltys97 (approved):

PR #733, opened 2023-08-10
React__todo-app-with-api

Soi4An:
DEMO - https://Soi4An.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,175 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { UserWarning } from './UserWarning';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Filter } from './components/Filter';
+import { Notifications } from './components/Notifications';
+import {
+  addTodo, deleteTodo, getTodos, updateTodo,
+} from './api/todos';
 
-const USER_ID = 0;
+import { Todo } from './types/Todo';
+import { ErrorText } from './types/ErrorText';
+import { Filters } from './types/Filters';
+
+const USER_ID = 11213;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loadedTodos, setLoadingTodos] = useState<number[]>([]);
+  const [errorMessage, setErrorMessage] = useState(ErrorText.Empty);
+  const [selectedFilter, setSelectedFilter] = useState<Filters>(Filters.All);
+  const [title, setTitle] = useState('');
+  const [isDisabledInput, setIsDisabledInput] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isFocusedHeader, setIsFocusedHeader] = useState(false);
+
+  const filteredTodos = (filterChar: Filters) => {
+    return todos.filter(todo => {
+      switch (filterChar) {
+        case Filters.All:
+          return true;
+
+        case Filters.Active:
+          return todo.completed === false;
+
+        case Filters.Completed:
+          return todo.completed === true;
+
+        default:
+          return false;
+      }
+    });
+  };
+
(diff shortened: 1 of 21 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const handlerCreatedTitle = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = e.target;
+
+    onSetTitle(value);
+  };
+
+  const handlerSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (title.length < 1) {
```
volodymyr-soltys97, line 55: Use trim() on the title to prevent creation of todos with blank spaces. It should be if (title.trim().length < 1)

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,96 @@
...
+    onAddTodos();
+  };
+
+  const editedCompletedInput = useRef<HTMLInputElement | null>(null);
+
+  useEffect(() => {
+    if (editedCompletedInput.current !== null && onIsFocusedHeader) {
+      editedCompletedInput.current.focus();
+    }
+  }, [todos]);
```
volodymyr-soltys97, line 70: The dependency array in the 'useEffect' hook should also include 'onIsFocusedHeader' since it's being used in the hook.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,42 @@
...
+    options.headers = {
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  return wait(300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
+      if (!response.ok) {
+        throw new Error();
```
volodymyr-soltys97, line 30: Throwing a new Error without a message is not recommended. It's better to provide some details about the error to make debugging easier.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,42 @@
...
+      }
+
+      return response.json();
+    });
+}
+
+export const client = {
+  get: <T>(url: string) => request<T>(url),
+  post: <T>(url: string, data: any) => request<T>(url, 'POST', data),
```
volodymyr-soltys97, line 39: Replace 'any' with a more specific type or consider using generics.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,42 @@
...
+  const options: RequestInit = { method };
+
+  if (data) {
+    options.body = JSON.stringify(data);
+    options.headers = {
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  return wait(300)
```
volodymyr-soltys97, line 26: The function 'wait' is not necessary for the fetch API to work properly. Unless there's a specific reason for this delay, you might consider removing it to improve performance.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
volodymyr-soltys97, line 1: Avoid using 'any' type. It's better to replace it with a more specific type or use generics when possible.

src/App.tsx
```diff
@@ -1,24 +1,175 @@
...
+
+const USER_ID = 11213;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loadedTodos, setLoadingTodos] = useState<number[]>([]);
+  const [errorMessage, setErrorMessage] = useState(ErrorText.Empty);
+  const [selectedFilter, setSelectedFilter] = useState<Filters>(Filters.All);
+  const [title, setTitle] = useState('');
+  const [isDisabledInput, setIsDisabledInput] = useState(false);
```
volodymyr-soltys97, line 25: Add a generic types everywhere for state

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #735, opened 2023-08-12
solution

artem-ivashchenko:
[DEMO LINK](https://artem-ivashchenko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoHeader.tsx
+++ b/src/components/TodoHeader.tsx
@@ -1,0 +1,86 @@
+import classNames from 'classnames';
+import React, { useCallback, useMemo, useState } from 'react';
+import { Todo } from '../types/Todo';
+import { ErrorMessage } from '../types/Error';
+import { USER_ID } from '../utils/USER_ID';
+
+type Props = {
+  todos: Todo[],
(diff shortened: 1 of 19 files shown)


src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+  onError: (error: ErrorMessage) => void;
+};
+
+export const TodoHeader: React.FC<Props> = ({
+  todos, onUpdate, onAdd, onError,
+}) => {
+  const [newTitle, setNewTitle] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  const isSomeVisible = useMemo(() => !!todos.length, [todos]);
```
lerastarynets, line 20: here useMemo is not really necessary
it can be beneficial if the todos array is large, but i doubt that this situation is possible in the project)
no need to fix this, just fyi

lerastarynets (approved):
You're doing brilliantly, keep it up!
Really advanced job, cool that you used CSSTransiton for the smooth animation, also good that you store filters in separate const 🚀

PR #736, opened 2023-08-12
The task is completed

AndruhaMan:
[DEMO LINK](https://AndruhaMan.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,21 +19,19 @@ const USER_ID = 11276;
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [isError, setIsError] = useState<boolean>(false);
   const [errorMessage, setErrorMessage] = useState<ErrorMessage>('');
   const [filterStatus, setFilterStatus] = useState(FilterStatus.All);
   const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
 
-  if (isError) {
-    setTimeout(() => setIsError(false), 3000);
+  if (errorMessage) {
+    setTimeout(() => setErrorMessage(''), 3000);
   }
 
   useEffect(() => {
     getTodos(USER_ID)
       .then(setTodos)
       .catch(() => {
-        setIsError(true);
         setErrorMessage('Unable to fetch todos');
       });
   }, []);
@@ -68,9 +66,8 @@ export const App: React.FC = () => {
       const newTodo = await postTodo(preparedTodo);
 
       setTodos([...todos, newTodo]);
-    } catch (e) {
+    } catch {
       setErrorMessage('Unable to add a todo');
-      setIsError(true);
     } finally {
       setTempTodo(null);
     }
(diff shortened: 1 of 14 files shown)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,113 @@
...
+              onClick={() => removeTodo(id)}
+            >
+              ×
+            </button>
+          </>
+        )
+        : (
+          <form onSubmit={handleFormSubmit}>
+            <input
+              type="text"
```
TarasHoliuk, line 86: add such placeholder
image

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,113 @@
...
+              className="todo__title-field"
+              ref={editInput}
+              value={editedTitle}
+              onChange={(event) => setEditedTitle(event.target.value)}
+              onBlur={handleFormSubmit}
+              onKeyUp={(event) => {
+                if (event.key === 'Escape') {
+                  setIsEditing(false);
+                }
+              }}
```
TarasHoliuk, line 96: let's move it to a separate function and use it here

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  const handleFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    const trimedTitle = editedTitle.trim();
+
+    if (!trimedTitle) {
+      removeTodo(id);
+    } else {
+      updateTodo({ ...todo, title: trimedTitle });
```
TarasHoliuk, line 42: you shouldn't send a request if nothing was changed

src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { getCompletedTodos } from '../../services/todo';
+import { Todo } from '../../types/Todo';
+import { ErrorMessage } from '../../types/ErrorMessage';
+
+type Props = {
+  todos: Todo[],
+  setIsError: (status: boolean) => void,
+  setErrorMessage: (message: ErrorMessage) => void
+  loadingTodoIds: number[],
+  setLoadingTodoIds: (ids: number[]) => void,
```
TarasHoliuk, line 12: don't pass it and remove from props - you don't use it

src/App.tsx
```diff
@@ -1,24 +1,167 @@
...
+import { Todo } from './types/Todo';
+import { FilterStatus } from './types/FilterStatus';
+import { ErrorMessage } from './types/ErrorMessage';
 
-const USER_ID = 0;
+const USER_ID = 11276;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isError, setIsError] = useState<boolean>(false);
```
TarasHoliuk, line 22: you may get rid of this state
an empty string in errorMessage is a falsy value
so when you want to hide an error - just set an empty string as errorMessage

then the code below will look like this:
```
if (errorMessage) {
    setTimeout(() => setErrorMessage(''), 3000);
  }
```

src/App.tsx
```diff
@@ -1,24 +1,167 @@
...
+      userId: USER_ID,
+    };
+
+    setTempTodo({ ...preparedTodo, id: 0 });
+
+    try {
+      const newTodo = await postTodo(preparedTodo);
+
+      setTodos([...todos, newTodo]);
+    } catch (e) {
```
TarasHoliuk, line 71:
```suggestion
    } catch {
```
you don't use it anyway

TarasHoliuk (changes requested):
Nicely done :+1: 
Let's improve your code a bit to make it even better :)

anastasiiavorobiova (approved):
Great work!

PR #737, opened 2023-08-12
react_todo-app-with-api

vintwp:
https://vintwp.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,18 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useEffect, useState } from 'react';
+import cn from 'classnames';
+
 import { UserWarning } from './UserWarning';
 import { TodoHeader } from './components/TodoHeader';
-import { client } from './utils/fetchClient';
-import { Todo } from './types/Todo';
 import { TodoList } from './components/TodoList';
 import { TodoFooter } from './components/TodoFooter';
 import { Notification } from './components/Notification';
+
+import { Todo } from './types/Todo';
 import { FilterBy } from './types/FilterBy';
 import { Errors } from './types/Errors';
 
+import { client } from './utils/fetchClient';
 import * as utils from './utils/utils';
 
 const USER_ID = 11246;
@@ -202,7 +205,12 @@ export const App: React.FC = () => {
   }
 
   return (
-    <div className="todoapp">
+    <div
+      className={cn(
+        'todoapp',
+        { 'has-error': error },
+      )}
+    >
       <h1 className="todoapp__title">todos</h1>
       <div className="todoapp__content">
         <TodoHeader
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,245 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { TodoHeader } from './components/TodoHeader';
+import { client } from './utils/fetchClient';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { TodoFooter } from './components/TodoFooter';
+import { Notification } from './components/Notification';
+import { FilterBy } from './types/FilterBy';
+import { Errors } from './types/Errors';
 
-const USER_ID = 0;
+import * as utils from './utils/utils';
+
+const USER_ID = 11246;
+const BASE_ADD_URL = '/todos';
+const URL_GET = `?userId=${USER_ID}`;
+const DELAY_ERROR = 3000;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [hasTodoLoaderIds, setHasTodoLoaderIds] = useState<number[]>([]);
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.all);
+  const [error, setError] = useState<Errors | null>(null);
+
+  const addLoader = (id: number) => {
+    setHasTodoLoaderIds(currentIds => [...currentIds, id]);
+  };
+
+  const deleteLoader = (id: number) => {
+    setHasTodoLoaderIds(currentIds => [...currentIds].filter(
+      hasLoaderId => hasLoaderId !== id,
+    ));
+  };
+
+  const changeAndToggleTodo = (
+    id: number,
+    field: keyof Todo,
+    value: Todo[typeof field],
+  ) => {
+    const indexOfMutableTodo = todos.findIndex(todo => todo.id === id);
+
+    switch (field) {
+      case 'completed':
+        setTodos(currentTodos => {
+          const copyTodos = [...currentTodos];
+
+          copyTodos[indexOfMutableTodo].completed
+            = !value;
+
+          return copyTodos;
+        });
+        break;
+      case 'title':
+        setTodos(currentTodos => {
+          const copyTodos = [...currentTodos];
+
+          copyTodos[indexOfMutableTodo].title
+            = `${value}`;
+
+          return copyTodos;
+        });
+        break;
+      default:
+        break;
+    }
+  };
+
+  const applyFilter = (filter: FilterBy) => {
+    setFilterBy(filter);
+  };
+
+  const addTodo = async (title: string) => {
+    if (title) {
+      const newTodo = {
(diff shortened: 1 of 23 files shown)


src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+        break;
+    }
+  };
+
+  const applyFilter = (filter: FilterBy) => {
+    setFilterBy(filter);
+  };
+
+  const addTodo = async (title: string) => {
+    if (title) {
```
TarasHoliuk, line 78: check trimmed value to avoid adding Todo with spaces only:
image

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+      try {
+        const addedTodo = await client.post<Todo>(BASE_ADD_URL, newTodo);
+
+        setTodos(currentTodos => [
+          ...currentTodos,
+          addedTodo,
+        ]);
+      } catch (caughtError) {
+        setError(Errors.add);
+        throw caughtError;
```
TarasHoliuk, line 100: do not throw it again - it will cause a page crash (this error will not be handled somewhere)

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+      });
+
+      try {
+        const addedTodo = await client.post<Todo>(BASE_ADD_URL, newTodo);
+
+        setTodos(currentTodos => [
+          ...currentTodos,
+          addedTodo,
+        ]);
+      } catch (caughtError) {
```
TarasHoliuk, line 98:
```suggestion
      } catch {
```
you may not use it at all

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+      };
+
+      setHasTodoLoaderIds([0]);
+      setTempTodo({
+        ...newTodo,
+        id: 0,
+      });
+
+      try {
+        const addedTodo = await client.post<Todo>(BASE_ADD_URL, newTodo);
```
TarasHoliuk, line 92: create a separate function for it in the api/todos.ts
`client.post<Todo>(BASE_ADD_URL, newTodo);`

do the same for other actions
and move consts related to it

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+  const deleteTodo = async (id: number) => {
+    addLoader(id);
+
+    try {
+      await client.delete(`${BASE_ADD_URL}/${id}`);
+      setTodos(currentTodos => {
+        return [...currentTodos].filter(todo => todo.id !== id);
+      });
+    } catch (errorCought) {
+      if (!error) {
```
TarasHoliuk, line 119: i guess you may get rid of this check

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+
+const USER_ID = 11246;
+const BASE_ADD_URL = '/todos';
+const URL_GET = `?userId=${USER_ID}`;
+const DELAY_ERROR = 3000;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [hasTodoLoaderIds, setHasTodoLoaderIds] = useState<number[]>([]);
```
TarasHoliuk, line 26:
```suggestion
  const [todoLoaderIds, setTodoLoaderIds] = useState<number[]>([]);
```
use `has`, `is`, `can` prefixes for boolean values: [article](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig)

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+
+          return copyTodos;
+        });
+        break;
+      default:
+        break;
+    }
+  };
+
+  const applyFilter = (filter: FilterBy) => {
```
TarasHoliuk, line 73: this wrapper is redundant - everywhere you use it (e.g. pass into child component as prop) you may just use `setFilterBy`

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+    }
+  });
+
+  const completedTodos = utils.getCompletedTodos(todos);
+  const activeTodos = utils.getActiveTodos(todos);
+  const areAllTodosCompleted = activeTodos.length === 0;
+  const areThereTodos = todos.length > 0;
+
+  useEffect(() => {
+    client.get<Todo[]>(BASE_ADD_URL + URL_GET)
```
TarasHoliuk, line 192: use `getTodos` form api/todos.ts

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+      .then(dTodos => setTodos(dTodos))
```
TarasHoliuk, line 193:
```suggestion
      .then(setTodos)
```

src/App.tsx
```diff
@@ -1,24 +1,253 @@
...
+        {todos.length > 0 && (
+          <>
+            <section className="todoapp__main">
+              <TodoList
+                todos={todosToRender}
+                tempTodo={tempTodo}
+                isActiveLoaderTodos={hasTodoLoaderIds}
+                onDeleteTodo={id => deleteTodo(id)}
+                onToggleTodo={(id, completed) => toggleTodo(id, completed)}
+                onChangeTodo={(id, title) => changeTodo(id, title)}
```
TarasHoliuk, line 232:
```suggestion
                onDeleteTodo={deleteTodo}
                onToggleTodo={toggleTodo}
                onChangeTodo={changeTodo}
```

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,60 @@
...
+};
+
+export const TodoFooter: React.FC<Props> = ({
+  onChangeFilter,
+  filterSelected,
+  activeTodos,
+  completedTodos,
+  clearCompleted,
+}) => {
+  const filterButtons = Object.keys(FilterBy);
```
TarasHoliuk, line 20:
```suggestion
  const filterButtons = Object.values(FilterBy);
```

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <nav className="filter">
+        {filterButtons.map(button => (
+          <a
+            key={button}
+            href={`#/${button === FilterBy.all ? '' : `${button}`}`}
+            className={cn(
+              'filter__link',
+              { selected: button === filterSelected },
+            )}
+            onClick={() => handleClickFilter(button)}
```
TarasHoliuk, line 44:
```suggestion
            onClick={() => onChangeFilter(button)}
```
if you will map values of FilterBy enum ts should recognize them

src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,88 @@
...
+            'todoapp__toggle-all',
+            { active: isCheckedToggler },
+          )}
+          aria-label="Close"
+          onClick={handlerToggleAll}
+        />
+      )}
+
+      <form>
+        <input
```
TarasHoliuk, line 78: make this input controlled (save value in the state)
and use ref (InputHTMLElement) only to set focus

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const { id, title, completed } = todo;
+  const [isEditingTodo, setIsEditingTodo] = useState<boolean>(false);
+  const editingTodoField = useRef<HTMLInputElement | null>(null);
+
+  const changeTitle = async (idTodo: number, titleTodo: string) => {
+    try {
+      await onChangeTodo(idTodo, titleTodo);
+      setIsEditingTodo(false);
+    } catch (e) {
+      editingTodoField.current?.focus();
```
TarasHoliuk, line 31: why focus on error? maybe it's better to show some error, WDYT?

TarasHoliuk (changes requested):

src/App.tsx
vintwp: Hello. Thanks for your comment! This error handles in TodoHeader, and input doesn't clear after mistake (only if OK)

src/App.tsx
vintwp: I've thought, it helps avoid to additional render (if we add or remove multiple todos) - we set error once

src/components/TodoItem/TodoItem.tsx
vintwp: Fixed it.  I've made as in original

src/App.tsx
vintwp: It prevent multiple adding errors to state

volodymyr-soltys97 (approved):

PR #738, opened 2023-08-13
last part added

AndriiYelieva:
[DEMO LINK](https://AndriiYelieva.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoApp.tsx
+++ b/src/components/TodoApp.tsx
@@ -1,0 +1,115 @@
+import { useEffect, useState } from 'react';
+
+import { Todo } from '../types/Todo';
+import { Notification } from '../types/Notification';
+import { Status } from '../types/Status';
+
+import * as TodosService from '../api/todos';
+
+import { UserWarning } from '../UserWarning';
+import { TodoFooter } from './TodoFoter';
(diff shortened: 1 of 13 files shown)


src/components/TodoApp.tsx
```diff
@@ -0,0 +1,115 @@
...
+      })
+      .catch((error) => {
+        setErrorMessage(Notification.add);
+        throw error;
+      });
+  };
+
+  const deleteTodo = (todoId: number) => {
+    TodosService.deleteTodo(todoId);
+    setTodos(currentTodo => currentTodo.filter(todo => todo.id !== todoId));
```
VitaliyBondarenko1982, line 54: change state only after receive response from server amd handle error in catch
check other similer cases.

```suggestion
    TodosService.deleteTodo(todoId).then(() => {
         setTodos(currentTodo => currentTodo.filter(todo => todo.id !== todoId));
    }).catch( // handle error);
```

VitaliyBondarenko1982 (changes requested):
Some functionality not implemented:

- Toggle button in header
- editing title of todo
- loader for processing todo.

Please make re-request after finish all points from task description.

PR #740, opened 2023-08-15
Develop

diana-tuz:
[DEMO LINK](https://diana-tuz.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoApp/TodoApp.tsx
+++ b/src/components/TodoApp/TodoApp.tsx
@@ -1,0 +1,100 @@
+import React, { useState } from 'react';
+import { client } from '../../utils/fetchClient';
+import { URL, USER_ID } from '../../utils/Url';
+import { Todo } from '../../types/Todo';
+import { TodoList } from '../TodoList/TodoList';
+
+type Props = {
+  setTodos: React.Dispatch<React.SetStateAction<Todo[]>>
+  setErrorMessage: (a: string) => void;
+  todos: Todo[];
+};
+
+export const TodoApp: React.FC<Props> = ({
+  todos,
+  setTodos,
+  setErrorMessage,
+}) => {
+  const [value, setValue] = useState<string>('');
+
+  const handleAddTodo = (event: React.ChangeEvent<HTMLInputElement>) => {
+    event.preventDefault();
+    const newValue: string = event.target.value;
+
+    setValue(newValue);
+  };
+
+  const addNewTodo = async () => {
+    if (!value.trim()) {
+      return;
+    }
+
+    try {
+      const newTodoData = {
+        title: value,
+        completed: false,
+        userId: USER_ID,
+        id: 0,
+      };
+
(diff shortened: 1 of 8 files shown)


src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,146 @@
...
+        setTodos(updatedTodos);
+      } else {
+        const updatedTodos = await Promise.all(allTodos.map(todo => {
+          const completed = !todo.completed;
+
+          return { ...todo, completed };
+        }));
+
+        setAllTodos(updatedTodos);
+        setIsToggle(true);
```
anastasiiavorobiova, line 96: You are setting `setIsToggle` to true regardless of the if condition. It should be set to the opposite of its current state.

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,146 @@
...
+    if (!value.trim()) {
+      return;
+    }
+
+    try {
+      const newTodoData = {
+        title: value,
+        completed: false,
+        userId: USER_ID,
+        id: 0,
```
anastasiiavorobiova, line 49: The id of the new todo is set to 0. It would be better to generate a unique id for each new todo.

src/components/TodosFiltered/TodosFiltered.tsx
```diff
@@ -0,0 +1,141 @@
...
+    setTodos(allTodos);
+  }, [allTodos]);
+
+  const handleSetFilteredTodos
+    = (filter: Status) => {
+      setFilterBy(filter);
+      getFilteredTodos(filter);
+    };
+
+  const completedTodoLength
```
anastasiiavorobiova, line 76: The completedTodoLength and uncompletedTodos are calculated every time the component re-renders. This is unnecessary computation if todos does not change. Consider using useMemo hook for these computations.

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,147 @@
...
+                className="todo__edit-todo"
+                type="text"
+                value={editingTitle}
+                onChange={event => setEditingTitle(event.target.value)}
+                onKeyDown={event => {
+                  if (event.key === 'Enter') {
+                    handleSaveEdit(id);
+                  }
+                }}
+                onBlur={() => setEditingId(NaN)}
```
anastasiiavorobiova, line 110: You are setting the editingId to NaN when the input loses focus. This could potentially cause issues. It would be better to set it to null or undefined.

anastasiiavorobiova (changes requested):
Great work! The filtering buttons don't work on the demo page, consider fixing it

anastasiiavorobiova (changes requested):
Good work! Almost done

anastasiiavorobiova (approved):
Good work!

PR #747, opened 2023-08-17
solution

vlkzmn:
[DEMO LINK](https://vlkzmn.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,145 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useState, useEffect, useMemo } from 'react';
+import classNames from 'classnames';
+import { AddTodo } from './сomponents/AddTodo';
+import { TodosFilter } from './сomponents/TodosFilter';
+import { TodoList } from './сomponents/TodoList';
+import * as todoService from './api/todos';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterStatus, setFilterStatus] = useState<Filter>(Filter.all);
+  const [todosIdInLoading, setTodosIdInLoading] = useState<number[]>([]);
+
+  const showError = (message: string) => {
+    setErrorMessage(message);
+    setTimeout(() => setErrorMessage(''), 3000);
+  };
+
+  useEffect(() => {
+    todoService.getTodos()
+      .then(setTodos)
+      .catch(() => showError('Unable to load a todos'));
+  }, []);
+
+  const createTodo = (newTitle: string) => {
+    setTempTodo({
+      id: 0,
+      userId: 11337,
+      title: newTitle,
+      completed: false,
+    });
+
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,24 +1,145 @@
...
+  useEffect(() => {
+    todoService.getTodos()
+      .then(setTodos)
+      .catch(() => showError('Unable to load a todos'));
+  }, []);
+
+  const createTodo = (newTitle: string) => {
+    setTempTodo({
+      id: 0,
+      userId: 11337,
```
maxim2310, line 32: why do you need to set hardcode user id and id

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

about loader its must be default behavior you cant change it easy

src/App.tsx
vlkzmn: about userId it's ok
but how not hardcode set id = 0?

src/сomponents/TodoItem.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+      <div className={classNames(
+        'modal overlay',
+        { 'is-active': loading },
+      )}
+      >
+        <div className="modal-background has-background-white-ter" />
+        <div className="Loader">
+          <div className="Loader__content" />
+        </div>
```
Anastasiia-Svintsova, line 126: Styles for the loader are already implemented 
```suggestion
        <div className="loader" />
```

src/сomponents/TodosFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+      </span>
+
+      <nav className="filter">
+        <a
+          href="#/"
+          className={cn('filter__link', { selected: status === Filter.all })}
+          onClick={() => onChangeStatus(Filter.all)}
+        >
+          All
+        </a>
```
Anastasiia-Svintsova, line 40: You can create a separate component for the filter link to not repeat the same code for each link

Anastasiia-Svintsova (changes requested):
Almost done! Check my comments, please

src/App.tsx
```diff
@@ -1,24 +1,151 @@
...
+      })
+      .catch(() => {
+        showError('Unable to delete a todos');
+      })
+      .finally(() => setTodosIdInLoading(
+        current => current.filter(todoId => todoId !== id),
+      ));
+  };
+
+  useMemo(() => {
```
witflash, line 89: Here should be used `useEffect` hook instead of `useMemo`. The last one needs to store some value. But if you have only side-effects, you should use `useEffect`.
Your solution is workable but it's not in React Docs conventions.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
+function wait(delay: number) {
+  return new Promise(resolve => {
+    setTimeout(resolve, delay);
+  });
+}
+
+// To have autocompletion and avoid mistypes
```
witflash, line 11: You don't need all these comments. Your code is self-described.

src/сomponents/AddTodo.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const [title, setTitle] = useState('');
+  const [disableInput, setDisableInput] = useState(false);
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    setDisableInput(true);
+    const newTitle = title.trim();
+
+    if (!newTitle) {
+      setError('Title can\'t be empty');
```
witflash, line 29: Here is small hint. You can use the double quotes for such situations:
```
"Title can't be empty"
```
It looks much better as for me.

witflash (approved):
Well done 👍🏻 
I've left several uncritical comments. Check them for you future work.

src/App.tsx
vlkzmn: Previous mentor recomended useMemo )))

PR #749, opened 2023-08-20
v1

victor-buldenko:
[DEMO LINK](https://victor-buldenko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,213 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState } from 'react';
+import { CSSTransition, TransitionGroup } from 'react-transition-group';
+import { Etodos, ResponseError } from './types/enum';
+import { UserData } from './types/userData';
+import { Todo } from './types/Todo';
+import {
+  createTodo,
+  deleteTodoOnServer,
+  getTodos,
+  updateTodoProp,
+} from './api';
+import { TodoItem } from './components/Todo';
+import { TodoHeader } from './components/TodoHeader';
+import { Footer } from './components/Footer';
+import { Notification } from './components/Notification';
 
-const USER_ID = 0;
+const user: UserData = {
+  createdAt: '2023-07-10T13:09:53.578Z',
+  email: 'gookidoo@gmail.com',
+  id: 11038,
+  name: 'Віктор Булденко',
+  phone: null,
+  updatedAt: '2023-07-10T13:09:53.578Z',
+  username: null,
+  website: null,
+};
+
+let isUncomplete = 0;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isShowFooter, setIsShowFooter] = useState<boolean>(true);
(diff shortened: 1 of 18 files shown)


src/App.tsx
```diff
@@ -1,24 +1,213 @@
...
+  email: 'gookidoo@gmail.com',
+  id: 11038,
+  name: 'Віктор Булденко',
+  phone: null,
+  updatedAt: '2023-07-10T13:09:53.578Z',
+  username: null,
+  website: null,
+};
+
+let isUncomplete = 0;
```
ivbutsykin, line 28: Don't use such variables. Use variables inside component (or state if need)

src/App.tsx
```diff
@@ -1,24 +1,213 @@
...
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isShowFooter, setIsShowFooter] = useState<boolean>(true);
```
ivbutsykin, line 32: No need in state. You can simple create variable inside component and use it:
```
const hasFooter = !!todos.length
```

src/App.tsx
```diff
@@ -1,24 +1,213 @@
...
+  };
+
+  const checkCompletedTodo = (arr: Todo[]) => {
+    isUncomplete = 0;
+    for (let i = 0; i < arr.length; i += 1) {
+      if (!arr[i].completed) {
+        isUncomplete += 1;
+      }
+    }
+  };
```
ivbutsykin, line 68: Rely on `todos` (rewrite it as need, but avoid such loops and changeable  variables outside):
```suggestion
   const completedTodos = todos.filter(todo => todo.completed);
   const completedTodosLength = completedTodos.length;
```

src/App.tsx
```diff
@@ -1,24 +1,213 @@
...
+
+  const deleteTodo = (id: number) => {
+    setIsLoading(true);
+
+    return deleteTodoOnServer(id)
+      .then(() => {
+        getTodos(user.id).then((todoList) => {
+          setTodos(todoList);
+          checkCompletedTodo(todoList);
+          setIsShowFooter(Boolean(todoList.length));
```
ivbutsykin, line 78: Mostly `!!` uses to convert in boolean type
```suggestion
          setIsShowFooter(!!(todoList.length));
```

ivbutsykin (changes requested):
Please, fix issues below and also add loader spinners for temp todo, toggling todo (see example: https://mate-academy.github.io/react_todo-app-with-api/).

Current result:
https://github.com/mate-academy/react_todo-app-with-api/assets/42767822/6e0a1908-cf5d-43fe-9e22-36dd4554851a

Expected result:

https://github.com/mate-academy/react_todo-app-with-api/assets/42767822/0c081f77-a0e0-4efc-b244-1422f2a9d6f7

src/App.tsx
victor-buldenko: no, we can`t, because what happens if active or completed todos where sum 0? Yeah, right - the footer was unmounted, but we still need the footer mounted. It's bad advice

src/App.tsx
victor-buldenko: fixed

src/App.tsx
victor-buldenko: https://mate-parttime.slack.com/archives/C03UQS59NRY/p1689857616536079

all these problems are discussed here and this option was accepted

lerastarynets (approved):
Keep up the great work!

PR #759, opened 2023-08-31
Solution

s3rg4sh:
https://s3rg4sh.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,233 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,24 +1,233 @@
...
+            />
+          )
+          : (
+            <TodoList
+              todos={filteredTodos}
+              onDelete={handleDeleteTodo}
+              onUpdate={handleUpdateTodo}
+              processingIds={processingIds}
+            />
+          )}
```
maxim2310, line 205: better to pass tempTodo as props and make proper logic inside component instead doing it by ternary operator with component

maxim2310 (changes requested):

https://www.loom.com/share/7501632b2d7c4d818000c82c77db2858

VitaliyBondarenko1982 (changes requested):
Good job.
Need a couple imrovements.

- do not show toggle button in header when no any todos in the list
Screenshot from 2023-09-01 13-01-11

- check logic for toggle button. 
working example:
design

actual:
actual

VitaliyBondarenko1982 (approved):

PR #765, opened 2023-09-16
initial commit

aliliolek:
[DEMO LINK](https://aliliolek.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,198 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  createTodo,
+  deleteTodo,
+  editTodo,
+  getTodos,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoFilter } from './types/TodoFilter';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Notifications } from './components/Notifications';
 
-const USER_ID = 0;
+const USER_ID = 11399;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [errors, setErrors] = useState({
+    load: false,
+    delete: false,
+    empty: false,
+    add: false,
+    edit: false,
+  });
+  const [currentFilter, setCurrentFilter] = useState(TodoFilter.All);
+  const [loadingId, setLoadingId] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const filterTodos = (initTodos?: Todo[]) => {
+    let filtered;
+
+    if (initTodos) {
+      filtered = [...initTodos];
+    } else {
+      filtered = [...todos];
+    }
+
+    switch (currentFilter) {
+      case TodoFilter.Active:
+        filtered = filtered.filter(todo => !todo.completed);
+        break;
+      case TodoFilter.Completed:
+        filtered = filtered.filter(todo => todo.completed);
+        break;
+      default:
+        break;
+    }
+
+    setVisibleTodos(filtered);
+  };
+
+  const loadTodos = () => {
+    getTodos(USER_ID).then(items => {
+      setTodos([...items]);
+      filterTodos(items);
+    }).catch(() => {
+      setErrors(prevErrors => ({ ...prevErrors, load: true }));
+    });
+  };
+
+  const removeTodo = (todoId: number) => {
+    setLoadingId(prevTodoId => [...prevTodoId, todoId]);
+
+    deleteTodo(todoId)
+      .then(() => {
+        setTodos(prevTodos => prevTodos
+          .filter(todo => todo.id !== todoId));
+        setLoadingId([]);
+      })
+      .catch(() => {
+        setErrors(prevErrors => ({
+          ...prevErrors,
+          delete: true,
+        }));
+        setLoadingId([]);
+      });
+  };
+
+  const addTodo = (newTitle: string) => {
+    const maxId = Math.max(...todos.map(todo => todo.id));
+
+    setTempTodo({
+      id: 0,
+      title: newTitle,
+      completed: false,
+      userId: USER_ID,
+    });
+
+    createTodo({
+      id: maxId + 1,
+      title: newTitle,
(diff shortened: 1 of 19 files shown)


src/App.tsx
```diff
@@ -1,24 +1,198 @@
...
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [errors, setErrors] = useState({
+    load: false,
+    delete: false,
+    empty: false,
+    add: false,
+    edit: false,
+  });
+  const [currentFilter, setCurrentFilter] = useState(TodoFilter.All);
+  const [loadingId, setLoadingId] = useState<number[]>([]);
```
TarasHoliuk, line 30:
```suggestion
  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
```
array - use plural form

src/App.tsx
```diff
@@ -1,24 +1,198 @@
...
+import { Footer } from './components/Footer';
+import { Notifications } from './components/Notifications';
 
-const USER_ID = 0;
+const USER_ID = 11399;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [errors, setErrors] = useState({
```
TarasHoliuk, line 22: use enum instead of object, where default value is Error.Empty:
```
enum Error {
  Empty = ''
  // other errors
}
```

src/App.tsx
```diff
@@ -1,24 +1,198 @@
...
+          ...prevErrors,
+          add: true,
+        }));
+        setTempTodo(null);
+      });
+  };
+
+  const changeTodo = (
+    property: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
TarasHoliuk, line 114: why any? try to figure out with types (maybe `boolean | string`, etc)

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const handleChangeTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setNewTitle(event.target.value);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (newTitle.trim()) {
+      addTodo(newTitle);
+    } else if (newTitle.trim() === '') {
```
TarasHoliuk, line 40:
```suggestion
    } else {
```

src/components/Notifications/Notifications.tsx
```diff
@@ -0,0 +1,91 @@
...
+    >
+      <button
+        type="button"
+        className="delete"
+        aria-label="delete-notification"
+        onClick={handleCLoseNotifications}
+      />
+
+      {/* show only one message at a time */}
+      {errors.load && (
```
TarasHoliuk, line 55: create function which returns you the correct error message (using `switch`) or just save the correct message in the enum mentioned above

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+    }
+  };
+
+  const handleSubmitEditing = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    saveChanges();
+  };
+
+  const handleBlur = () => {
```
TarasHoliuk, line 53: redundant function wrapper

`onBlur={saveChanges}`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const currentValue = event.target.value;
+
+    setInputValue(currentValue);
+  };
+
+  const saveChanges = () => {
+    if (inputValue.trim() && inputValue.trim() !== title.trim()) {
+      changeTodo('title', inputValue, id);
+      setIsEditingId(null);
+    } else if (inputValue.trim() === '') {
```
TarasHoliuk, line 40:
```suggestion
    } else if (!inputValue.trim()) {
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+  };
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const currentValue = event.target.value;
+
+    setInputValue(currentValue);
+  };
+
+  const saveChanges = () => {
+    if (inputValue.trim() && inputValue.trim() !== title.trim()) {
```
TarasHoliuk, line 37: no need to trim `title` here, save `inputValue.trim()` to a variable and reuse

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+      setIsEditingId(null);
+    }
+  };
+
+  const handleToggle = () => {
+    changeTodo('completed', !completed, id);
+  };
+
+  return (
+    <div className={`todo ${completed ? 'completed' : ''}`}>
```
TarasHoliuk, line 68: use classnames library for such cases

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+export const TodoItem: React.FC<TodoItemProps> = ({
+  todo: { id, title, completed },
+  loadingId,
+  removeTodo,
+  changeTodo,
+}) => {
+  const [isEditingId, setIsEditingId] = useState<number | null>(null);
+  const [inputValue, setInputValue] = useState(title);
+
+  const handleDoubleClick = () => {
```
TarasHoliuk, line 26: you should also focus on input when it appears - use ref for it (ask in chat if you have questions)

src/App.tsx
```diff
@@ -1,24 +1,198 @@
...
+    }
+  };
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  useEffect(() => {
+    filterTodos();
+  }, [currentFilter, todos]);
```
TarasHoliuk, line 153: you don't need `visibleTodos` state
general rule if you calculate some value based on other state values/props no need to save it in state

save it in a common variable, wrap in `useMemo` with the same dependency array

src/App.tsx
```diff
@@ -1,24 +1,198 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const filterTodos = (initTodos?: Todo[]) => {
+    let filtered;
+
+    if (initTodos) {
+      filtered = [...initTodos];
+    } else {
+      filtered = [...todos];
+    }
```
TarasHoliuk, line 40: you should get rid of it - make `todos` argument mandatory

TarasHoliuk (changes requested):

PR #770, opened 2023-09-19
add task solution

AnnaViolin23:
[DEMO LINK](https://AnnaViolin23.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,263 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  FormEvent, useEffect, useMemo, useState,
+} from 'react';
+import classNames from 'classnames';
+import { Todo } from './types/Todo';
+import { FilterType } from './types/FilterTypes';
+import { Error } from './types/Error';
+import { getTodos } from './api/todos';
+import * as postService from './api/todos';
 import { UserWarning } from './UserWarning';
+import { TodoList } from './components/TodoList';
+import { TodoFilter } from './components/FilterTodo';
+import { TodoError } from './components/TodoError';
+import { TodoItem } from './components/TodoItem';
 
-const USER_ID = 0;
+const USER_ID = 11707;
+
+const getVisibleTodos = (todos: Todo[], status: FilterType) => {
+  return todos.filter(todo => {
+    switch (status) {
+      case FilterType.Completed:
+        return todo.completed;
+
+      case FilterType.Active:
+        return !todo.completed;
+
+      default:
+        return true;
+    }
+  });
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState(FilterType.All);
+  const [errorMessage, setErrorMessage] = useState(Error.None);
+  const [query, setQuery] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingIds, setLoadingIds] = useState<number[]>([]);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => setErrorMessage(Error.Load));
+  }, []);
+
+  const visibleTodos = useMemo(() => getVisibleTodos(todos, status),
+    [todos, status]);
+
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,24 +1,263 @@
...
+      onDeleteTodo(todo.id);
+    });
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const handleToggleAll = () => {
+    const uncompletedTodos = todos.filter(todo => todo.completed === false);
```
volodymyr-soltys97, line 120:
```suggestion
    const uncompletedTodos = todos.filter(todo => !todo.completed);
```

src/App.tsx
```diff
@@ -1,24 +1,263 @@
...
+            {tempTodo && (
+              <TodoItem
+                todo={tempTodo}
+                onDeleteTodo={onDeleteTodo}
+                loadingIds={loadingIds}
+                setTodos={setTodos}
+                setLoadingIds={setLoadingIds}
+                todos={visibleTodos}
+
+              />
```
volodymyr-soltys97, line 224:
```suggestion
                todos={visibleTodos}
              />
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,191 @@
...
+            <span
+              className="todo__title"
+              onDoubleClick={handleDoubleClick}
+            >
+              {todo.title}
+            </span>
+            <button
+              type="button"
+              className="todo__remove"
+              onClick={() => onDeleteTodo(todo.id)}
```
volodymyr-soltys97, line 174: You can use a destructuring for `todo`

volodymyr-soltys97 (changes requested):
Great work 👍 
Let's improve your code

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  onDeleteTodo,
+  loadingIds,
+  setTodos,
+  setLoadingIds,
+  todos,
+}) => {
+  const [, setErrorMessage] = useState(Error.None);
```
etojeDenys, line 27: why do we need this state, if we do not need the value

src/App.tsx
```diff
@@ -1,24 +1,262 @@
...
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+
+          {todos.length !== 0 && (
+            <button
+              type="button"
+              className={classNames('todoapp__toggle-all', {
+                active: completedTodoCount !== 0,
+              })}
+              onClick={() => handleToggleAll()}
```
etojeDenys, line 191:
```suggestion
              onClick={handleToggleAll}
```
we can simplify it

src/App.tsx
```diff
@@ -1,24 +1,262 @@
...
+    Promise.all(uncompletedTodos.map(todo => {
+      return postService.updateTodo(todo.id, {
+        completed: !todo.completed,
+      });
+    }))
+      .then(() => {
+        getTodos(USER_ID)
+          .then((value) => {
+            setTodos(value);
+            setTodos(value);
```
etojeDenys, line 163: why do we need to call it twice?

etojeDenys (changes requested):

maxim2310 (changes requested):
You need to fix last comment and make commit and push

src/components/TodoItem.tsx
AnnaViolin23: Using setErrorMessage instead of directly throwing errors with throw provides more flexibility and control over how errors are managed and displayed in app. Also,  It centralizes error handling within the component, making it easier to manage and customize error responses.

volodymyr-soltys97: Still not fixed from the previous review, It's bad practice to write code like that, need to fix it.
You can create a default function and then call it at the right places with a certain value, and using state without using state is not recommended

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const [doubleClicked, setDoubleClicked] = useState(false);
+  const [editingTitle, setEditingTitle] = useState(todo.title);
```
volodymyr-soltys97, line 29: Consider will be better if you add a generic types everywhere for state for better readability

volodymyr-soltys97 (changes requested):

etojeDenys (approved):
lgtm

PR #772, opened 2023-09-20
Everything works outside of escape. Would love some advice on that.

szymonpachucki:
[DEMO LINK](https://szymonpachucki.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,101 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
+import { Errors } from './types/Errors';
+import { FilterOption } from './types/filterOption';
+import { TodoList } from './components/todoList';
+import { InCaseOfError } from './components/inCaseOfError';
+import { InputOfTodos } from './components/inputOfTodos';
+import { Footer } from './components/footer';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([
+    // Initialize with some example todos
+
+    {
+      id: 1,
+      title: 'Example Todo 1',
+      completed: false,
+      removed: false,
+      editing: false,
+    },
+
+    {
+      id: 2,
+      title: 'Example Todo 2',
+      completed: true,
+      removed: false,
+      editing: false,
+    },
+  ]);
+  const [error, setError] = useState<Errors | null>(null);
+  const [filterTodos, setFilterTodos] = useState<FilterOption>('All');
+  const [newTodo, setNewTodo] = useState<string>(''); // State for new todo input
+
+  const handleSetFilter = (newFilter: FilterOption) => {
+    setFilterTodos(newFilter);
+  };
+
+  const closeError = () => {
+    setError(null);
+  };
+
+  const addTodo = () => {
+    if (newTodo.trim() !== '') {
+      const newTodoItem: Todo = {
+        id: todos.length + 1,
+        title: newTodo,
+        completed: false,
+        removed: false,
+        editing: false,
+      };
+
+      setTodos([...todos, newTodoItem]);
+      setNewTodo(''); // Clear the input field
+    }
+  };
+
+  useEffect(() => {
+    // You can perform any additional actions here when the todos state changes.
+  }, [todos]);
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        {/* Input for adding new todos */}
+        <InputOfTodos
+          setNewTodo={setNewTodo}
+          newTodo={newTodo}
+          addTodo={addTodo}
+          todos={todos}
+          setTodos={setTodos}
+        />
+
+        {/* Pass the filtered todos to TodoList */}
+        {todos && (
+          <TodoList
+            todos={todos}
+            filterTodos={filterTodos}
+            setTodos={setTodos}
+          />
+        )}
+
+        {/* Hide the footer if there are no todos */}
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,24 +1,101 @@
...
+import { InCaseOfError } from './components/inCaseOfError';
+import { InputOfTodos } from './components/inputOfTodos';
+import { Footer } from './components/footer';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([
+    // Initialize with some example todos
```
domalewski, line 12: please remove comment, code it's simple to understand

src/components/footer.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { FilterOption } from '../types/filterOption';
+
+interface Props {
+  handleSetFilter: (newFilter: FilterOption) => void;
+  todos: Todo[];
+  setTodos: (todos: Todo[]) => void;
+}
+
+// eslint-disable-next-line max-len
+export const Footer: React.FC<Props> = ({ handleSetFilter, todos, setTodos }) => {
```
domalewski, line 13:
```suggestion
export const Footer: React.FC<Props> = ({
handleSetFilter,
todos,
setTodos }) => {
```
I know it's looks worst but then you don't have to disabling eslint

domalewski (commented):
It's working so I guess it's ok
Removed the unnecessary comments

domalewski (approved):
It's working so I guess it's ok
Removed the unnecessary comments

src/components/footer.tsx
choeqq, line 1: current filter should have different styles, when picked, as a user I want to know what kind of todos the app is displaying

src/App.tsx
```diff
@@ -1,24 +1,101 @@
...
+      };
+
+      setTodos([...todos, newTodoItem]);
+      setNewTodo(''); // Clear the input field
+    }
+  };
+
+  useEffect(() => {
+    // You can perform any additional actions here when the todos state changes.
+  }, [todos]);
```
choeqq, line 59: empty effect?

src/App.tsx
```diff
@@ -1,24 +1,101 @@
...
+            handleSetFilter={handleSetFilter}
+            todos={todos}
+            setTodos={setTodos}
+          />
+        )}
+      </div>
+
+      {/* Notification is shown in case of any error */}
+      {/* Add the 'hidden' class to hide the message smoothly */}
+      {error !== null && (
```
choeqq, line 96:
```suggestion
      {error && (
```

src/components/inputOfTodos.tsx
```diff
@@ -0,0 +1,77 @@
...
+        completed: !t.completed,
+      }));
+    }
+
+    setTodos(updatedTodos);
+  };
+
+  return (
+    <header className="todoapp__header">
+      {todos.length !== 0 && (
```
choeqq, line 56:
```suggestion
      {todos.length && (
```

src/components/task.tsx
```diff
@@ -0,0 +1,114 @@
...
+          ref={inputRef}
+          type="text"
+          className="todoapp__new"
+          placeholder="Empty todo will be deleted"
+          value={inputValue}
+          onChange={handleInputChange}
+          onKeyPress={handleInputKeyPress}
+          onBlur={handleInputBlur}
+        />
+      )}
```
choeqq, line 111: make sure to apply correct styles to editing input - it should look the same as Todo styles

src/components/todoList.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+    // Filter out todos with removed set to true
+    filteredTodos = filteredTodos.filter((todo) => !todo.removed);
+
+    return filteredTodos;
+  };
+
+  return (
+    <section className="todoapp__main">
+      {visibleTodos().map((todo) => (
```
choeqq, line 31:
```suggestion
      {visibleTodos.map((todo) => (
```
this should be a value instead of a function call

choeqq (changes requested):
please add comments to your code only when you want to implement some normally unexpected behaviour

consider using context in your app

src/types/filterOption.ts
```diff
@@ -0,0 +1 @@
+export type FilterOption = 'All' | 'Active' | 'Completed';
```
mbruhler, line 1: Change type according to `const enum` change request.

src/components/todoList.tsx
```diff
@@ -0,0 +1,34 @@
...
+  todos: Todo[];
+  filterTodos: FilterOption;
+  setTodos: React.Dispatch<React.SetStateAction<Todo[]>>;
+}
+
+export const TodoList: React.FC<Props> = ({ todos, filterTodos, setTodos }) => {
+  const visibleTodos = (() => {
+    let filteredTodos = todos;
+
+    if (filterTodos === 'Active') {
```
mbruhler, line 16: With the `const enum` change, you should do:
```
if (filterTodos === FilterOption.ACTIVE) { ... }
```

Change others too.

src/components/inputOfTodos.tsx
```diff
@@ -0,0 +1,77 @@
...
+        completed: !t.completed,
+      }));
+    }
+
+    setTodos(updatedTodos);
+  };
+
+  return (
+    <header className="todoapp__header">
+      {/* i can not write it as {todos.lenght && as when its value is 0 it will display  '0' on the screen  */}
```
mbruhler, line 55: yeah, the condition you've provided is fine, please clear the comment

mbruhler (changes requested):
For the problem with `cancel edit when Esc key is up`, you can do this in two ways:
1. register a listener for this button in `useEffect` and remove it on component unmount. If escape is pressed then trigger a function that changes the edit state
2. register an event listener on the input using `onKeyUp` and handling it there, similar to what you did with Enter.

I think 2. is better, because the scope of the event handler is just for the input, but I wanted to show you that there is also a possibility to register such event globally for component, not just input :)

src/components/inputOfTodos.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    setTodos(updatedTodos);
+  };
+
+  return (
+    <header className="todoapp__header">
+      {todos.length > 0 && (
+        <button
+          type="button"
+          className={`todoapp__toggle-all ${todos.every((t) => t.completed) ? 'active' : ''}`}
```
pawel-peksa, line 58: You itarate over the array of Todos in two places to get the same information (see line 38). This is not the best performance optimization. Let's try to do this only once and use the result in both places.

pawel-peksa (changes requested):
Good job! 💪 Almost done! 
-> Added one minor comment regarding the performance optimization. (Try to avoid unnecessary itarations in your code)

choeqq (approved):
LGTM ✅

PR #779, opened 2023-09-21
added functional for change completed status and edit todo title, cre…

DenisChakhalian:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,295 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useEffect,
+  useState,
+  useMemo,
+  useRef,
+} from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import {
+  deleteTodoById,
+  getTodos,
+  setNewTodo,
+  updateTodoById,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { Notification } from './components/Notification/Notification';
+import { StatusEnum } from './types/StatusEnum';
+import { TodosFilter } from './components/TodosFilter/TodosFilter';
+import { TodoList } from './components/TodoList/TodoList';
 
-const USER_ID = 0;
+const USER_ID = 11497;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isLoadingAddTodo, setIsLoadingAddTodo] = useState(false);
+  const [todosIdToDelete, setTodosIdToDelete] = useState<number[]>([]);
+  const [todosIdToUpdate, setTodosIdToUpdate] = useState<number[]>([]);
+  const [filter, setFilter] = useState<StatusEnum>(
+    window.location.hash.slice(2) as StatusEnum || StatusEnum.All,
+  );
+
+  const formInput = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (!todosIdToDelete.length && !isLoadingAddTodo && formInput.current) {
+      formInput.current.focus();
+    }
+  }, [isLoadingAddTodo, todosIdToDelete]);
+
+  const handleChangeFilter = (newFilter: StatusEnum) => {
+    setFilter(newFilter);
+  };
+
+  const handleChangeNewTodoTitle
+    = (event: React.ChangeEvent<HTMLInputElement>) => {
+      setNewTodoTitle(event.target.value);
+    };
+
+  const hasCompletedTodo = useMemo(() => {
+    return todos.some(todo => todo.completed);
+  }, [todos]);
+
+  const notCompletedTodosLength = useMemo(() => {
+    return todos.filter(todo => !todo.completed).length;
+  }, [todos]);
+
+  const isCompletedAll = useMemo(() => {
+    return todos.every(todo => todo.completed);
+  }, [todos]);
+
+  const handelSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!newTodoTitle.trim()) {
+      setErrorMessage('Title should not be empty');
+
+      return;
+    }
+
+    setErrorMessage('');
+    setIsLoadingAddTodo(true);
+
+    setTempTodo({
+      completed: false,
+      id: 0,
+      title: newTodoTitle.trim(),
+      userId: USER_ID,
+    });
+
+    setNewTodo({
+      userId: USER_ID,
+      title: newTodoTitle.trim(),
+      completed: false,
+    })
+      .then(response => {
+        setNewTodoTitle('');
+
+        return response;
+      })
+      .then((newTodo) => {
+        setTodos((prevState) => [...prevState, newTodo]);
+      })
+      .catch(() => {
+        setErrorMessage('Unable to add a todo');
+      })
+      .finally(() => {
+        setIsLoadingAddTodo(false);
+        setTempTodo(null);
+      });
+  };
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
+      });
+  }, []);
+
+  const handelDeleteTodoById = (todoId: number) => {
+    setTodosIdToDelete(prevState => [...prevState, todoId]);
+
+    deleteTodoById(todoId)
+      .then(() => {
+        setTodos(prevState => prevState.filter(todo => todo.id !== todoId));
+      })
+      .catch(() => {
+        setErrorMessage('Unable to delete a todo');
+      })
+      .finally(() => {
+        setTodosIdToDelete(prevState => prevState.filter(id => id !== todoId));
+      });
+  };
+
+  const handleChangeCompletedStatus
+    = (todoId: number, isCompleted: boolean) => {
+      setTodosIdToUpdate(prevState => [...prevState, todoId]);
+
+      updateTodoById(todoId, { completed: !isCompleted })
+        .then(() => {
+          setTodos(prevState => {
+            const currentTodo = prevState
+              .find(todo => todo.id === todoId);
+
+            if (!currentTodo) {
+              return prevState;
+            }
+
+            const newTodos = [...prevState];
+            const index = prevState.indexOf(currentTodo);
+
+            newTodos.splice(index, 1, {
+              ...currentTodo,
+              completed: !currentTodo.completed,
+            });
+
+            return newTodos;
+          });
+        })
+        .catch(() => {
+          setErrorMessage('Unable to update a todo');
+        })
+        .finally(() => {
+          setTodosIdToUpdate(prevState => prevState
+            .filter(id => id !== todoId));
+        });
+    };
+
+  const handleChangeTodoTitle
+    = (todoId: number, newTitle: string) => {
+      setTodosIdToUpdate(prevState => [...prevState, todoId]);
+
+      updateTodoById(todoId, { title: newTitle })
+        .then(() => {
+          setTodos(prevState => {
+            const currentTodo = prevState
+              .find(todo => todo.id === todoId);
+
+            if (!currentTodo) {
+              return prevState;
+            }
+
+            const newTodos = [...prevState];
+            const index = prevState.indexOf(currentTodo);
+
+            newTodos.splice(index, 1, {
+              ...currentTodo,
+              title: newTitle,
+            });
+
+            return newTodos;
+          });
+        })
+        .catch(() => {
+          setErrorMessage('Unable to update a todo');
+        })
+        .finally(() => {
+          setTodosIdToUpdate(prevState => prevState
+            .filter(id => id !== todoId));
+        });
+    };
+
+  const handelDeleteCompletedTodos = async () => {
+    const completedTodosId = todos
+      .filter(({ completed }) => completed)
+      .map(({ id }) => id);
+
+    await Promise
+      .all([...completedTodosId.map(id => handelDeleteTodoById(id))]);
+  };
+
+  const handleChangeAllCompletedStatus = async () => {
+    const todosId = todos
+      .filter(({ completed }) => {
+        if (!isCompletedAll) {
+          return !completed;
+        }
+
+        return true;
+      })
+      .map(({ id }) => id);
+
+    await Promise
+      .all([...todosId
+        .map(id => handleChangeCompletedStatus(id, isCompletedAll))]);
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {!!todos.length && (
+            <button
+              type="button"
+              className={classNames('todoapp__toggle-all', {
+                active: isCompletedAll,
+              })}
+              data-cy="ToggleAllButton"
+              onClick={handleChangeAllCompletedStatus}
+            />
+          )}
+
+          <form onSubmit={handelSubmit}>
+            <input
+              data-cy="NewTodoField"
+              type="text"
+              className="todoapp__new-todo"
+              placeholder="What needs to be done?"
+              value={newTodoTitle}
+              onChange={handleChangeNewTodoTitle}
+              disabled={isLoadingAddTodo}
+              ref={formInput}
+            />
+          </form>
+        </header>
+
+        <TodoList
+          todos={todos}
+          tempTodo={tempTodo}
+          todosIdToDelete={todosIdToDelete}
+          todosIdToUpdate={todosIdToUpdate}
+          onDeleteTodo={handelDeleteTodoById}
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,24 +1,295 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isLoadingAddTodo, setIsLoadingAddTodo] = useState(false);
+  const [todosIdToDelete, setTodosIdToDelete] = useState<number[]>([]);
+  const [todosIdToUpdate, setTodosIdToUpdate] = useState<number[]>([]);
+  const [filter, setFilter] = useState<StatusEnum>(
+    window.location.hash.slice(2) as StatusEnum || StatusEnum.All,
+  );
```
BudnikOleksii, line 34: You can use `useLocation` hook to get the hash.

```javascript
import useLocation from "react-router-dom";

const { hash } = useLocation();
```

src/App.tsx
```diff
@@ -1,24 +1,295 @@
...
+      formInput.current.focus();
+    }
+  }, [isLoadingAddTodo, todosIdToDelete]);
+
+  const handleChangeFilter = (newFilter: StatusEnum) => {
+    setFilter(newFilter);
+  };
+
+  const handleChangeNewTodoTitle
+    = (event: React.ChangeEvent<HTMLInputElement>) => {
```
BudnikOleksii, line 49:
```suggestion
  const handleChangeNewTodoTitle = (
    event: React.ChangeEvent<HTMLInputElement>
  ) => {
```
Don't forget about formating

src/App.tsx
```diff
@@ -1,24 +1,295 @@
...
+
+    setErrorMessage('');
+    setIsLoadingAddTodo(true);
+
+    setTempTodo({
+      completed: false,
+      id: 0,
+      title: newTodoTitle.trim(),
+      userId: USER_ID,
+    });
```
BudnikOleksii, line 82:
```suggestion
    const newTodo = {
      completed: false,
      title: newTodoTitle.trim(),
      userId: USER_ID,
    };

    setTempTodo({
      ...newTodo,
      id: 0,
    });
```
As you pass the same below to `setNewTodo`

src/App.tsx
```diff
@@ -1,24 +1,295 @@
...
+
+    setNewTodo({
```
BudnikOleksii, line 84:
```suggestion
    addNewTodo({
```
Let's change the name to `addNewTodo` as it's a bit confusing, as when set we expect that we just set it to state

src/App.tsx
```diff
@@ -1,24 +1,295 @@
...
+      });
+  };
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
+      });
+  }, []);
```
BudnikOleksii, line 112: Let's add a loader to notify user while loading todos

src/App.tsx
```diff
@@ -1,24 +1,295 @@
...
+
+            const newTodos = [...prevState];
+            const index = prevState.indexOf(currentTodo);
+
+            newTodos.splice(index, 1, {
+              ...currentTodo,
+              completed: !currentTodo.completed,
+            });
+
+            return newTodos;
```
BudnikOleksii, line 151: Just use `map` instead of `indexOf` + `splice` and `find` we don't need in this case too

src/App.tsx
```diff
@@ -1,24 +1,295 @@
...
+          });
+        })
+        .catch(() => {
+          setErrorMessage('Unable to update a todo');
+        })
+        .finally(() => {
+          setTodosIdToUpdate(prevState => prevState
+            .filter(id => id !== todoId));
+        });
```
BudnikOleksii, line 194: You do the same in case of a change status, you can create one handler for both cases

src/App.tsx
```diff
@@ -1,24 +1,295 @@
...
+    };
+
+  const handelDeleteCompletedTodos = async () => {
+    const completedTodosId = todos
+      .filter(({ completed }) => completed)
+      .map(({ id }) => id);
+
+    await Promise
+      .all([...completedTodosId.map(id => handelDeleteTodoById(id))]);
```
BudnikOleksii, line 203: Great 🔥 
But don't forget to wrap it into `try/catch` block

src/App.tsx
```diff
@@ -1,24 +1,295 @@
...
+          return !completed;
+        }
+
+        return true;
+      })
+      .map(({ id }) => id);
+
+    await Promise
+      .all([...todosId
+        .map(id => handleChangeCompletedStatus(id, isCompletedAll))]);
```
BudnikOleksii, line 219: The same as above and improve a bit formating

src/App.tsx
```diff
@@ -1,24 +1,295 @@
...
+              type="text"
+              className="todoapp__new-todo"
+              placeholder="What needs to be done?"
+              value={newTodoTitle}
+              onChange={handleChangeNewTodoTitle}
+              disabled={isLoadingAddTodo}
+              ref={formInput}
+            />
+          </form>
+        </header>
```
BudnikOleksii, line 255: Create a separate component `Header`

src/App.tsx
```diff
@@ -1,24 +1,295 @@
...
+              type="button"
+              className="todoapp__clear-completed"
+              disabled={!hasCompletedTodo}
+              onClick={() => handelDeleteCompletedTodos()}
+            >
+              Clear completed
+            </button>
+
+          </footer>
+        )}
```
BudnikOleksii, line 286: Create a separate component `Footer`

src/api/todos.ts
```diff
@@ -0,0 +1,21 @@
...
+export const deleteTodoById = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const updateTodoById
+  = (todoId: number, data: Partial<Todo>) => {
+    return client.patch(`/todos/${todoId}`, data);
+  };
+
+// Add more methods here
```
BudnikOleksii, line 21: Remove comments

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,139 @@
...
+          </button>
+        </>
+      )}
+
+      <div
+        data-cy="TodoLoader"
+        className={classnames('modal', 'overlay', {
+          'is-active': id === 0
+            || todosIdToDelete.includes(id)
+            || todosIdToUpdate.includes(id),
```
BudnikOleksii, line 131: Create a variable for this condition

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,79 @@
...
+  onChangeCompletedStatus,
+  onChangeTitle,
+  onDeleteTodo,
+  tempTodo,
+  todosIdToDelete,
+  todosIdToUpdate,
+}) => {
+  return (
+    <section data-cy="TodoList" className="todoapp__main">
+      <TransitionGroup>
```
BudnikOleksii, line 33: 🚀

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,79 @@
...
+          >
+            <TodoItem
+              todo={tempTodo}
+              todosIdToDelete={todosIdToDelete}
+              todosIdToUpdate={todosIdToUpdate}
+              onDeleteTodo={onDeleteTodo}
+              onChangeTitle={onChangeTitle}
+              onChangeCompletedStatus={onChangeCompletedStatus}
+            />
+          </CSSTransition>
```
BudnikOleksii, line 74: You repeat this twice, maybe it's better to keep with `TodoItem`, but I'm not sure if `CSSTransition` works properly in this case.

Also, a few hints:
- to avoid props drilling you can use context.
- if you don't need at this level props and just want to pass them to the next component you can use the `spread` syntax. Example:
```javascript
function Profile(props) {
  return (
    <div className="card">
      <Avatar {...props} />
    </div>
  );
}
```
https://react.dev/learn/passing-props-to-a-component
- also, in this case, you can reuse props type

BudnikOleksii (changes requested):
Looks great 🚀 
Left some suggestions to improve your solution

src/App.tsx
DenisChakhalian: it dosen't work

src/App.tsx
DenisChakhalian: why would I do try/catch if I already have error handling in handelDeleteTodoById?

src/App.tsx
```diff
@@ -1,24 +1,284 @@
...
+  } = useContext(LoadingTodosContext);
+
+  const [isLoadingTodos, setIsLoadingTodos] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+
+  const [isLoadingAddTodo, setIsLoadingAddTodo] = useState(false);
+
+  const formInput = useRef<HTMLInputElement>(null);
```
roman-mirzoian, line 57:
```suggestion
  const formInputRef = useRef<HTMLInputElement>(null);
```

src/App.tsx
```diff
@@ -1,24 +1,284 @@
...
+    event: React.ChangeEvent<HTMLInputElement>,
+  ) => {
+    setNewTodoTitle(event.target.value);
+  };
+
+  const hasCompletedTodo = useMemo(() => {
+    return todos.some(todo => todo.completed);
+  }, [todos]);
+
+  const notCompletedTodosLength = useMemo(() => {
```
roman-mirzoian, line 75:
```suggestion
  const activeTodosLength = useMemo(() => {
```

src/App.tsx
```diff
@@ -1,24 +1,284 @@
...
+    return todos.filter(todo => !todo.completed).length;
+  }, [todos]);
+
+  const isCompletedAll = useMemo(() => {
```
roman-mirzoian, line 79:
```suggestion
  const isAllTodoCompleted = useMemo(() => {
```

src/App.tsx
```diff
@@ -1,24 +1,284 @@
...
+  const handelDeleteTodoById = (todoId: number) => {
+    setTodosIdToDelete(prevState => [...prevState, todoId]);
+
+    return deleteTodoById(todoId)
+      .then(() => {
+        setTodos(prevState => prevState.filter(todo => todo.id !== todoId));
+      })
+      .catch(() => {
+        setErrorMessage('Unable to delete a todo');
+        throw new Error();
```
roman-mirzoian, line 146: It is also worth passing the error text to the `Error` constructor, because without it there is no point in throwing out the error (below is a screenshot of the error from your application):
image
Fix this in all code.

src/App.tsx
```diff
@@ -1,24 +1,284 @@
...
+      .finally(() => {
+        setTodosIdToUpdate(prevState => prevState
+          .filter(id => id !== todoId));
+      });
+  };
+
+  const handelDeleteCompletedTodos = async () => {
+    const completedTodosId = todos
+      .filter(({ completed }) => completed)
+      .map(({ id }) => id);
```
roman-mirzoian, line 204: The same can be done through the `reduce` method and walk through the array once instead of twice.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const Footer :React.FC<Props> = ({
+  notCompletedTodosLength,
+  hasCompletedTodo,
+  onDeleteCompletedTodos,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${notCompletedTodosLength} items left`}
```
roman-mirzoian, line 18: When the number is one, there should be no plural message:
image

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
roman-mirzoian, line 1: Try to rewrite the code so that it is not necessary to disable the linter

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
roman-mirzoian, line 1: Try to rewrite the code so that it is not necessary to disable the linter

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,47 @@
...
+  errorMessage,
+  onCloseErrorMessage,
+}) => {
+  useEffect(() => {
+    if (errorMessage) {
+      setTimeout(() => {
+        onCloseErrorMessage('');
+      }, 3000);
+    }
+  }, [errorMessage]);
```
roman-mirzoian, line 21: https://upmostly.com/tutorials/settimeout-in-react-components-using-hooks

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,178 @@
...
+  onChangeTitle,
+  onChangeCompletedStatus,
+}) => {
+  const {
+    todosIdToDelete,
+    todosIdToUpdate,
+  } = useContext(LoadingTodosContext);
+  const { title, completed, id } = todo;
+
+  const editInput = useRef<HTMLInputElement>(null);
```
roman-mirzoian, line 34:
```suggestion
  const editInputRef = useRef<HTMLInputElement>(null);
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,178 @@
...
+    }
+  };
+
+  const saveChange = async () => {
+    try {
+      if (!editingTitle.trim()) {
+        await onDeleteTodo(id);
+      } else if (editingTitle.trim() !== title) {
+        await onChangeTitle(id, editingTitle.trim());
+      }
```
roman-mirzoian, line 83:
```suggestion
      const preparedTitle = editingTitle.trim();
      if (!preparedTitle) {
        await onDeleteTodo(id);
      } 
      if (preparedTitle !== title) {
        await onChangeTitle(id, preparedTitle);
      }
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const handleDeleteTodo = async () => {
+    try {
+      await onDeleteTodo(id);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.log(error);
+    }
+  };
+
+  const saveChange = async () => {
```
roman-mirzoian, line 77:
```suggestion
  const saveChanges = async () => {
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,178 @@
...
+        editing: isEdit,
+      })}
+      data-cy="Todo"
+    >
+      <label className="todo__status-label">
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          onChange={() => handleChangeCompletedStatus()}
```
roman-mirzoian, line 123:
```suggestion
          onChange={handleChangeCompletedStatus}
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,178 @@
...
+            className="todo__title"
+          >
+            {title}
+          </span>
+
+          <button
+            data-cy="TodoDelete"
+            type="button"
+            className="todo__remove"
+            onClick={() => handleDeleteTodo()}
```
roman-mirzoian, line 160:
```suggestion
            onClick={handleDeleteTodo}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,75 @@
...
+        {todos.filter(todo => {
+          switch (filter) {
+            case StatusEnum.Active:
+              return !todo.completed;
+            case StatusEnum.Completed:
+              return todo.completed;
+            case StatusEnum.All:
+            default:
+              return todo;
+          }
```
roman-mirzoian, line 43: It is worth putting it in a separate variable and using the finished result to improve the readability of the code.

roman-mirzoian (changes requested):

src/components/Footer/Footer.tsx
DenisChakhalian: 😅😅
image

src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
+    if (todo.id === updatedTodo.id) {
+      return {
+        ...todo,
+        [updatedField]: updatedTodo[updatedField],
+      };
+    }
+
+    return todo;
+  });
+}
```
BudnikOleksii, line 40: Move to `utils` folder

src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
 
 export const App: React.FC = () => {
+  const { todos, setTodos } = useContext(TodosContext);
+  const {
+    todosIdToDelete,
+    setTodosIdToDelete,
+    setTodosIdToUpdate,
+  } = useContext(LoadingTodosContext);
```
BudnikOleksii, line 48: You can create hooks for contexts so you don't have to pass every time the context.
```javascript
const useTodos = () => useContext(TodosContext);
const useLoadingTodos = () => useContext(LoadingTodosContext);
```

src/components/TodosFilter/TodosFilter.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useContext } from 'react';
+import classnames from 'classnames';
+import { StatusEnum } from '../../types/StatusEnum';
+import { FilterTodosContext } from '../../context/TodosContexts';
+
+type Props = {};
+
+export const TodosFilter: React.FC<Props> = () => {
```
BudnikOleksii, line 8:
```suggestion

export const TodosFilter = () => {
```

BudnikOleksii (approved):
Nicely done 🔥 
Approved, just one suggestion. Split your logic, don't store all logic in the `App` component.
For example:
- filter logic in the filter component
- toggle all in header as the button in this component
- add new todo in the form component or in the header too
- and so on

The smaller your components, the easier it is to manage and support them.

PR #780, opened 2023-09-21
Develop

AndrMar1939:
[DEMO LINK](https://andrmar1939.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/types/actionTypes.ts
+++ b/src/types/actionTypes.ts
@@ -1,0 +1,21 @@
+import { Todo, TodosListType } from './todosTypes';
+
+export type Actions = {
+  type: 'LOAD',
+  payload: TodosListType,
+} | {
+  type: 'POST',
+  payload: Todo,
+} | {
+  type: 'DELETE',
+  payload: number,
+} | {
+  type: 'PATCH',
+  payload: Todo,
+} | {
+  type: 'IS_DELETING',
+  payload: number,
+} | {
+  type: 'REMOVE_IS_DELETING',
+  payload: number,
+};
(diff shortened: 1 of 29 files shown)


src/types/actionTypes.ts
```diff
@@ -0,0 +1,24 @@
...
+} | {
+  type: 'ALL_ACTIVE',
+  payload: null,
+} | {
+  type: 'IS_SPINNING',
+  payload: number,
+} | {
+  type: 'REMOVE_SPINNING',
+  payload: number,
+};
```
BudnikOleksii, line 24: It's not readable, create a type for each action and then combine them with union type

src/types/apiErrorsType.ts
```diff
@@ -0,0 +1,5 @@
+export type EmptyInputErrorType = 'REQUIRED';
+
+export const emptyInputError: EmptyInputErrorType = 'REQUIRED';
```
BudnikOleksii, line 3: It's too much, if you set some string it gets only this string as type so you don't need to create another type for constant

BudnikOleksii (changes requested):
Looks good 🚀 
I Left some comments to improve your solution, check them.
Also, a lot of tests failed:
image

AndrMar1939:
> Looks good 🚀 I Left some comments to improve your solution, check them. Also, a lot of tests failed: image



> Looks good 🚀 I Left some comments to improve your solution, check them. Also, a lot of tests failed: image

Yes!!! I was trying to work with Bulma's additional styles, so the editing tests failed. Okay.

Moroz-Dmytro (changes requested):
I see Oleksii has already reviewed it
Just some bug with the platform, I was not supposed to get this PR to review as it was not fixed yet
Need to reject this 🙂
Please, fix previous comments and I'll review it 😉

PR #781, opened 2023-09-21
todo app final part

AnVBondar:
[DEMO LINK](https://AnVBondar.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/TodoContext.tsx
+++ b/src/TodoContext.tsx
@@ -1,0 +1,92 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import React, { useEffect, useState } from 'react';
+import { TodoType } from './types/TodoType';
+import { Filter } from './types/Filter';
+import { ErrorsType } from './types/ErrorsType';
+import { getTodos } from './api/todos';
+
+const USER_ID = 10236;
+
+type TodoContextType = {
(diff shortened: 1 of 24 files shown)


src/TodoContext.tsx
```diff
@@ -0,0 +1,92 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
VitaliyBodnar, line 1: why do you need `unused-vars` ?

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const toggleAll = async () => {
+    const isAnySelected = todos.some(({ completed }) => completed);
+    let preparedList: TodoType[] = [];
+
+    if (isAnySelected) {
+      const sortAllInSelect = todos.filter(({ completed }) => !completed);
+
+      preparedList = sortAllInSelect.map(item => {
+        return { ...item, completed: true };
+      });
```
VitaliyBodnar, line 78: you can combine map and filter with using method reduce

VitaliyBodnar (changes requested):

VitaliyBondarenko1982 (approved):
Looks good

PR #783, opened 2023-09-22
Add solution

WaveOfDandelions:
[DEMO LINK](https://WaveOfDandelions.github.io/react_todo-app-with-api/)


HaniaNassalska (changes requested):
I’ve noticed you don't have focus automatically set on input after entering the page.
When there is an error when editing a task - the task is saved. It should still be in edit mode, and the focus should be on it.
The error ‘Unable to add a todo’ is not displayed.
Check everything in the Chrome DevTools (network) and set it offline 🙂

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+      .then((data) => {
+        setTodos(data);
+        setTempTodo(null);
+      }).catch(() => setNotification(NOTIFICATION.LOAD));
+  }, []);
+
+  useEffect(() => {
+    setListOfTodoId([]);
+    setUpdatingTodos([]);
+  }, [todos]);
```
choeqq, line 40: why are you setting todos to empty lists when todos are changing?

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+  const [listOfTodoId, setListOfTodoId] = useState<number[]>([]);
+  const [updatingTodos, setUpdatingTodos] = useState<number[]>([]);
+  const [titleFocus, setTitleFocus] = useState(true);
+
+  useEffect(() => {
+    todoService.getTodos(USER_ID)
+      .then((data) => {
+        setTodos(data);
+        setTempTodo(null);
+      }).catch(() => setNotification(NOTIFICATION.LOAD));
```
choeqq, line 34: there should probably be a loading state for fetching todos

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+      setLoading(false);
+
+      return;
+    }
+
+    completedTodosList.map(todo => deleteTodo(todo.id));
+  }, [todos]);
+
+  const changeTodoCompleted = useCallback((todoId: number) => {
+    setLoading(true);
```
choeqq, line 106: I think there should be a separate loading state for each operation

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+
+    const todoForUpdate = findTodoById(todos, todoId);
+
+    if (todoForUpdate !== null) {
```
choeqq, line 110:
```suggestion
    if (todoForUpdate) {
```

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+          loading={loading}
+          todos={filteredTodos}
+          deleteTodo={deleteTodo}
+          tempTodo={tempTodo}
+          updateTodo={(todo, title) => updateTodo(todo, title)}
+          onChange={(todoId) => changeTodoCompleted(todoId)}
+          updatingTodos={updatingTodos}
+        />
+
+        {todos.length !== 0
```
choeqq, line 221:
```suggestion
        {todos.length
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { completedTodos } from '../../utils/TodoFilter';
+
+type Props = {
+  todos: Todo[]
+  filterType: FilterType,
+  setFilterType: React.Dispatch<React.SetStateAction<FilterType>>,
+  removeCompleted: () => void,
+};
+
+export const Footer: React.FC<Props> = React.memo(
```
choeqq, line 14: why is this component so heavy it has to be memoized?

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+type Props = {
+  loading: boolean,
+  todos: Todo[]
+  addTodo: (newTodo: Todo) => void
+  reverteCompletedTodo: (todosForRevert: Todo[]) => void,
+  titleFocus: boolean,
+};
+
+export const Form: React.FC<Props> = React.memo(
```
choeqq, line 21: is React.memo necessary?

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,91 @@
...
+      if (formInput.current) {
+        formInput.current.focus();
+      }
+    }, [titleFocus]);
+
+    const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+      event.preventDefault();
+
+      const newTodo = {
+        id: 0,
```
choeqq, line 53: I don't think `id: 0` is good for reach mapping 😄

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,91 @@
...
+        completed: false,
+      };
+
+      addTodo(newTodo);
+      setValue('');
+    };
+
+    return (
+      <>
+        {todos.length !== 0
```
choeqq, line 65:
```suggestion
        {todos.length
```

src/components/Notifications/Notifications.tsx
```diff
@@ -0,0 +1,35 @@
...
+  notification: NOTIFICATION;
+};
+
+export const Notifications: React.FC<Props> = ({ notification }) => {
+  const [isHidden, setIsHidden] = useState(false);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setIsHidden(true);
+    }, 3000);
```
choeqq, line 16: If you want to use timeout to artificially extended closing time of notifications , you should remember to clean it up after usage

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,185 @@
...
+};
+
+export const TodoList: React.FC<Props> = ({
+  listOfTodoId,
+  loading,
+  todos,
+  deleteTodo = () => { },
+  tempTodo,
+  updateTodo = () => { },
+  onChange = () => { },
```
choeqq, line 29: are your using prettier? those brackets looks little off 😄

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  useEffect(() => {
+    if (selectedTodo !== null) {
+      setTempTitle(selectedTodo.title);
+    }
+  }, [selectedTodoId]);
+
+  useEffect(() => {
+    setIsUpdating(null);
+  }, todos);
```
choeqq, line 48: you CANNOT rely on array in `useEffect` arrays in JS are compared by reference, so [1,2] === [] // gives us true - so since this comparison will always give us true - your component will render infinitely (Stack Overflow 😄 ) if you want to use `useEffect` make sure to NEVER pass any object or array into dependecies

choeqq (changes requested):

choeqq (changes requested):
Waiting for any updates @WaveOfDandelions 👀

src/App.tsx
WaveOfDandelions: > why are you setting todos to empty lists when todos are changing?

fixed

src/App.tsx
WaveOfDandelions: > there should probably be a loading state for fetching todos

In this example, I was create a loading state and initialize it to false. In the useEffect hook, I initiate the API request and set loading to true before the request. Once the API request is complete, we set loading back to false.

In the return statement, I check if loading is true, and if so, we display a loading message. If loading is false, I continue to render the fetched data.

src/App.tsx
WaveOfDandelions: fixed

src/App.tsx
WaveOfDandelions: fixed

WaveOfDandelions:
> Waiting for any updates @WaveOfDandelions 👀

totally rewrite and decompose app, because many spaghetti code in app.tsx

src/components/TodoList/TodoList.tsx
WaveOfDandelions: > you CANNOT rely on array in `useEffect` arrays in JS are compared by reference, so [1,2] === [] // gives us true - so since this comparison will always give us true - your component will render infinitely (Stack Overflow 😄 ) if you want to use `useEffect` make sure to NEVER pass any object or array into dependecies

fixed

src/components/TodoList/TodoList.tsx
WaveOfDandelions: > are your using prettier? those brackets looks little off 😄

fixed

src/components/Notifications/Notifications.tsx
WaveOfDandelions: > If you want to use timeout to artificially extended closing time of notifications , you should remember to clean it up after usage

fixed

src/components/Form/Form.tsx
WaveOfDandelions: > 

fixed

src/components/Form/Form.tsx
WaveOfDandelions: > I don't think `id: 0` is good for reach mapping 😄

fixed

cypress/integration/page.spec.js
pawel-peksa, line 1: You should skip test execution as a last resort. Please revert changes in this file and aim for passing as much tests as possible. It is not a problem to skip couple of tests when the app functionality is correct, but skipping more than 300 is not a way to go.

src/components/Form/Form.tsx
pawel-peksa: Is it? I still can see it in the code ;P

src/components/Form/Form.tsx
pawel-peksa: Is it?

src/components/Notifications/Notifications.tsx
pawel-peksa: It is not fixed.
Please refer to this section in React docs to see how we can clean up timeouts in useEffect:
https://react.dev/learn/synchronizing-with-effects#putting-it-all-together

src/components/TodoList/TodoList.tsx
pawel-peksa: Please apply changes to the code according to this comment.

src/App.tsx
pawel-peksa: I cannot see any changes applied here

src/App.tsx
pawel-peksa: In the code, I can still see `todos.length !== 0`

pawel-peksa (changes requested):
It looks like you did not push all the changes required by previeous reviewer. Please have another look.

src/App.tsx
WaveOfDandelions: > I cannot see any changes applied here
I was rewrite all code because many mistakes was made

Please check once more

src/components/Form/Form.tsx
pawel-peksa: We might remove memo from here.

WaveOfDandelions:
> Looks better, but still there are some comments from previous reviews that need to to handled

So it was some problems because I forgot delete old src files, now I was update Notification, Form and App.tsx
Please check again

pawel-peksa (changes requested):
Now everything was fixed, good job! 👍 
Just one last thing, please remove all `skip` from the test suite and let's check how many tests are failing. Then we can skip only those that are failing.

WaveOfDandelions:
> Now everything was fixed, good job! 👍 Just one last thing, please remove all `skip` from the test suite and let's check how many tests are failing. Then we can skip only those that are failing.

Okay!

WaveOfDandelions:
> Now everything was fixed, good job! 👍 Just one last thing, please remove all `skip` from the test suite and let's check how many tests are failing. Then we can skip only those that are failing.

Okay!

WaveOfDandelions:
> Now everything was fixed, good job! 👍 Just one last thing, please remove all `skip` from the test suite and let's check how many tests are failing. Then we can skip only those that are failing.

Fixed

choeqq (approved):
LGTM ✅

PR #784, opened 2023-09-23
implenent change todo functionality

kaavka:
[DEMO LINK](https://kaavka.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/contexts/TodosContext.tsx
+++ b/src/contexts/TodosContext.tsx
@@ -1,0 +1,276 @@
+import React, {
+  createContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import { Todo } from '../types/Todo';
+import { FilterType } from '../types/TodoStatus';
+import { filterBy } from '../utils/filterBy';
+import {
+  INVALID_TITLE_ERROR_MESSAGE,
+  UNABLE_ADD_ERROR_MESSAGE,
+  UNABLE_DELETE_ERROR_MESSAGE,
+  UNABLE_DOWNLOAD_ERROR_MESSAGE,
+  UNABLE_UPDATE_ERROR_MESSAGE,
+  USER_ID,
+} from '../utils/constants';
+import {
+  addTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+} from '../api/todos';
+import { ErrorType } from '../types/ErrorType';
+import { TodoUpdateData } from '../types/TodoUpdateData';
+
+interface TodosContextType {
+  todos: Todo[];
+  setTodos: (todos: Todo[]) => void;
+
+  filterType: FilterType;
+  setFilterType: (filterType: FilterType) => void;
+
+  tempTodo: Todo | null;
+  setTempTodo: (todo: Todo | null) => void;
+
+  todosIdsUpdating: number[];
+  setTodosIdsUpdating: (ids: number[]) => void;
+
+  errorMessage: ErrorType | null;
+  setErrorMessage: (error: ErrorType) => void;
+
+  visibleTodos: Todo[];
+
+  handleFilterChange: (type: FilterType) => void;
+  handleAddTodo: (
+    title: string
+  ) => Promise<boolean>;
+  handleChangeTodo: (todoId: number, data: TodoUpdateData) => void;
+  handleDeleteTodo: (...ids: number[]) => void;
+  handleDeleteCompletedTodo: () => void;
+  handleToogleTodo: () => void,
+}
+
+const initialTodosState: TodosContextType = {
+  todos: [],
+  setTodos: () => {},
+  filterType: FilterType.All,
+  setFilterType: () => {},
+  tempTodo: null,
+  setTempTodo: () => {},
+  todosIdsUpdating: [],
+  setTodosIdsUpdating: () => {},
+  errorMessage: null,
+  setErrorMessage: () => {},
+  visibleTodos: [],
+  handleFilterChange: () => {},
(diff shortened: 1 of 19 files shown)


AntonHerasymov (approved):
Good job!!
But fix pressing "Esc" while editing

src/utils/filterBy.ts
```diff
@@ -0,0 +1,21 @@
+import { Todo } from '../types/Todo';
+import { FilterType } from '../types/TodoStatus';
+
+export const filterBy = (todos: Todo[], type: FilterType) => {
+  return todos.filter(({ completed }) => {
+    switch (type) {
+      case FilterType.Active: {
+        return completed === false;
```
Moroz-Dmytro, line 8:
```suggestion
        return !completed;
```

src/utils/filterBy.ts
```diff
@@ -0,0 +1,21 @@
...
+      }
+
+      case FilterType.Completed: {
+        return completed === true;
```
Moroz-Dmytro, line 12:
```suggestion
        return completed;
```

Moroz-Dmytro (changes requested):
1. Should be `1 item` without `s`

image

2. Not implemented

image

3. Not implemented

image

4. I can rename todo to empty name(enter just spaces), i expect that todo will also be deleted

image

5. If I click `toggleAll` in `offline` mode -> I don't see error message, just loaders

image

6. If I click `Delete all` in `offline` mode -> I don't see any error message

src/contexts/TodosContext.tsx
```diff
@@ -0,0 +1,287 @@
...
+        const index = currentTodos.findIndex(todo => todo.id === id);
+        const updatedTodos = [...currentTodos];
+
+        updatedTodos.splice(index, 1, updatedTodo);
+
+        return updatedTodos;
+      });
+    } catch {
+      setError(UNABLE_UPDATE_ERROR_MESSAGE);
+      throw new Error();
```
roman-mirzoian, line 168: It would be good to also pass the error text to the `Error` constructor.

src/contexts/TodosContext.tsx
```diff
@@ -0,0 +1,287 @@
...
+    } finally {
+      setTodosIdsUpdating([]);
+    }
+  };
+
+  const handleDeleteCompletedTodo = async () => {
+    setErrorMessage(null);
+    const completedTodosId = todos
+      .filter(todo => todo.completed)
+      .map(todo => todo.id);
```
roman-mirzoian, line 238: Use the `reduce` method to walk the array once instead of twice and improve the performance of your function.

src/contexts/TodosContext.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+    setTodosIdsUpdating(completedTodosId);
+
+    const deletePromises = completedTodosId.map(async (id) => {
```
roman-mirzoian, line 242:
```suggestion
    const deleteByIdPromises = completedTodosId.map(async (id) => {
```

src/contexts/TodosContext.tsx
```diff
@@ -0,0 +1,287 @@
...
+      try {
+        await deleteTodo(id);
+
+        return id;
+      } catch {
+        return null;
```
roman-mirzoian, line 248: An error should be caught in the `catch` block.

src/contexts/TodosContext.tsx
```diff
@@ -0,0 +1,287 @@
...
+      }
+    });
+
+    const deletedIds = await Promise.all(deletePromises);
+
+    if (deletedIds.includes(null)) {
+      setError(UNABLE_DELETE_ERROR_MESSAGE);
+    }
```
roman-mirzoian, line 256: This functionality should be combined with the `catch` block in the `map` method above.

src/contexts/TodosContext.tsx
```diff
@@ -0,0 +1,287 @@
...
+        tempTodo,
+        todosIdsUpdating,
+        errorMessage,
+        visibleTodos,
+        handleFilterChange,
+        handleAddTodo,
+        handleChangeTodo,
+        handleDeleteTodo,
+        handleDeleteCompletedTodo,
+        handleToogleTodo: handleToggleTodo,
```
roman-mirzoian, line 280: Why?
Use just `handleToggleTodo` right away, because `toogle` it`s not correct.

src/utils/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const USER_ID = 11499;
+export const DELAY_TO_HIDE_ERROR = 3000;
+export const UNABLE_DOWNLOAD_ERROR_MESSAGE = 'Unable to load todos';
+export const UNABLE_ADD_ERROR_MESSAGE = 'Unable to add a todo';
+export const UNABLE_DELETE_ERROR_MESSAGE = 'Unable to delete a todo';
+export const UNABLE_UPDATE_ERROR_MESSAGE = 'Unable to update a todo';
```
roman-mirzoian, line 6:
```suggestion
export const HIDE_ERROR_DELAY = 3000;
export const DOWNLOAD_ERROR_MESSAGE = 'Unable to load todos';
export const ADD_TODO_ERROR_MESSAGE = 'Unable to add a todo';
export const DELETE_TODO_ERROR_MESSAGE = 'Unable to delete a todo';
export const UPDATE_TODO_ERROR_MESSAGE = 'Unable to update a todo';
```

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Good updates, well done 👍

PR #790, opened 2023-09-25
Task solution - implemented toggling and editing

yuliiaant:
[DEMO LINK](https://yuliiaant.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,240 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import * as postService from './api/todos';
+import { getFilteredTodos } from './utils/functions';
+import { TodoItem } from './components/TodoItem';
+import { TodoFilter } from './components/TodoFilter';
+import { Header } from './components/Header';
+import { ErrorMessage } from './components/ErrorMessage';
+import { TempTodo } from './components/TempTodo';
+import { Status } from './types/Status';
+import { ErrorMessages } from './utils/ErrorMessages';
 
-const USER_ID = 0;
+const USER_ID = 11457;
+const initialTodos: Todo[] = [];
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[]>(initialTodos);
+  const [filterBy, setFilterBy] = useState(Status.All);
+  const [errorMessage, setErrorMessage] = useState(ErrorMessages.NoError);
+  const [title, setTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingId, setLoadingId] = useState<number[]>([]);
+  const [isLoaderActive, setIsLoaderActive] = useState(false);
+  const [isToggled, setIsToggled] = useState(false);
+
+  const filteredTodos = getFilteredTodos(todoList, filterBy);
+
+  useEffect(() => {
+    postService.getTodos(USER_ID)
+      .then(setTodoList)
+      .catch(() => {
+        setErrorMessage(ErrorMessages.LoadError);
+      });
+    const timerId = setInterval(() => {
+      setErrorMessage(ErrorMessages.NoError);
+    }, 3000);
+
+    return () => {
+      clearInterval(timerId);
+    };
+  }, []);
+
+  const allTodosId = todoList.map(({ id }) => id);
+
+  const completedTodosId = todoList
+    .filter(({ completed }) => completed === true)
+    .map(({ id }) => id);
+
+  const hasCompletedTodosCount = completedTodosId.length !== 0;
+
+  const activeTodosCount = todoList
+    .filter(({ completed }) => completed === false).length;
+
+  const handleDeleteTodo = (todoId: number) => {
+    setLoadingId(prevTodoIds => [...prevTodoIds, todoId]);
+    setIsLoaderActive(true);
+
+    return postService.deleteTodo(todoId)
+      .then(() => {
+        setTodoList(currentList => currentList
+          .filter(todo => todo.id !== todoId));
+      })
+      .catch(() => {
+        setErrorMessage(ErrorMessages.DeleteError);
+      })
+      .finally(() => {
+        setLoadingId([]);
+        setIsLoaderActive(false);
+      });
+  };
+
+  const handleClearCompleted = () => {
+    completedTodosId
+      .forEach((id) => {
+        handleDeleteTodo(id);
+      });
+  };
+
+  const addNewTodo = (todo: Todo) => {
+    setTempTodo({
+      id: 0,
+      userId: USER_ID,
+      title: title.trim(),
+      completed: false,
+    });
+
+    setErrorMessage(ErrorMessages.NoError);
+    setLoadingId([0]);
+    setIsLoaderActive(true);
+
+    postService.createTodo(todo)
+      .then(newTodo => {
+        setTodoList(currentTodos => [...currentTodos, newTodo]);
+        setTitle('');
+        setTimeout(() => {
+          setIsLoaderActive(false);
+          setLoadingId([]);
+        }, 3000);
+      })
+      .catch(() => {
+        setIsLoaderActive(false);
+        setLoadingId([0]);
+        setErrorMessage(ErrorMessages.AddError);
+      })
+      .finally(() => {
+        setTempTodo(null);
+        setLoadingId([]);
+        setIsLoaderActive(false);
+      });
+  };
+
+  const updateTodo = (todoToUpdate: Todo) => {
+    return postService.updateTodo(todoToUpdate)
+      .then(todo => {
+        setTodoList(currentTodos => {
+          const newTodos = [...currentTodos];
+          const index = newTodos.findIndex(({ id }) => id === todo.id);
+
+          newTodos.splice(index, 1, todo);
+
+          return newTodos;
+        });
+      })
+      .catch(() => {
+        setErrorMessage(ErrorMessages.UpdateError);
+      })
+      .finally(() => {
+        setLoadingId([]);
+        setIsLoaderActive(false);
+      });
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!title.trim()) {
+      setErrorMessage(ErrorMessages.EmptyTitleError);
+
+      return;
+    }
+
+    addNewTodo({
+      id: +new Date(),
+      userId: USER_ID,
+      title: title.trim(),
+      completed: false,
+    });
+  };
+
+  const handleComplete = (updatedTodo: Todo) => {
+    setErrorMessage(ErrorMessages.NoError);
+    setLoadingId([updatedTodo.id]);
+    setIsLoaderActive(true);
+
+    const foundTodo = { ...updatedTodo };
+
+    foundTodo.completed = !foundTodo.completed;
+
+    updateTodo(foundTodo);
+  };
+
+  const handleToggleAll = () => {
+    setIsToggled(prev => !prev);
+    setIsLoaderActive(true);
+    setLoadingId([...allTodosId]);
+
+    todoList.map(todo => {
+      const item = { ...todo, completed: !isToggled };
+
+      updateTodo(item);
+
+      return item;
+    });
+  };
+
+  const handleEditTodo = (updatedTodo: Todo) => {
+    setIsLoaderActive(true);
+    setLoadingId([updatedTodo.id]);
+
+    return updateTodo(updatedTodo);
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <Header
+          handleSubmit={handleSubmit}
+          loadingId={loadingId}
+          title={title}
+          onTitleChange={setTitle}
+          isLoaderActive={isLoaderActive}
+          handleToggleAll={handleToggleAll}
+          activeTodosCount={activeTodosCount}
+        />
+
+        <section className="todoapp__main" data-cy="TodoList">
+          {filteredTodos && (
+            filteredTodos.map(todo => (
+              <TodoItem
+                key={todo.id}
+                todo={todo}
+                loadingId={loadingId}
+                isLoaderActive={isLoaderActive}
+                handleComplete={handleComplete}
+                handleEditTodo={handleEditTodo}
+                handleDeleteTodo={handleDeleteTodo}
+              />
+            ))
+          )}
+        </section>
+
+        {tempTodo !== null && (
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+    return () => {
+      clearInterval(timerId);
+    };
+  }, []);
+
+  const allTodosId = todoList.map(({ id }) => id);
+
+  const completedTodosId = todoList
+    .filter(({ completed }) => completed === true)
+    .map(({ id }) => id);
```
roman-mirzoian, line 48: You can do the same thing through `reduce` and walk through the array once instead of twice, which will improve the performance of your code.

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+
+  const hasCompletedTodosCount = completedTodosId.length !== 0;
```
roman-mirzoian, line 50:
```suggestion
  const hasCompletedTodosCount = !!completedTodosId.length;
```

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+                key={todo.id}
+                todo={todo}
+                loadingId={loadingId}
+                isLoaderActive={isLoaderActive}
+                handleComplete={handleComplete}
+                handleEditTodo={handleEditTodo}
+                handleDeleteTodo={handleDeleteTodo}
+              />
+            ))
+          )}
```
roman-mirzoian, line 216:
```suggestion
          {filteredTodos?.map(todo => (
              <TodoItem
                key={todo.id}
                todo={todo}
                loadingId={loadingId}
                isLoaderActive={isLoaderActive}
                handleComplete={handleComplete}
                handleEditTodo={handleEditTodo}
                handleDeleteTodo={handleDeleteTodo}
              />
            ))
          }
```

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+        </section>
+
+        {tempTodo !== null && (
+          <TempTodo tempTodo={tempTodo} />
+        )}
```
roman-mirzoian, line 221:
```suggestion
        {tempTodo && (
          <TempTodo tempTodo={tempTodo} />
        )}
```

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+
+        {todoList.length !== 0 && (
+          <TodoFilter
+            filterBy={filterBy}
+            onFilterChange={setFilterBy}
+            activeTodosCount={activeTodosCount}
+            hasCompletedTodosCount={hasCompletedTodosCount}
+            handleClearCompleted={handleClearCompleted}
+          />
+        )}
```
roman-mirzoian, line 231:
```suggestion
        {!!todoList.length && (
          <TodoFilter
            filterBy={filterBy}
            onFilterChange={setFilterBy}
            activeTodosCount={activeTodosCount}
            hasCompletedTodosCount={hasCompletedTodosCount}
            handleClearCompleted={handleClearCompleted}
          />
        )}
```

src/components/ErrorMessage.tsx
```diff
@@ -0,0 +1,33 @@
...
+  onErrorMessageChange,
+}) => {
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={classNames(
+        'notification is-danger is-light has-text-weight-normal', {
+          hidden: !errorMessage,
+        },
+      )}
```
roman-mirzoian, line 21:
```suggestion
      className={classNames(
        'notification', 'is-danger', 'is-light', 'has-text-weight-normal', {
          hidden: !errorMessage,
        },
      )}
```

src/components/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+  activeTodosCount,
+
+}) => {
+  const inputRef = React.useRef<any>();
+
+  React.useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  });
```
roman-mirzoian, line 30:
```suggestion
  useEffect(() => {
    if (inputRef.current) {
      inputRef.current.focus();
    }
  }, []);
```

src/components/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <header className="todoapp__header">
+      <button
+        data-cy="ToggleAllButton"
+        aria-label="button"
+        type="button"
+        className={classNames('todoapp__toggle-all', {
+          active: activeTodosCount === 0,
+        })}
```
roman-mirzoian, line 40:
```suggestion
        className={classNames('todoapp__toggle-all', {
          active: !activeTodosCount,
        })}
```

src/components/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+      />
+
+      <form onSubmit={handleSubmit}>
+        <input
+          ref={inputRef}
+          data-cy="NewTodoField"
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          disabled={loadingId.length > 0 || isLoaderActive}
```
roman-mirzoian, line 51:
```suggestion
          disabled={!!loadingId.length || isLoaderActive}
```

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,74 @@
...
+  hasCompletedTodosCount,
+  handleClearCompleted,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeTodosCount} items left`}
+      </span>
+
+      {/* Active filter should have a 'selected' class */}
```
roman-mirzoian, line 26: Do not leave unnecessary comments.

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,74 @@
...
+  filterBy,
+  onFilterChange,
+  activeTodosCount,
+  hasCompletedTodosCount,
+  handleClearCompleted,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeTodosCount} items left`}
```
roman-mirzoian, line 23: When there is only one todo there should not be plural in the message:
image

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,74 @@
...
+          className={classNames('filter__link', {
+            selected: filterBy === Status.Completed,
+          })}
+          onClick={() => onFilterChange(Status.Completed)}
+        >
+          Completed
+        </a>
+      </nav>
+
+      {/* don't show this button if there are no completed todos */}
```
roman-mirzoian, line 62: Do not leave unnecessary comments.

src/components/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+  handleSubmit,
+  loadingId,
+  title,
+  onTitleChange,
+  isLoaderActive,
+  handleToggleAll,
+  activeTodosCount,
+
+}) => {
+  const inputRef = React.useRef<any>();
```
roman-mirzoian, line 24:
```suggestion
  const inputRef = useRef<HTMLInputElement | null>(null);
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+  }, [isUpdateFormActive]);
+
+  const handleDoubleClick = (event: React.MouseEvent<HTMLDivElement>) => {
+    if (event.detail === 2) {
+      setIsUpdateFormActive(true);
+    }
+  };
+
+  const submit = async () => {
+    if (editedTitle.trim() !== '') {
```
roman-mirzoian, line 44:
```suggestion
    if (editedTitle.trim()) {
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  const handleCancelEdit = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Escape') {
+      setEditedTitle(title);
+      setIsUpdateFormActive(false);
+    }
+  };
+
+  return (
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
```
roman-mirzoian, line 77: Try to rewrite the code so that it is not necessary to disable the linter

roman-mirzoian (changes requested):

src/App.tsx
```diff
@@ -1,24 +1,239 @@
...
+        setLoadingId([]);
+        setIsLoaderActive(false);
+      });
+  };
+
+  const handleClearCompleted = () => {
+    completedTodosId
+      .forEach((id) => {
+        handleDeleteTodo(id);
+      });
```
BudnikOleksii, line 78: When you need to send a batch of requests it's better to create an array of requests and handle them with `Promise.all` or `Promise.allsettled`

src/App.tsx
```diff
@@ -1,24 +1,239 @@
...
+    setTempTodo({
+      id: 0,
+      userId: USER_ID,
+      title: title.trim(),
+      completed: false,
+    });
+
+    setErrorMessage(ErrorMessages.NoError);
+    setLoadingId([0]);
+    setIsLoaderActive(true);
```
BudnikOleksii, line 91: Why do you need `setIsLoaderActive` if you store todos ids and should show loader if todo id in the current array?
Also, `0` as I understand temporary todo id, so maybe create a constant for this id?

src/App.tsx
```diff
@@ -1,24 +1,239 @@
...
+      .catch(() => {
+        setIsLoaderActive(false);
+        setLoadingId([0]);
+        setErrorMessage(ErrorMessages.AddError);
+      })
+      .finally(() => {
+        setTempTodo(null);
+        setLoadingId([]);
+        setIsLoaderActive(false);
+      });
```
BudnikOleksii, line 111:
```suggestion
    postService.createTodo(todo)
      .then(newTodo => {
        setTodoList(currentTodos => [...currentTodos, newTodo]);
        setTitle('');
      })
      .catch(() => {
        setErrorMessage(ErrorMessages.AddError);
      })
      .finally(() => {
        setTempTodo(null);
        setLoadingId([]);
        setIsLoaderActive(false);
      });
```
I have no idea why in a successful case you remove the loader only after 3 seconds 🤷‍♂️ 
It looks strange, and why do you do `setLoadingId([0])` in case of error, as it will be an infinite spinner for tempTodo,  but in finally block you remove it.

src/App.tsx
```diff
@@ -1,24 +1,239 @@
...
+    return postService.updateTodo(todoToUpdate)
+      .then(todo => {
+        setTodoList(currentTodos => {
+          const newTodos = [...currentTodos];
+          const index = newTodos.findIndex(({ id }) => id === todo.id);
+
+          newTodos.splice(index, 1, todo);
+
+          return newTodos;
+        });
```
BudnikOleksii, line 124: Don't use `findIndex` and `splice`, use `map` instead

src/App.tsx
```diff
@@ -1,24 +1,239 @@
...
+    setLoadingId([...allTodosId]);
+
+    todoList.map(todo => {
+      const item = { ...todo, completed: !isToggled };
+
+      updateTodo(item);
+
+      return item;
+    });
+  };
```
BudnikOleksii, line 176: When a user has uncompleted todos they should be marked as completed, in your app it works in the opposite way.

The flow should look something like this:
1) If you have uncompleted todos, filter only them. If not all todos should be changed.
2) Map todos for update array, for each iteration toggle status, and send the request to API.
3) With `Promise.all` check the response, if all is good set updates, if an error occurs notify the user with an error message
4) In finally block remove loaders from these todos

src/components/TempTodo.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  tempTodo: Todo,
+};
+
+export const TempTodo: React.FC<Props> = ({
```
BudnikOleksii, line 8: You have a `TodoItem` component with the same design, use it.
As in your case if you have any design changes you have to keep in mind and change in two components.

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,76 @@
...
+export const TodoFilter: React.FC<Props> = ({
+  filterBy,
+  onFilterChange,
+  activeTodosCount,
+  hasCompletedTodosCount,
+  handleClearCompleted,
+}) => {
+  const itemsLeftMessage = activeTodosCount === 1
+    ? `${activeTodosCount} item left`
+    : `${activeTodosCount} items left`;
```
BudnikOleksii, line 22:
```suggestion
  const itemsLeftMessage = `${activeTodosCount} ${activeTodosCount === 1 ? 'item' : 'items'} left`;
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  const submit = async () => {
+    if (editedTitle.trim()) {
+      await handleEditTodo({
+        id, title: editedTitle, userId, completed,
+      });
+    } else {
+      await handleDeleteTodo(id);
+    }
```
BudnikOleksii, line 50: API calls with `await` have to be wrapped with `try/catch` block to handle errors

src/utils/functions.ts
```diff
@@ -0,0 +1,19 @@
...
+  const filteredTodos = todosList.filter((todo: Todo) => {
+    switch (filterBy) {
+      case Status.Active:
+        return todo.completed === false;
+
+      case Status.Completed:
+        return todo.completed === true;
+
+      default:
+        return todosList;
```
BudnikOleksii, line 14: In default case `filteredTodos` is an array of `todosList` arrays

src/App.tsx
```diff
@@ -1,24 +1,239 @@
...
+              handleComplete={handleComplete}
+              handleEditTodo={handleEditTodo}
+              handleDeleteTodo={handleDeleteTodo}
+            />
+          ))}
+        </section>
+
+        {tempTodo && (
+          <TempTodo tempTodo={tempTodo} />
+        )}
```
BudnikOleksii, line 220: Create a `TodoList` component.

BudnikOleksii (changes requested):
Good work 👍 
Left some suggestions. Also, let's split logic, don't store all logic in the `App` component.
For example:
- filter logic in the filter component
- toggle all in header as the button in this component
- add new todo in the form component or in the header too
- and so on

The smaller your components, the easier it is to manage and support them.

src/App.tsx
```diff
@@ -1,24 +1,139 @@
...
+  const [loadingId, setLoadingId] = useState<number[]>([]);
+
+  const filteredTodos = getFilteredTodos(todoList, filterBy);
+
+  useEffect(() => {
+    postService.getTodos(USER_ID)
+      .then(setTodoList)
+      .catch(() => {
+        setErrorMessage(ErrorMessages.LoadError);
+        throw new Error();
```
roman-mirzoian, line 29: The error text must be passed to the constructor of the `Error` class, otherwise it makes little sense.
Fix this in all code.

src/utils/functions.ts
```diff
@@ -0,0 +1,19 @@
...
+      case Status.Completed:
+        return todo.completed === true;
+
+      default:
+        return todo;
+    }
+  });
+
+  return filteredTodos;
+}
```
roman-mirzoian, line 19: We need to make a `filter` only if `filterBy` is not an empty string, so it is worth checking this condition first and only then make a `filter`, so as not to make unnecessary traversal of the array.
```suggestion
export function getFilteredTodos(todosList: Todo[], filterBy: Status) {
  if (filterBy === Status.Active) {
    todosList = todosList.filter((todo: Todo) => !todo.completed).
  }
  if (filterBy === Status.Completed) {
    todosList = todosList.filter((todo: Todo) => todo.completed).
  }

  return todosList;
}
```

roman-mirzoian (changes requested):
Almost done!

roman-mirzoian (approved):
Well done 👍

PR #794, opened 2023-09-25
initial commit

cardinal1991:
[DEMO LINK](https://cardinal1991.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,146 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
+import { TodoFilter } from './components/TodoFilter';
+import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm';
+import { TContext, useTodoContext } from './components/TodoContext';
+import { SortTypes, Todo } from './types/Todo';
 
-const USER_ID = 0;
+const USER_ID = 11550;
 
 export const App: React.FC = () => {
+  // const [isDeleting, setIsDeleting] = useState<boolean>(false);
+
+  const {
+    todos,
+    // addTodo,
+    hasError,
+    setHasError,
+    sortType,
+    setSortType,
+    handleToggleAllStatus,
+    // isToggledAll,
+    setIsToggledAll,
+  } = useTodoContext() as TContext;
+
+  const handleSorting = (type: string) => setSortType(type as SortTypes);
+
+  const sortedTodos: {
+    all: Todo[];
+    completed: Todo[];
+    active: Todo[];
+  } = {
+    all: todos || [],
+    completed: (todos || []).filter((todo: Todo) => todo.completed),
+    active: (todos || []).filter((todo: Todo) => !todo.completed),
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+  const isAllCompleted = todos.every((todo) => todo.completed === true);
+  // const isAllNotCompleted = todos.every((todo) => todo.completed === false);
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+  const toggleAll = () => {
+    if (isAllCompleted) {
+      handleToggleAllStatus();
+      setIsToggledAll(true);
+      setTimeout(() => setIsToggledAll(false), 500);
+    }
+  };
+
+  // const deleteCompleted = (todoIds: number[]) => {
+  //   setIsDeleting(true);
+
+  //   return todoIds.map((todoId) => deleteTodo(todoId)
+  //     .then(() => getTodos(USER_ID))
+  //     .then((res) => {
+  //       setTodos(res);
+  //       setIsDeleting(false);
+  //     })
+  //     .catch(() => {
+  //       handleError('Unable to delete a todo');
+  //     });
+  //   )
+  // };
+
+  return (
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {/* this buttons is active only if there are some active todos */}
+          <button
+            type="button"
+            className={cn('todoapp__toggle-all',
+              {
+                // eslint-disable-next-line quote-props
+                'active': (isAllCompleted),
+              })}
+            onClick={toggleAll}
+            data-cy="ToggleAllButton"
+          />
+
+          {/* Add a todo on form submit */}
+          <TodoForm />
+        </header>
+
+        <TodoList todos={sortedTodos[sortType]} />
+
+        {/* Hide the footer if there are no todos */}
+        {(todos.length > 0) && (
+          <footer className="todoapp__footer" data-cy="Footer">
+            <span className="todo-count" data-cy="TodosCounter">
+              {`${sortedTodos.active.length} items left`}
+              {/* items left */}
+            </span>
+
+            {/* Active filter should have a 'selected' class */}
+            <TodoFilter sortType={sortType} handleSort={handleSorting} />
+
+            {/* don't show this button if there are no completed todos */}
+            {(todos.some(todo => todo.completed === true)) && (
+              <button
+                type="button"
+                className="todoapp__clear-completed"
+                data-cy="ClearCompletedButton"
+                // onClick={deleteCompleted((todos.filter(todo.id => todo.completed === true))}
+              >
+                Clear completed
+              </button>
+            )}
+          </footer>
+        )}
+      </div>
+
+      {/* Notification is shown in case of any error */}
+      {/* Add the 'hidden' class to hide the message smoothly */}
+      <div
+        data-cy="ErrorNotification"
+        className={`notification is-danger is-light has-text-weight-normal ${hasError === null ? 'hidden' : ''}`}
+      >
+
+        <button
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={() => setHasError(null)}
+        />
+        {hasError}
+        {/* <br />
+          Title should not be empty
+          <br />
+          Unable to add a todo
+          <br />
+          Unable to delete a todo
+          <br />
+          Unable to update a todo */}
+
+      </div>
+    </div>
   );
 };
--- a/src/components/TodoContext.tsx
+++ b/src/components/TodoContext.tsx
@@ -1,0 +1,150 @@
+import React, {
+  ReactNode, SetStateAction, createContext, useContext, useEffect, useState,
+} from 'react';
+import { editTodo, getTodos } from '../api/todos';
+import { SortTypes, Todo } from '../types/Todo';
+
+export interface TContext {
+  todos: Todo[];
+  setTodos: React.Dispatch<SetStateAction<Todo[]>>;
+  hasError: null | string;
+  handleError: (error: string) => void;
+  sortType: SortTypes;
+  setSortType: React.Dispatch<SetStateAction<SortTypes>>;
+  setHasError: React.Dispatch<SetStateAction<string | null>>;
+  tempTodos: Todo | null;
+  setTempTodos: React.Dispatch<SetStateAction<Todo | null>>;
+  idNew: number | null;
+  setIdNew: React.Dispatch<SetStateAction<number | null>>;
+  handleToggleStatus: (todoId: number) => void ;
+  isToggled: boolean,
+  setIsToggled: React.Dispatch<SetStateAction<boolean>>,
+  handleToggleAllStatus:() => void,
+  isToggledAll: boolean,
+  setIsToggledAll: React.Dispatch<SetStateAction<boolean>>,
+}
+
+// Tworzymy kontekst
+const TodoContext = createContext<TContext | null>(null);
+
+// Tworzymy funkcje dostarczające kontekst do komponentów
+export function useTodoContext() {
+  return useContext(TodoContext);
+}
+
+const USER_ID = 11550;
+
+export function TodoProvider({ children }: { children: ReactNode }) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState<string | null>(null);
+  const [sortType, setSortType] = useState<SortTypes>('all');
+  const [tempTodos, setTempTodos] = useState<Todo | null>(null);
+  const [idNew, setIdNew] = useState<number | null>(null);
+  const [isToggled, setIsToggled] = useState<boolean>(false);
+  const [isToggledAll, setIsToggledAll] = useState<boolean>(false);
+
+  const handleError = (error: string) => {
+    setHasError(error);
+    setTimeout(() => setHasError(null), 3000);
+  };
+
+  useEffect(() => {
+    // GET TODOS
+    getTodos(USER_ID)
+      .then((res) => {
+        setTodos(res);
+      })
+      .catch(() => {
+        handleError('Unable to load todos');
+      });
+  }, []);
+
+  // const addTodo = (title: string) => {
+  //   // Dodaj zadanie do stanu zadań
+  //   setTodos([...todos: any, { title, completed: false }]);
+  // };
+
+  const handleToggleStatus = (todoId: number) => {
+    const updatedTodoIndex = todos
+      .findIndex((toggledtodo) => toggledtodo.id === todoId);
+
+    // test czy znalazł
+    if (updatedTodoIndex !== -1) {
+      const updatedTodo = todos[updatedTodoIndex];
+      // eslint-disable-next-line max-len
+      const updatedTodoWithNewStatus = { ...updatedTodo, completed: !updatedTodo.completed };
+
+      setIsToggled(true);
+
+      editTodo(todoId, updatedTodoWithNewStatus)
+        .then((res) => {
+          // Skopiuj tablicę todos
+          const updatedTodos = [...todos];
+
+          // Zaktualizuj tylko jeden element w tablicy
+          updatedTodos[updatedTodoIndex] = res;
+
+          setTodos(updatedTodos);
+        })
+        .catch((error) => {
+          handleError('Unable to update a todo');
+          // eslint-disable-next-line no-console
+          console.log(error);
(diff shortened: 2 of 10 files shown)


src/components/TodoContext.tsx
choeqq, line 1: I think this file should be in different directory

src/components/TodoContext.tsx
```diff
@@ -0,0 +1,168 @@
...
+      })
+      .catch(() => {
+        handleError('Unable to load todos');
+      });
+  }, []);
+
+  // const addTodo = (title: string) => {
+  //   // Dodaj zadanie do stanu zadań
+  //   setTodos([...todos: any, { title, completed: false }]);
+  // };
```
choeqq, line 75: remove comments

src/components/TodoContext.tsx
choeqq, line 1: you need to type your context corretly, then you won't have to type it manually every time you use it

src/App.tsx
```diff
@@ -1,24 +1,181 @@
...
+    setTodos,
+    hasError,
+    setHasError,
+    handleError,
+    sortType,
+    setSortType,
+    handleToggleAllStatus,
+    // isToggledAll,
+    setIsToggledAll,
+    // isGroupDeleting,
```
choeqq, line 26: cleanup

src/App.tsx
```diff
@@ -1,24 +1,181 @@
...
+    setIsGroupDeleting,
+    titleInputRef,
+  } = useTodoContext() as TContext;
```
choeqq, line 29: check comments in `TodoContext.tsx`

src/App.tsx
```diff
@@ -1,24 +1,181 @@
...
+
+  const handleSorting = (type: string) => setSortType(type as SortTypes);
```
choeqq, line 31:
```suggestion
  const handleSorting = (type: SortTypes) => setSortType(type);
```

src/App.tsx
```diff
@@ -1,24 +1,181 @@
...
+  //     .then((res) => {
+  //       setTodos(res);
+  //     })
+  //     .catch(() => {
+  //       handleError('Unable to delete a todo');
+  //     })
+  //     .finally(() => {
+  //       setIsGroupDeleting(false);
+  //     });
+  // };
```
choeqq, line 102: check comments in `TodoContext.tsx`

src/App.tsx
```diff
@@ -1,24 +1,181 @@
...
+        {hasError}
+        {/* <br />
+          Title should not be empty
+          <br />
+          Unable to add a todo
+          <br />
+          Unable to delete a todo
+          <br />
+          Unable to update a todo */}
+
```
choeqq, line 177: cleanup

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,59 @@
...
+  sortType: SortTypes;
+};
+
+const types = ['All', 'Active', 'Completed'];
+
+export const TodoFilter: React.FC<Props> = ({ handleSort, sortType }) => {
+  // const {
+  //   sortType,
+  //   // setSortType,
+  // } = useTodoContext() as TContext;
```
choeqq, line 16: cleanup

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  //   <a
+  //     href="#/completed"
+  //     className="filter__link"
+  //     data-cy="FilterLinkCompleted"
+  //   >
+  //     Completed
+  //   </a>
+  // </nav>
+  );
```
choeqq, line 58: cleanup

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export const TodoFilter: React.FC<Props> = ({ handleSort, sortType }) => {
+  // const {
+  //   sortType,
+  //   // setSortType,
+  // } = useTodoContext() as TContext;
+
+  return (
+    <nav className="filter" data-cy="Filter">
+      {types.map((type:string) => (
```
choeqq, line 20: prettier

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+export const TodoForm: React.FC = () => {
+  const [title, setTitle] = useState<string>('');
+  const [isSubmitting, setIsSubmitting] = useState<boolean>(false);
+  const {
+    todos,
+    setTodos,
+    handleError,
+    setTempTodos,
+    titleInputRef,
+  } = useTodoContext() as TContext;
```
choeqq, line 17: check comments in `TodoContext.tsx`

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const counter = useMemo(() => {
+    return Math.max(...todos.map(todo => todo.id)) + 1;
+  }, [todos]);
+
+  useEffect(() => {
+    if (!isSubmitting) {
+      titleInputRef.current?.focus(); // ustaw fokus na input po dodaniu
+    }
+  }, [!isSubmitting]);
```
choeqq, line 27:
```suggestion
  }, [isSubmitting]);
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,225 @@
...
+  const [newTitle, setNewTitle] = useState(todo.title);
+  const [isLoading, setIsLoading] = useState(false);
+  const [editedTodo, setIsEditedTodo] = useState<Todo | null>(null);
+
+  const {
+    todos,
+    setTodos,
+    handleError,
+    // tempTodos,
+    // idTemp,
```
choeqq, line 24: cleanup

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,225 @@
...
+  //       setTodos(res);
+  //       setIsDeleting(false);
+  //     })
+  //     .catch(() => {
+  //       handleError('Unable to delete a todo');
+  //     })
+  //     .finally(() => {
+  //       titleInputRef.current?.focus();
+  //     });
+  // };
```
choeqq, line 82: cleanup

src/components/TodoList.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+// import cn from 'classnames';
```
choeqq, line 2: cleanup

src/components/TodoList.tsx
```diff
@@ -0,0 +1,75 @@
...
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  const {
+    // todos,
+    // setTodos,
+    // handleError,
+    tempTodos,
+    // idTemp,
```
choeqq, line 17: cleanup

src/components/TodoList.tsx
```diff
@@ -0,0 +1,75 @@
...
+  } = useTodoContext() as TContext;
```
choeqq, line 18: check comments in `TodoContext.tsx`

src/components/TodoList.tsx
```diff
@@ -0,0 +1,75 @@
...
+          >
+            ×
+          </button>
+
+          <div
+            data-cy="TodoLoader"
+            className={cn('modal overlay')}
+            // { 'is-active': tempTodos.id === 0 }
+
+          >
```
choeqq, line 67: cleanup

choeqq (changes requested):
PLEASE MAKE SURE TO REMOVE EVERY COMMENT BEFORE PUSHING YOUR CODE

also we want to use comments only we unexpected behaviour occurs 😉

Matylda-Lewandowska (approved):
I have nothing to add, code looks nice to me :)

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+export function useTodoContext() {
+  return useContext(TodoContext);
+}
+
+const USER_ID = 11550;
+
+export function TodoProvider({ children }: { children: ReactNode }) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState<string | null>(null);
```
pawel-peksa, line 42: Let's narrow down the type of this state to union of strings which we are actually using there + null

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,149 @@
...
+  createContext, useContext, useEffect, useRef, useState,
+} from 'react';
+import { editTodo, getTodos } from '../api/todos';
+import { SortTypes, Todo } from '../types/Todo';
+
+export interface TContext {
+  todos: Todo[];
+  setTodos: React.Dispatch<SetStateAction<Todo[]>>;
+  hasError: null | string;
+  handleError: (error: string) => void;
```
pawel-peksa, line 14: Please see the comment related to types in `const [hasError, setHasError]`.
Instead of `string` let's use a string union

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,197 @@
...
+          <form
+            onSubmit={(e) => handleSubmit(e)}
+          >
+            <input
+              data-cy="TodoTitleField"
+              type="text"
+              className="todo__title-field"
+              placeholder="Empty todo will be deleted"
+              value={newTitle}
+              // defaultValue={todo.title}
```
pawel-peksa, line 136: Cleanup needed

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,197 @@
...
+      )}
+
+      <div
+        data-cy="TodoLoader"
+        className={cn('modal overlay',
+          {
+            'is-active': (isDeleting === true)
+          || (todo.id === 0) || (isToggled && todo.id === toggledId)
+          || (isToggledAll) || (isLoading)
+          || ((todo.completed === true) && isGroupDeleting),
```
pawel-peksa, line 188: There is no need to check booleans against true.
```suggestion
          || (todo.completed && isGroupDeleting),
```

pawel-peksa, line 188: The whole logic here looks really complicated, It's hard to read and probably does not work as intended. Try to refactor this code. If the logic needs to be so complicated then extract it to function outside `return` and just call it here.

src/App.tsx
```diff
@@ -1,24 +1,146 @@
...
+    completed: (todos || []).filter((todo: Todo) => todo.completed),
+    active: (todos || []).filter((todo: Todo) => !todo.completed),
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const isAllCompleted = todos.every((todo) => todo.completed === true);
+  // const isAllNotCompleted = todos.every((todo) => todo.completed === false);
```
pawel-peksa, line 46: Cleanup needed

src/App.tsx
```diff
@@ -1,24 +1,146 @@
...
+      setIsToggledAll(true);
+      setTimeout(() => setIsToggledAll(false), 500);
+    }
+
+    // if (isAllNotCompleted) {
+    //   handleToggleAllStatus();
+    //   setIsToggledAll(true);
+    //   setTimeout(() => setIsToggledAll(false), 500);
+    // }
+  };
```
pawel-peksa, line 60: Cleanup needed

src/App.tsx
```diff
@@ -1,24 +1,146 @@
...
+
+  const arrayCompleted = [...todos].filter((todo) => todo.completed === true);
```
pawel-peksa, line 62:
```suggestion
  const arrayCompleted = [...todos].filter((todo) => todo.completed);
```

src/App.tsx
```diff
@@ -1,24 +1,146 @@
...
+    all: todos || [],
+    completed: (todos || []).filter((todo: Todo) => todo.completed),
+    active: (todos || []).filter((todo: Todo) => !todo.completed),
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const isAllCompleted = todos.every((todo) => todo.completed === true);
```
pawel-peksa, line 45:
```suggestion
  const isAllCompleted = todos.every((todo) => todo.completed);
```

src/App.tsx
pawel-peksa, line 62: Same logic as in `sortedTodos.completed`. Try to not repeat your code.

src/App.tsx
```diff
@@ -1,24 +1,146 @@
...
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {(todos.length > 0) && (
+            <button
+              type="button"
+              className={cn('todoapp__toggle-all',
+                {
+                // eslint-disable-next-line quote-props
+                  'active': (isAllCompleted),
```
pawel-peksa, line 99: No need for quotes here

src/App.tsx
```diff
@@ -1,24 +1,146 @@
...
+              {`${sortedTodos.active.length} items left`}
+            </span>
+            <TodoFilter sortType={sortType} handleSort={handleSorting} />
+
+            <button
+              type="button"
+              className="todoapp__clear-completed"
+              data-cy="ClearCompletedButton"
+              onClick={deleteCompleted}
+              disabled={!(todos.some(todo => todo.completed === true))}
```
pawel-peksa, line 122:
```suggestion
              disabled={!(todos.some(todo => todo.completed))}
```

pawel-peksa (changes requested):
Almost there ;)

src/App.tsx
```diff
@@ -1,24 +1,145 @@
...
+    }
+
+    if (isAllNotCompleted) {
+      handleToggleAllStatus();
+      setIsToggledAll(true);
+      setTimeout(() => setIsToggledAll(false), 500);
+    }
+  };
+
+  // const arrayCompleted = [...todos].filter((todo) => todo.completed);
```
pawel-peksa, line 62: cleanup needed

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,198 @@
...
+    }
+  }, [isEditing]);
+
+  const USER_ID = 11550;
+
+  const isActiveConditions = (isDeleting === true)
+  || (todo.id === 0) || (isToggled && todo.id === toggledId)
+  || (isToggledAll) || (isLoading)
+  || ((todo.completed) && isGroupDeleting);
+
```
pawel-peksa, line 43: It's good that you extracted it outside JSX, however this logic still causes some of the tests to fail. Try to fix it a bit. It should not be dependent on so many cases.

src/App.tsx
```diff
@@ -1,24 +1,145 @@
...
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {(todos.length > 0) && (
+            <button
+              type="button"
+              className={cn('todoapp__toggle-all',
+                {
+                  active: (isAllCompleted || isAllNotCompleted),
```
pawel-peksa, line 98: I think you want this class to appear only when isAllCompleted is true.

pawel-peksa (changes requested):
Great job 🔥 This task is really complicated and the code looks much better. There is, however, still a possiblity to fix many of the failing tests. Try to focus on setting 'is-active' class basing on specific states. Currently it looks like there is too much going on. First, try to write down on paper all the cases when this class needs to be applied and then reflect it in `isActiveConditions`

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,76 @@
...
+        })
+        .catch((error) => {
+          handleError('Unable to add a todo');
+          // eslint-disable-next-line no-console
+          console.log(error);
+        })
+        .finally(() => {
+          setIsSubmitting(false);
+          setTempTodos(null);
+          counter += 1;
```
mbruhler, line 56: In overall this counter has no effect because is always null. If you want to get next ID then you should check it against the last item in the todos collection.

mbruhler: Actually as I can see you don't have to pass ID to the `POST`

Please refer to this documentation link: https://mate-academy.github.io/fe-students-api/todos.html

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,149 @@
...
+      .findIndex((toggledtodo) => toggledtodo.id === todoId);
+
+    if (updatedTodoIndex !== -1) {
+      const updatedTodo = todos[updatedTodoIndex];
+      const updatedTodoWithNewStatus
+      = { ...updatedTodo, completed: !updatedTodo.completed };
+
+      setIsToggled(true);
+
+      editTodo(todoId, updatedTodoWithNewStatus)
```
mbruhler, line 77: You don't have to pass the whole `Todo` to the `PATCH`.
It's always good to transfer at least amount of data between backend and frontend.

Here we are only interested to set the completed status to the opposite, so the payload should be 
```
{
    completed: !updatedTodo.completed
}
```

Please refer to the API documentation for further information: https://mate-academy.github.io/fe-students-api/todos.html

mbruhler (changes requested):
Good job, you are almost there!

Just a few fixes in your API handling.
Remember that:
- In `PATCH` request, you don't have to (and you shouldn't) pass whole object to update, just partial data that needs the update
- Always read the backend API documentation and align your API handling to it. For example, you don't have to pass ID to the `POST` request. If there are typing issues, think of using `Partial`, `Pick`, `Omit` typescript utility types which helps to have one type and it's transformations.

In your case I would go with creating types like that:
- Todo - which contains the response of Todo from API
- PatchTodo - which is a Partial type of Todo - `Partial<Todo>`. This makes all the properties optional,
- AddTodo - which has everything but `id`. So you can use `Omit` here like `Omit<Todo, 'id'>`
- others accordingly if needed

More on this here: https://www.typescriptlang.org/docs/handbook/utility-types.html

Of course you don't have to use those directives, you can create just another types.

My suggestion may be an overkill for now, but I can see that you do really good and you have much potential so why not trying something extra! If that's over the top, next person propably will accept this PR :)

src/components/TodoForm.tsx
```diff
@@ -43,6 +40,8 @@ export const TodoForm: React.FC = () => {
       addTodo(newTodo)
         .then((res) => {
           setTodos([...todos, res]);
+          // eslint-disable-next-line no-console
```
mbruhler, line 43: clear console.log

src/App.tsx
```diff
@@ -7,23 +7,21 @@ import { TodoList } from './components/TodoList';
 import { TodoForm } from './components/TodoForm';
 import { TContext, useTodoContext } from './context/TodoContext';
 import { SortTypes, Todo } from './types/Todo';
-import { deleteTodo } from './api/todos';
+// import { deleteTodo } from './api/todos';
```
mbruhler, line 10: clear comment

mbruhler (changes requested):
Cleanup the code and I'll accept it. Check also if there are no not needed `console.log` and `//` comments

choeqq (approved):
LGTM ✅

PR #796, opened 2023-09-26
implemented logic for updating todo

irahos:


yuliiaant (approved):
Nice work!

src/App.tsx
```diff
@@ -1,24 +1,260 @@
...
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterCase, setFilterCase] = useState(FilterCase.all);
+  const [error, setError] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const newTodoField = useRef<HTMLInputElement>(null);
```
roman-mirzoian, line 21:
```suggestion
  const newTodoFieldRef = useRef<HTMLInputElement>(null);
```

src/App.tsx
```diff
@@ -1,24 +1,260 @@
...
+    setError('');
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setError('Unable to load todos');
+      });
+  }, []);
+
+  const timerId = useRef<number>(0);
```
roman-mirzoian, line 39:
```suggestion
  const timerIdRef = useRef<number>(0);
```

src/App.tsx
```diff
@@ -1,24 +1,260 @@
...
+
+  useEffect(() => {
+    if (timerId.current) {
+      window.clearTimeout(timerId.current);
+    }
+
+    timerId.current = window.setTimeout(() => {
+      setError('');
+    }, 3000);
```
roman-mirzoian, line 48: https://upmostly.com/tutorials/settimeout-in-react-components-using-hooks

src/App.tsx
```diff
@@ -1,24 +1,260 @@
...
+
+    return addTodo(todoTitle)
+      .then((newTodo) => {
+        setTodos((prevTodos) => [...prevTodos, newTodo]);
+
+        newTodoField.current?.focus();
+      })
+      .catch(() => {
+        setError('Unable to add a todo');
+        throw new Error();
```
roman-mirzoian, line 69: Error text must be added to the `Error` class, otherwise it has no meaning.

src/App.tsx
```diff
@@ -1,24 +1,260 @@
...
+      })
+      .finally(() => {
+        setTempTodo(null);
+        setIsLoading(false);
+      });
+  };
+
+  const handleDeleteTodo = (todoId: number) => {
+    setLoadingTodoIds((prevLoadingTodoIds) => [...prevLoadingTodoIds, todoId]);
```
roman-mirzoian, line 78:
```suggestion
    setLoadingTodoIds((prevIds) => [...prevIds, todoId]);
```

src/App.tsx
```diff
@@ -1,24 +1,260 @@
...
+        )));
+      })
+      .catch(() => {
+        setError('Unable to update a todo');
+      })
+      .finally(() => {
+        setLoadingTodoIds((prevTodoIds) => prevTodoIds
+          .filter((id) => id !== todo.id));
+        setIsLoading(false);
+      });
```
roman-mirzoian, line 118: You use similar logic to update the status there, so it can be a single handler that takes as an argument the data to be updated.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+  setFilterCase,
+  activeTodos,
+  handleClearCompleted,
+}) => {
+  const completedTodos = todos.filter(({ completed }) => completed);
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeTodos.length} items left`}
```
roman-mirzoian, line 25: If there is one item, there should not be a plural in the text.

src/App.tsx
```diff
@@ -1,24 +1,260 @@
...
+    }
+  }, [isLoading]);
+
+  useEffect(() => {
+    setError('');
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setError('Unable to load todos');
```
roman-mirzoian, line 35: It would be good to put error texts in a separate constant.

roman-mirzoian (changes requested):
Also, there is no link to the demo in the PR description.

src/App.tsx
```diff
@@ -1,24 +1,249 @@
...
+            filterCase={filterCase}
+            setFilterCase={setFilterCase}
+            activeTodos={activeTodos}
+            handleClearCompleted={handleClearCompleted}
+          />
+        )}
+      </div>
+
+      <div
+        data-cy="ErrorNotification"
```
BudnikOleksii, line 228: Create an `ErrorNotification` component

src/App.tsx
```diff
@@ -1,24 +1,249 @@
...
+      });
+  };
+
+  const toggleAllTodos = async () => {
+    setLoadingTodoIds(todos.map((todo) => todo.id));
+
+    try {
+      const areAllCompleted = todos.every((todo) => todo.completed);
+
+      const updatedTodos = await Promise.all(
```
BudnikOleksii, line 134: If not all completed update only todos that aren't completed.
It doesn't make sense to set the status `completed` for todos that are already completed

BudnikOleksii (changes requested):
Good work 👍 
Some features that you should implement:
- if not all are completed, update only todos that aren't completed.
- cancel editing on Esс key keyup event;

Moroz-Dmytro (changes requested):
1. Double click doesn't work as expected
image

2. 
image

3. When I click `Toggle All` -> I should see loader only on todos that should be changed(not all)
image

4. I should not be able to rename todo to empty(enter just some spaces)
image

PR #799, opened 2023-09-26
Solution

Sancz0pansa:
[DEMO LINK](https://Sancz0pansa.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,252 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/no-autofocus */
+/* eslint-disable no-useless-return */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  FormEventHandler, useEffect, useRef, useState,
+} from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
+import {
+  addTodo, updateTodo, deleteTodo, getTodos,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
+import { ErrorMess } from './types/Error';
+import { TodoList } from './components/TodoList';
 
-const USER_ID = 0;
+const USER_ID = 10521;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<ErrorMess>(null);
+  const [filter, setFilter] = useState<Filter>('All');
+  const [title, setTitle] = useState<string>('');
+  const [temporaryTodo, setTemporaryTodo] = useState<Todo | null>(null);
+  const [isSubmiting, setIsSubmiting] = useState<boolean>(false);
+
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  const handleError = (mess: ErrorMess) => {
+    setError(mess);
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,24 +1,274 @@
-/* eslint-disable max-len */
+/* eslint-disable no-plusplus */
+/* eslint-disable no-param-reassign */
+/* eslint-disable jsx-a11y/no-autofocus */
+/* eslint-disable no-useless-return */
```
choeqq, line 4: I don't you should have some many `eslint-disable` in your code - at some point it will stop to make sense to even use it

src/App.tsx
choeqq, line 1: I think most of the code belongs to global state - this should not be in App.tsx file

src/App.tsx
```diff
@@ -1,24 +1,272 @@
...
+            temporaryTodo={temporaryTodo}
+            handleDelete={handleDelete}
+            handleComplete={handleComplete}
+            handleError={handleError}
+          />
+        )}
+
+        {/* Hide the footer if there are no todos */}
+        {(todos.length > 0 || temporaryTodo) && (
+          <footer className="todoapp__footer" data-cy="Footer">
```
choeqq, line 195: I think this should be a separate component

src/App.tsx
```diff
@@ -1,24 +1,272 @@
...
+              type="text"
+              className="todoapp__new-todo"
+              placeholder="What needs to be done?"
+              value={title}
+              onChange={(e) => setTitle(e.target.value)}
+              autoFocus
+              disabled={isSubmiting}
+              ref={inputRef}
+            />
+          </form>
```
choeqq, line 179: this should also be a separate component IMO

src/components/TempTodo.tsx
```diff
@@ -0,0 +1,50 @@
...
+type Props = {
+  temporaryTodo: Todo | null;
+};
+
+export const TempTodo = ({ temporaryTodo }: Props) => {
+  if (temporaryTodo === null) {
+    return (
+      <div />
+    );
+  }
```
choeqq, line 13: why exactly do you return empty div when there is no temporaryTodo?

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,145 @@
+/* eslint-disable no-param-reassign */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
```
choeqq, line 3: as in App.tsx ⬆️

src/components/TodoList.tsx
```diff
@@ -0,0 +1,51 @@
...
+    default:
+  }
+
+  return filteredTodos;
+};
+
+export const TodoList = ({
+  todos, filter, temporaryTodo, handleDelete, handleComplete, handleError,
+} : Props) => {
+  return (
```
choeqq, line 36:
```suggestion
} : Props) => (
```

choeqq (changes requested):

Sancz0pansa:
I had a Q&A and we didn't figure out why the tests weren't passing.

ogospadzipalina (approved):
looks  good to me

choeqq (approved):
LGTM ✅

PR #803, opened 2023-09-26
almost Fully functional Todo App🎀

polinavafik:
[DEMO LINK](https://polinavafik.github.io/react_todo-app-with-api/)  🎀

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # React Todo App with API
 [DEMO LINK](https://polinavafik.github.io/react_todo-app-with-api/)
 
-Todo App is a simple tool made with React for managing your to-do list. It's built with reusable components, making it highly modular and maintainable. It uses useContext for state management and adds a touch of elegance with smooth animations via <TransitionGroup>
+Todo App is a simple tool made with React for managing your to-do list. It's built with reusable components, making it highly modular and maintainable. It uses useContext for state management and adds a touch of elegance with smooth animations via TransitionGroup
 
 Key Features:
 ✧ API Integration: Load, update, delete, and add todos seamlessly from an external API, ensuring your task list is always up-to-date.


voronine (approved):
Ehh, good work, but a lot of test errors. Go to QN, they helped me yesterday.

src/App.tsx
```diff
@@ -1,24 +1,99 @@
...
-    return <UserWarning />;
-  }
+  const [todoFilter, setTodoFilter] = useState<TodoFilter>(TodoFilter.All);
+
+  const {
+    todos,
+    setTodos,
+    error,
+    setError,
+  } = useContext(TodoContext);
```
BudnikOleksii, line 29: You can create hooks for contexts so you don't have to pass context every time.
```javascript
const useTodo = () => useContext(TodoContext);
```

src/App.tsx
```diff
@@ -1,24 +1,99 @@
...
+
+  useEffect(() => {
+    todoService.getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        setError(CurrentError.LoadingError);
+        throw new Error();
```
BudnikOleksii, line 36: You don't need to throw an error in this case, as in this case, you don't catch this error
image

src/Context/TodoContext.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+    todoService.deleteTodo(todoId)
+      .then(() => {
+        setTodos(prevTodos => {
+          return prevTodos.filter(todo => todo.id !== todoId);
+        });
+      })
+      .catch(() => {
+        setError(CurrentError.DeleteError);
+        throw new Error();
```
BudnikOleksii, line 80: The same, just set an error to notify user, it's enough in this case

src/Context/TodoContext.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const handleTodoRename = (todo: Todo, newTodoTitle: string) => {
+    if (todo.title === newTodoTitle) {
+      return Promise.resolve();
+    }
+
+    if (!newTodoTitle.trim()) {
+      setError(CurrentError.EmptyTitleError);
+
+      return Promise.reject();
```
BudnikOleksii, line 115: You don't need to reject or resolve in this cases, just return in case you don't need to proceed

src/Context/TodoContext.tsx
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    setIsLoading(true);
+    setProcessingTodoIds(prevState => [...prevState, todo.id]);
+
+    // eslint-disable-next-line consistent-return
```
BudnikOleksii, line 121: Don't disable eslint, I think in this case you  actually don't need to return

src/components/TodoElement.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+      if (error === CurrentError.DeleteError) {
+        setError(CurrentError.DeleteError);
+        throw new Error();
+      }
+
+      if (error === CurrentError.EmptyTitleError) {
+        setError(CurrentError.EmptyTitleError);
+        throw new Error();
+      }
```
BudnikOleksii, line 64: You don't need this as you set errors previously in you handlers, so you don't need to throw errors and handle them here one more time

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,108 @@
...
+    isLoading,
+    handleToggleChange,
+  } = useContext(TodoContext);
+  const [title, setTitle] = useState('');
+
+  const activeTodosCount = activeTodos.length;
+  const inputField = useRef<HTMLInputElement>(null);
+
+  const isAllCompleted = useMemo(() => (
+    todos.every(el => el.completed)
```
BudnikOleksii, line 33:
```suggestion
    todos.every(el => todo.completed)
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,44 @@
...
+type Props = {
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  const { tempTodo } = useContext(TodoContext);
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      <TransitionGroup>
```
BudnikOleksii, line 16: 🔥

src/components/TodoList.tsx
```diff
@@ -0,0 +1,44 @@
...
+        {todos?.map(todo => (
```
BudnikOleksii, line 17: `todos` is required prop in accordance with type

src/utils/getFilteredTodos.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const getFilteredTodos = (
+  todos: Todo[],
+  selectedFilter: string,
+) => {
+  let preparedTodos = [...todos];
+
+  if (selectedFilter !== TodoFilter.All) {
+    preparedTodos = preparedTodos.filter(({ completed }) => {
+      switch (selectedFilter) {
```
BudnikOleksii, line 12:
```suggestion
  if (selectedFilter !== TodoFilter.All) {
    return todos.filter(({ completed }) => {
      switch (selectedFilter) {
```
`filter` returns new array so you don't need create a copy

BudnikOleksii (changes requested):
Looks great 🔥 
Left some comments to improve your solution

src/Context/TodoContext.tsx
polinavafik: I had a problem without resolve reject since I have typization  ` handleTodoRename: (todo: Todo, newTodoTitle: string) => Promise<void>` 

but I added ` | void`,  to it and now it passes 😅 how I didn't think of that before

src/Context/TodoContext.tsx
polinavafik: If I don't have return, then when you edit todo, after edit on load it shows the previous version of the title and then shows the new one (after load)

but with return, everything works as it should...

src/utils/getFilteredTodos.tsx
polinavafik: 😅ooopsss

src/App.tsx
```diff
@@ -1,24 +1,97 @@
...
+        )}
+      >
+        <button
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={() => setError(CurrentError.Default)}
+        />
+        {error}
+      </div>
```
BudnikOleksii, line 94: Create a component `ErrorNotification` and move the logic for errors to it too

BudnikOleksii (approved):
Good work 🚀

PR #811, opened 2023-09-27
add task solution

desevoker:
[DEMO LINK](https://desevoker.github.io/react_todo-app-with-api/)

As far as I can go. Two tests are failing no matter what. The third one is more interesting, look at the screenshots.
I suppose that the issue is in the tests.
image
image


src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+    clearErrorMessage();
+    startProcessingTodo(todo);
+
+    if (!title) {
+      return todoService.deleteTodo(todo.id)
+        .then(() => removeTodoFromState(todo))
+        .catch(() => {
+          setErrorMessage(ErrorMessage.DeleteTodo);
+
+          throw new Error(ErrorMessage.DeleteTodo);
```
Moroz-Dmytro, line 143: Why you need to throw new error here?

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+        })
+        .finally(() => stopProcessingTodo(todo));
+    }
+
+    return todoService.editTodo(todo.id, title)
+      .then(replaceTodoInState)
+      .catch(() => {
+        setErrorMessage(ErrorMessage.UpdateTodo);
+
+        throw new Error(ErrorMessage.UpdateTodo);
```
Moroz-Dmytro, line 153: Same here

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+      case Status.Active:
+        return !completed;
+
+      case Status.Completed:
+        return completed;
+
+      case Status.All:
+      default:
+        return true;
+    }
```
Moroz-Dmytro, line 175: I'd move this to separate helper function

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+              active={isAllCompleted}
+              onToggleAll={handleToggleAllTodos}
+            />
+          )}
+
+          <NewTodo
+            refocus={totalTodoCount}
+            onAdd={handleAddTodo}
+            onError={setErrorMessage}
+          />
```
Moroz-Dmytro, line 195: Why not to move this inside `Header` component?

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+                  onDelete={handleDeleteTodo}
+                  onToggle={handleToggleTodo}
+                  onEdit={handleEditTodo}
+                  processing={processingIds.includes(todo.id)}
+                />
+              ))}
+
+              {tempTodo && (
+                <TodoItem todo={tempTodo} processing />
+              )}
```
Moroz-Dmytro, line 214: Why not to move this inside `TodoList`?

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+
+              <TodoFilter
+                value={filterValue}
+                onValueChange={setFilterValue}
+              />
+
+              <ClearCompletedButton
+                active={hasCompletedTodos}
+                onClear={handleClearCompletedTodos}
+              />
```
Moroz-Dmytro, line 228: Why not to move this inside `Footer`?

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+      'has-text-weight-normal',
+      { hidden: !message },
+    )}
+  >
+    <button
+      type="button"
+      className="delete"
+      aria-label="Hide Error"
+      data-cy="HideErrorButton"
+      onClick={() => onHide()}
```
Moroz-Dmytro, line 30:
```suggestion
      onClick={onHide}
```

src/components/NewTodo.tsx
```diff
@@ -0,0 +1,64 @@
...
+      onError(ErrorMessage.EmptyTitle);
+
+      return;
+    }
+
+    setIsProcessing(true);
+
+    onAdd(title)
+      .then(() => setInputValue(''))
+      .catch(() => {})
```
Moroz-Dmytro, line 46: Why do nothing in `catch`?

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+  const handleToggleAllTodos = useCallback(() => {
+    const hasActive = todos.some(({ completed }) => !completed);
+
+    todos.forEach(todo => {
+      if (hasActive && todo.completed) {
+        return;
+      }
+
+      handleToggleTodo(todo);
+    });
```
Moroz-Dmytro, line 130: Let's use `Promise.all`

src/App.tsx
```diff
@@ -1,24 +1,240 @@
...
+    todoService.deleteTodo(todo.id)
+      .then(() => removeTodoFromState(todo))
+      .catch(() => setErrorMessage(ErrorMessage.DeleteTodo))
+      .finally(() => stopProcessingTodo(todo));
+  }, []);
+
+  const handleClearCompletedTodos = useCallback(() => {
+    todos
+      .filter(({ completed }) => completed)
+      .forEach(handleDeleteTodo);
```
Moroz-Dmytro, line 108: Let's use `Promise.all`

Moroz-Dmytro (changes requested):
Good work 🔥 
UI works as expected 👍 
Only some comments below

src/App.tsx
desevoker: It is inside `Header` component. `Header` was implemented as a simple wrapper because it doesn't need any logic on its own and I don't want to pass a bunch of props to child components through it.

src/App.tsx
desevoker: It is inside `TodoList` component. Same explanation as above.

src/App.tsx
desevoker: It is inside `Footer`. Same explanation as above.

src/App.tsx
desevoker: To catch it in child component and do something there too.

src/App.tsx
desevoker: Same explanation as above.

src/App.tsx
```diff
@@ -1,24 +1,209 @@
...
+  const handleEditTodo = useCallback((todo: Todo, title: string) => {
+    clearErrorMessage();
+    startProcessingTodo(todo);
+
+    return todoService.editTodo(todo.id, title)
+      .then(replaceTodoInState)
+      .finally(() => stopProcessingTodo(todo));
+  }, []);
+
+  const hasTodos = todos.length > 0 || !!tempTodo;
```
roman-mirzoian, line 135:
```suggestion
  const hasTodos = !!todos.length || !!tempTodo;
```

src/App.tsx
roman-mirzoian: @desevoker ok, if you don't want to expose it as a separate component to avoid doing a lot of code, then why expose as a wrapper one tag that can be used directly?
now this is an unnecessary increase in code and dependencies, and it is better then to use ordinary tags

src/components/TodoCounter.tsx
```diff
@@ -0,0 +1,11 @@
+import { memo } from 'react';
+
+type Props = {
+  value: number;
+};
+
+export const TodoCounter:React.FC<Props> = memo(({ value }) => (
+  <span className="todo-count" data-cy="TodosCounter">
+    {`${value} items left`}
```
roman-mirzoian, line 9: If there is one item, there should not be a plural in the text.

roman-mirzoian (changes requested):
Almost done.

src/components/TodoCounter.tsx
desevoker: Test thinks differently, and will fail on singular `item` text.

src/App.tsx
desevoker: Ok I get it. But it is still a valid solution, just a different one.

roman-mirzoian (approved):
Well done 👍

PR #813, opened 2023-09-27
add working version

dpidlutska:
work in progress...
[DEMO LINK](https://dpidlutska.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,97 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, {
+  useContext,
+  useEffect, useMemo, useRef, useState,
+} from 'react';
+import cn from 'classnames';
+import { TodoForm } from './components/TodoForm';
+import { TodoList } from './components/TodoList';
+import { TodoFilter } from './components/TodoFilter';
+import * as todoService from './api/todos';
+import { TodosFilter } from './types/TodosFilter';
+import { getFilteredTodos } from './utils/getFilteredTodos';
+import { TodoContext } from './Context/TodoContext';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const {
+    todos,
+    setTodos,
+    errorMessage,
+    setErrorMessage,
+  } = useContext(TodoContext);
+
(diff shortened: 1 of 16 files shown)


Yevheniia-Sid (approved):
Looks good to me👍

src/App.tsx
```diff
@@ -1,24 +1,97 @@
...
+        {/* Unable to load todos
+        <br />
+        Title should not be empty
+        <br />
+        Unable to add a todo
+        <br />
+        Unable to delete a todo
+        <br />
+        Unable to update a todo */}
+      </div>
```
BudnikOleksii, line 94: Remove comments, create a separate `ErrorNotification` component, and move related logic and `useEffect` to it

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,91 @@
...
+    case TodosFilter.All:
+    default:
+      return '#/';
+  }
+};
+
+const setFilterDataCy = (filter: TodosFilter) => {
+  switch (filter) {
+    case TodosFilter.Active:
+      return 'FilterLinkActive';
```
BudnikOleksii, line 23: You do the same checks so you can return an object and one function is enough in this case.

```javascript
    case TodosFilter.Active:
      return {
        href: '#/active',
        dataCy: 'FilterLinkActive',
      };
```

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,91 @@
...
+    uncompletedTodos,
+
+    handleClearCompletedTodos,
+  } = useContext(TodoContext);
+  const uncompletedTodosAmount = uncompletedTodos.length;
+  const hasCompletedTodos = !!completedTodos.length;
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      {/* Hide the footer if there are no todos */}
```
BudnikOleksii, line 54: Remove all comments

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    setIsAdding(true);
+
+    handleAddTodo(preparedTitle)
+      .then(() => {
+        setTitle('');
+      })
+      .catch(() => {
+
+      })
```
BudnikOleksii, line 58: Why `catch` block is empty? Notify a user in case of error

BudnikOleksii (changes requested):
Looks great 🔥 
Just get rid of comments and take a look at suggestions.

alexander-ignatow (approved):

PR #821, opened 2023-09-27
Develop

Yevheniia-Sid:
[DEMO LINK](https://Yevheniia-Sid.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,7 +3,6 @@ import React, { useEffect, useRef, useState } from 'react';
 import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
 import { Todo } from './types/Todo';
-// import { addTodo, deleteTodo, getTodos } from './api/todos';
 import { ErrorType } from './types/ErrorType';
 import { StatusFilter } from './types/StatusFilter';
 import { TodoList } from './components/TodoList';
@@ -27,13 +26,11 @@ const getFilterTodos = (
 
     case StatusFilter.Active: {
       filteredTodos = todos.filter(todo => todo.completed === false);
-
       break;
     }
 
     case StatusFilter.Completed: {
       filteredTodos = todos.filter(todo => todo.completed === true);
-
       break;
     }
 
@@ -51,7 +48,6 @@ export const App: React.FC = () => {
   const [newTodoTitle, setNewTodoTitle] = useState('');
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [isRequesting, setIsRequesting] = useState(false);
-  // const [activeTodoId, setActiveTodoId] = useState(0);
   const [processingTodoIds, setProcessingTodoIds] = useState<number[]>([0]);
 
   const inputRef = useRef<HTMLInputElement | null>(null);
@@ -95,7 +91,7 @@ export const App: React.FC = () => {
     .filter(todo => todo.completed === false)
     .length;
   const countCompletedTodos = todos
-    .filter(todo => todo.completed === true)
+    .filter(todo => todo.completed)
     .length;
 
   const handleStatusChange = (filteredKey: StatusFilter) => {
@@ -191,14 +187,14 @@ export const App: React.FC = () => {
 
   const handleTitleUpdate = (todo: Todo, updatedTitle: string) => {
     setProcessingTodoIds(prevState => [...prevState, todo.id]);
-    todoService.updateTodo({
+
+    return todoService.updateTodo({
       id: todo.id,
       title: updatedTitle,
       userId: USER_ID,
       completed: todo.completed,
     })
       .then(updatedTodo => {
-        // setActiveTodoId(0);
         setTodos(prevState => {
           return prevState.map(currentTodo => {
             return currentTodo.id !== updatedTodo.id
@@ -207,9 +203,9 @@ export const App: React.FC = () => {
           });
         });
       })
-      .catch(() => {
+      .catch((error) => {
         showError(ErrorType.Update);
-        // setActiveTodoId(0);
+        throw error;
       })
       .finally(() => {
         return setProcessingTodoIds(prevState => prevState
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,24 +1,317 @@
...
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {/* this buttons is active only if there are some active todos */}
```
roman-mirzoian, line 239: Do not leave unnecessary comments.

src/App.tsx
```diff
@@ -1,24 +1,317 @@
...
+            <button
+              type="button"
+              className={classNames('todoapp__toggle-all',
+                { active: !countActiveTodos })}
+              data-cy="ToggleAllButton"
+              onClick={handleToggleTodo}
+            />
+          )}
+
+          {/* Add a todo on form submit */}
```
roman-mirzoian, line 250: Do not leave unnecessary comments.

src/App.tsx
```diff
@@ -1,24 +1,317 @@
...
+            {!!todos.length && (
+              <TodoFilter
+                status={status}
+                handleStatusChange={handleStatusChange}
+                countActiveTodos={countActiveTodos}
+                countCompletedTodos={countCompletedTodos}
+                handleClearCompleted={handleClearCompleted}
+              />
+            )}
+
```
roman-mirzoian, line 292:
```suggestion
```

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,68 @@
...
+  status,
+  handleStatusChange,
+  countActiveTodos,
+  countCompletedTodos,
+  handleClearCompleted,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${countActiveTodos} items left`}
```
roman-mirzoian, line 22: If there is one item, there should not be a plural in the text.

roman-mirzoian (changes requested):
Now, after pressing Enter, it does not update there (but it should according to the task):
https://gyazo.com/4b857711b0dde3b29be9273e13ae73fd

src/components/TodoFilter.tsx
Yevheniia-Sid: > If there is one item, there should not be a plural in the text.

I can't change it because the tests failed

src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
+      });
+  }, []);
+
+  useEffect(() => {
+    const timeoutId = setTimeout(() => {
+      setTodoError(null);
+    }, 3000);
+
+    return () => clearTimeout(timeoutId);
+  }, []);
```
BudnikOleksii, line 49:
```suggestion
  useEffect(() => {
    if (!todoError) {
      return;
    }
  
    const timeoutId = setTimeout(() => {
      setTodoError(null);
    }, 3000);

    return () => clearTimeout(timeoutId);
  }, [todoError]);
```
Add dependency to create timeout every time you have an error.
Also, create an `ErrorNotification` component and move this logic to it too.

src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
+  const handleStatusChange = (filteredKey: StatusFilter) => {
+    setStatus(filteredKey);
+  };
+
+  const showError = (error: ErrorType) => {
+    setTodoError(error);
+
+    setTimeout(() => {
+      setTodoError(null);
+    }, 3000);
```
BudnikOleksii, line 78: Set this time in `useEffect` as I mentioned before

src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (!newTodoTitle.trim()) {
+      showError(ErrorType.Title);
+    } else {
```
BudnikOleksii, line 85:
```suggestion
      showError(ErrorType.Title);
      return;
    }
    // Other code
```
Try to avoid `else`

src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
```
BudnikOleksii, line 210: Create a `Header` component

src/App.tsx
```diff
@@ -1,24 +1,289 @@
...
+                countCompletedTodos={countCompletedTodos}
+                handleClearCompleted={handleClearCompleted}
+              />
+            )}
+          </>
+        )}
+      </div>
+
+      <div
+        data-cy="ErrorNotification"
```
BudnikOleksii, line 268: Create an `ErrorNotification` component

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+    setNewTitle(todo.title);
+  };
+
+  const handleFormSubmit = (event:
+  React.FocusEvent<HTMLInputElement>
+  | React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (newTitle === todo.title) {
+      return;
+    }
```
BudnikOleksii, line 42:
```suggestion
    if (newTitle === todo.title) {
      setIsEditing(false);
      return;
    }
```
Maybe you should add `setIsEditing(false);` here, I can't close editing if the title is the same

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    if (newTitle.trim() === '') {
```
BudnikOleksii, line 44:
```suggestion
    if (!newTitle.trim()) {
```
Don't compare it with an empty string, just check if exists

BudnikOleksii (changes requested):
Good work 👍 
Left some suggestions, check them

PR #823, opened 2023-09-27
task solution was added

Liliya-Kalinichenko:
[DEMO LINK](https://Liliya-Kalinichenko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,83 @@
+import React, { useContext } from 'react';
+import cn from 'classnames';
+
+import { Status } from '../../types/FilterStatus';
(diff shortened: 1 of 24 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,83 @@
...
+          href="#/completed"
+          className={cn('filter__link', {
+            selected: filterParam === Status.Completed,
+          })}
+          onClick={() => onFilterChange(Status.Completed)}
+          data-cy="FilterLinkCompleted"
+        >
+          Completed
+        </a>
+      </nav>
```
anastasiia-tilikina, line 68: You can use map here. Use enum values and key for rendering

PR #824, opened 2023-09-27
initial commit: create todoList, todoItem, logic for item and input

maksym-mishchanchuk:
[DEMO LINK](https://maksym-mishchanchuk.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,320 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, {
+  useEffect, useState, useMemo, useRef,
+} from 'react';
+import { Todo } from './types/Todo';
+import { TodosList } from './components/TodosList/TodosList';
+import { Header } from './components/Header/Header';
+import { Footer } from './components/Footer/Footer';
+import {
+  ErrorNotification,
+} from './components/ErrorNotification/ErrorNotification';
+import { Loader } from './components/Loader';
 
-const USER_ID = 0;
+import * as todoService from './api/todos';
+import { ERROR_MESSAGES } from './utils/constants/ERROR_MESSAGES';
+import { getFilteredTodos } from './utils/helpers/getFilteredTodos';
+import { SortType } from './types/SortType';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loadingTodos, setLoadingTodos] = useState(true);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [selectFilter, setSelectFilter] = useState(SortType.All);
+  const textInputRef = useRef<HTMLInputElement | null>(null);
+  const [isProsessingTodoIds, setIsProsessingTodoIds] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    todoService.getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage(ERROR_MESSAGES.unableToLoadTodos);
+        // throw new Error(ERROR_MESSAGES.unableToLoadTodos);
+      })
+      .finally(() => {
+        setLoadingTodos(false);
+
+        textInputRef.current?.removeAttribute('disabled');
+        if (textInputRef.current) {
+          textInputRef.current.focus();
+        }
+      });
+  }, []);
+
+  useEffect(() => {
+    let timerError: NodeJS.Timeout | undefined;
+
+    if (errorMessage) {
+      timerError = setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    }
+
+    return () => {
+      clearTimeout(timerError);
+    };
+  }, [errorMessage]);
+
+  useEffect(() => {
+    textInputRef.current?.removeAttribute('disabled');
+    if (textInputRef.current) {
+      textInputRef.current.focus();
+    }
+  }, [todos]);
+
+  const handleAddTodo = (todoTitle: string) => {
+    setErrorMessage('');
+    setTempTodo({
+      id: 0,
+      title: todoTitle,
+      userId: 0,
+      completed: false,
+    });
+
+    return todoService.addTodo(todoTitle)
+      .then(todo => {
+        setTodos((prevState) => {
+          return [...prevState, todo];
+        });
+      })
+      .catch(() => {
+        setErrorMessage(ERROR_MESSAGES.unableToAddTodo);
+        throw new Error(ERROR_MESSAGES.unableToAddTodo);
+      })
+      .finally(() => {
+        setTempTodo(null);
+
+        textInputRef.current?.removeAttribute('disabled');
+        if (textInputRef.current) {
+          textInputRef.current.focus();
+        }
+      });
+  };
+
+  const handleDeleteTodo = (todoId: number) => {
+    setErrorMessage('');
+    setIsProsessingTodoIds((prevState) => {
+      return [...prevState, todoId];
+    });
+
+    return todoService.deleteTodos(todoId)
+      .then(() => {
+        // eslint-disable-next-line no-console
+        console.log('then');
+        setTodos(prevState => {
+          return prevState.filter(({ id }) => id !== todoId);
+        });
+      })
+      .catch(() => {
+        // eslint-disable-next-line no-console
+        console.log('catch');
+        setErrorMessage(ERROR_MESSAGES.unableToDeleteTodo);
+        throw new Error(ERROR_MESSAGES.unableToDeleteTodo);
+      })
+      .finally(() => {
+        // eslint-disable-next-line no-console
+        console.log('finally');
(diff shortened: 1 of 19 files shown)


OlhaMomot (approved):
Well done! App works smoothly👌

annak1104 (approved):
Great job!

src/App.tsx
```diff
@@ -1,24 +1,288 @@
...
+  const textInputRef = useRef<HTMLInputElement | null>(null);
+  const [isProsessingTodoIds, setIsProsessingTodoIds] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    todoService.getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage(ERROR_MESSAGES.unableToLoadTodos);
+        // throw new Error(ERROR_MESSAGES.unableToLoadTodos);
```
BudnikOleksii, line 33: Remove comments

src/App.tsx
```diff
@@ -1,24 +1,288 @@
...
+    if (errorMessage) {
+      timerError = setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    }
+
+    return () => {
+      clearTimeout(timerError);
+    };
+  }, [errorMessage]);
```
BudnikOleksii, line 57: You can move this logic to `ErrorNotification` component

src/App.tsx
```diff
@@ -1,24 +1,288 @@
...
+
+    return todoService.addTodo(todoTitle)
+      .then(todo => {
+        setTodos((prevState) => {
+          return [...prevState, todo];
+        });
+      })
+      .catch(() => {
+        setErrorMessage(ERROR_MESSAGES.unableToAddTodo);
+        throw new Error(ERROR_MESSAGES.unableToAddTodo);
```
BudnikOleksii, line 83: You don't need to throw errors if you don't handle them after

src/App.tsx
```diff
@@ -1,24 +1,288 @@
...
+      })
+      .finally(() => {
+        setTempTodo(null);
+
+        textInputRef.current?.removeAttribute('disabled');
+        if (textInputRef.current) {
+          textInputRef.current.focus();
+        }
```
BudnikOleksii, line 91: You do the same in `useEffect` above

src/App.tsx
```diff
@@ -1,24 +1,288 @@
...
+      })
+      .finally(() => {
+        setIsProsessingTodoIds((prevState) => {
+          return prevState.filter(id => id !== todoId);
+        });
+
+        textInputRef.current?.removeAttribute('disabled');
+        if (textInputRef.current) {
+          textInputRef.current.focus();
+        }
```
BudnikOleksii, line 119: The same

src/App.tsx
```diff
@@ -1,24 +1,288 @@
...
+        setIsProsessingTodoIds(prevState => {
+          return prevState.filter(id => id !== todo.id);
+        });
+      });
+  };
+
+  const handleToogleAllButton = async (allTodos: Todo[]) => {
+    const isAllTodosTrueOrFalse
+      = allTodos.every(todo => todo.completed)
+    || allTodos.every(todo => !todo.completed);
```
BudnikOleksii, line 205: It looks strange. You can just check if all completed.
In this case, toggle status for all todos, in another case `filter` them and toggle only uncompleted.

Also, set the loader for todos that are changing so the user case see them.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+        <a
+          onClick={() => changeSelectFilter(SortType.Completed)}
+          href="#/completed"
+          className={cn('filter__link',
+            { selected: selectFilter === SortType.Completed })}
+          data-cy="FilterLinkCompleted"
+        >
+          Completed
+        </a>
```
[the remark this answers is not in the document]
BudnikOleksii: > You can use SortType to render these elements, all logic is the same in each

Good point

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,91 @@
...
+        // eslint-disable-next-line jsx-a11y/control-has-associated-label
+        <button
+          onClick={() => handleToogleAllButton(todos)}
+          type="button"
+          className={cn('todoapp__toggle-all', { active: isAllTodosCompleted })}
+          data-cy="ToggleAllButton"
+        />
+      )}
+
+      {/* Add a todo on form submit */}
```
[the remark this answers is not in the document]
BudnikOleksii: > Probably these comments are not needed any more

Agree, remove all comments in your code

BudnikOleksii (changes requested):
Good job 👍 
Left some comments, check them. And add loaders for todos that are updating

src/App.tsx
```diff
@@ -1,24 +1,285 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
roman-mirzoian, line 1: Try to rewrite the code so that it is not necessary to disable the linter

src/App.tsx
roman-mirzoian: @maksym-mishchanchuk If you call some actions several times, you should put them in separate functions and reuse them

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+    if (!todoTitle) {
+      await onTodoDelete(todo.id);
+    } else {
+      await onTodoUpdate(todo, todoTitle);
+    }
+
+    setIsEditing(false);
+  };
+
+  const titleInput = useRef<HTMLInputElement | null>(null);
```
roman-mirzoian, line 59:
```suggestion
  const titleInputRef = useRef<HTMLInputElement | null>(null);
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+      titleInput.current.focus();
+    }
+  }, [isEditing]);
+
+  return (
+    <div
+      key={todo.id}
+      data-cy="Todo"
+      className={cn('todo',
+        { completed: todo.completed, 'item-enter-done': true })}
```
roman-mirzoian, line 72: Why are you adding a second class with the "true" condition? It will always be executed, so this class can simply be added to the list of classes without a condition

src/types/SortType.ts
```diff
@@ -0,0 +1,5 @@
+export enum SortType {
```
roman-mirzoian, line 1: These types will be used for `filtering`, why the name `Sort`?

roman-mirzoian (changes requested):
You can also now update the todo title to empty:
https://gyazo.com/4e89b442cd2cf2eb43245244965b16a3

src/App.tsx
```diff
@@ -1,24 +1,284 @@
...
+      .catch(() => {
+        setErrorMessage(ERROR_MESSAGES.unableToLoadTodos);
+      })
+      .finally(() => {
+        setLoadingTodos(false);
+
+        textInputRef.current?.removeAttribute('disabled');
+        if (textInputRef.current) {
+          textInputRef.current.focus();
+        }
```
roman-mirzoian, line 39: https://github.com/mate-academy/react_todo-app-with-api/pull/824#discussion_r1355043809 - I see that you have created a function for this, but you don't use it everywhere

src/App.tsx
```diff
@@ -1,24 +1,284 @@
...
+          return prevState.filter(id => id !== todo.id);
+        });
+      });
+  };
+
+  const handleToogleAllButton = async (allTodos: Todo[]) => {
+    const isAllTodosTrueOrFalse
+      = allTodos.every(todo => todo.completed);
+
+    const unCompletedTodos = allTodos.filter(todo => !todo.completed);
```
roman-mirzoian, line 203:
```suggestion
    const uncompletedTodos = allTodos.filter(todo => !todo.completed);
```

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,29 @@
...
+};
+
+export const ErrorNotification: React.FC<Props> = ({
+  errorMessage,
+  setErrorMessage,
+}) => {
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={cn('notification is-danger is-light has-text-weight-normal',
```
roman-mirzoian, line 16:
```suggestion
      className={cn('notification', 'is-danger', 'is-light', 'has-text-weight-normal',
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,66 @@
...
+            {sortType}
+          </a>
+        ))}
+      </nav>
+
+      <button
+        onClick={handleClearCopmpletedTodo}
+        disabled={!clearCompleteStatusButton}
+        type="button"
+        className={cn('todoapp__clear-completed')}
```
roman-mirzoian, line 59: Why is `classnames` used here, if we don't have any conditions, but just need to register a static class?

roman-mirzoian (changes requested):
You can still update the todo title to empty:
https://gyazo.com/172d00593e08b5c04705095f22159381

roman-mirzoian (approved):
Well done 👍

PR #829, opened 2023-09-27
solution

Dolnys:
[DEMO LINK](https://Dolnys.github.io/react_todo-app-with-api/) 

I'm sending my solution, some tests are not passing, although I don't understand why because in most cases the elements behave as they should.. I think I need to explore some other approach and maybe find better solutions .


pawel-peksa (changes requested):
Please use prettier to format the code. It will drastically increase readability.

src/App.tsx
```diff
@@ -1,24 +1,258 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+/* eslint-disable */
```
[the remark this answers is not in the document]
Dolnys: I knew I would have to make corrections, so I left the formatting until the end. I use Prettier, but it doesn't format my code in line with ESLint's expectations, and I didn't want to unnecessarily fix those errors with each correction.

cypress/integration/page.spec.js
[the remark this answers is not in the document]
Dolnys: Thee conflict arises because one of the tests was modified in the meantime.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+import React, { useEffect, useRef, useState } from "react";
+import { Todo } from "../types/Todo";
+import { updateTodo } from "../api/todos";
+
+type TodoItemProps = {
+  todo: Todo;
+  handleDeleteTodo: (todo: Todo) => void;
+  handleToggleComplete: (todo: Todo) => void;
+  isLoading: boolean;
+  handleErrorMessage: (message: string | null) => void;
```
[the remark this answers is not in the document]
Dolnys: I have created a separate enum containing all possible errors

src/App.tsx
```diff
@@ -1,24 +1,258 @@
...
+          try {
+            await deleteTodo(todo.id);
+            setTodos((prevTodos) => prevTodos.filter((t) => t.id !== todo.id));
+          } catch (err) {
+            handleErrorMessage("Unable to delete a todo");
+          }
+        })
+      );
+    } catch (err) {
+      handleErrorMessage("Unable to delete a todo");
```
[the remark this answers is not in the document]
Dolnys: Ok. Replaced with one  block

src/App.tsx
```diff
@@ -1,24 +1,258 @@
...
+    } finally {
+      setIsLoading(false);
+      setCurrentTodoLoading(null);
+      if (newTodoInputRef.current) {
+        setTimeout(() => {
+          newTodoInputRef.current?.focus();
+        }, 0);
```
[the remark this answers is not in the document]
Dolnys: I moved it to the  separate function  without timeout

Dolnys:
I've applied the suggested corrections and fixed the linting errors, there is few more errors now but the application still works as expected.. Next time, I'll definitely use context and I will better organize the structure of the app😬

src/App.tsx
```diff
@@ -1,24 +1,254 @@
...
-const USER_ID = 0;
+const USER_ID = 11546;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [filterType, setFilterType] = useState<'All' | 'Active' | 'Completed'>(
```
pawel-peksa, line 18: Rember about the DRY rule.
Let's export this to type defintion and use it in all 5 places,

src/App.tsx
```diff
@@ -1,24 +1,254 @@
...
 
-const USER_ID = 0;
+const USER_ID = 11546;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
```
pawel-peksa, line 17: Let's use ErrorType here

src/components/ErrorPopup.tsx
```diff
@@ -0,0 +1,28 @@
+import classNames from 'classnames';
+
+type ErrorPopupProps = {
+  error: string | null;
+  setError: (error: string | null) => void;
```
pawel-peksa, line 5:
```suggestion
  setError: (error: ErrorType | null) => void;
```

src/App.tsx
```diff
@@ -1,24 +1,254 @@
...
+  const [filterType, setFilterType] = useState<'All' | 'Active' | 'Completed'>(
+    'All',
+  );
+  const [newTitle, setNewTitle] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [TodoItem, setTodoItem] = useState<Todo | null>(null);
```
pawel-peksa, line 23:
```suggestion
  const [todoItem, setTodoItem] = useState<Todo | null>(null);
```

src/App.tsx
```diff
@@ -1,24 +1,254 @@
...
+  const handleAddTodo = async (title: string) => {
+    const trimmedTitle = title.trim();
+
+    const newTodo: Omit<Todo, 'id'> = {
+      title: trimmedTitle,
+      completed: false,
+      userId: USER_ID,
+    };
+
+    setTodoItem({ ...newTodo, id: 0 });
```
pawel-peksa, line 59: Are you sure that you want to setTodoItem before the check for empty title?

src/App.tsx
```diff
@@ -1,24 +1,254 @@
...
+    } catch (err) {
+      handleErrorMessage(ErrorType.UnableToUpdateTodo);
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    setError(null);
+    if (USER_ID) {
```
pawel-peksa, line 182: This is always true, I think

src/components/Footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+type FooterProps = {
+  todos: Todo[];
+  filterType: 'All' | 'Active' | 'Completed';
+  changeFilterStatus: (type: 'All' | 'Active' | 'Completed') => void;
+  handleClearCompleted: () => void;
+};
+
+export const Footer: React.FC<FooterProps> = ({
+  todos,
```
pawel-peksa, line 12: Nitpick: This component does not need to receive the whole list as a prop. I always prefer to pass as little information as needed, so in this case I would calculate `remainingItems` and `completedItems` in parent element and pass down only the number. 
It is not required however, so you don't need to change it if you don't want to. 
I just wrote this for your consideretion.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+    if (!editedTitle.trim()) {
+      handleDeleteTodo(todo);
+    } else if (editedTitle !== todo.title) {
+      try {
+        setIsEditingLoading(true);
+        const trimmedTitle = editedTitle.trim();
+
+        await updateTodo(todo.id, { title: trimmedTitle });
+        // eslint-disable-next-line no-param-reassign
+        todo.title = trimmedTitle;
```
pawel-peksa, line 44: Props are immutable, let's refactor this.

src/components/TodoList.tsx
```diff
@@ -0,0 +1,62 @@
...
+          handleErrorMessage={handleErrorMessage}
+        />
+      ))}
+      {todoItem && (
+        <TodoItem
+          key={todoItem.id}
+          todo={todoItem}
+          handleDeleteTodo={handleDeleteTodo}
+          handleToggleComplete={handleToggleComplete}
+          isLoading
```
pawel-peksa, line 56: Is it needed here?

pawel-peksa (changes requested):
Looking better, good job! I've added couple more comments, mainly regarding code duplication.
Almost there 💪

src/App.tsx
Dolnys: Removed the check

src/components/TodoList.tsx
Dolnys: The "temptodo" data should be enough; I've made a new component for temporary todos. I didn't need to use my "todoitem," which required passing all those props unnecessarily.

Dolnys:
I have implemented the suggested corrections.

choeqq (approved):
LGTM ✅

PR #831, opened 2023-09-27
add task solution

AngerDESTROYS:
[DEMO LINK](https://AngerDESTROYS.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -1,0 +1,24 @@
+import { TodoType } from '../types/TodoType';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<TodoType[]>(`/todos?userId=${userId}`);
+};
+
+export const deleteTodo = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const addTodo = ({ userId, title, completed }: Omit<TodoType, 'id'>) => {
+  return client.post<TodoType>('/todos', { userId, title, completed });
+};
+
+export const toggleTodoCompleteState = (todoId: number, completed: boolean) => {
+  return client.patch<TodoType>(`/todos/${todoId}`, { completed });
+};
+
+export const updateTodoTitle = (todoId: number, title: string) => {
+  return client.patch<TodoType>(`/todos/${todoId}`, { title });
+};
+
+// Add more methods here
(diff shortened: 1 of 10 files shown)


src/components/Filter.tsx
```diff
@@ -0,0 +1,39 @@
...
+    ))}
+    {/* // <a href="#/active" className="filter__link" data-cy="FilterLinkActive">
+    //   Active
+    // </a>
+    // <a */}
+    {/* //   href="#/completed"
+    //   className="filter__link"
+    //   data-cy="FilterLinkCompleted"
+    // >
+    //   Completed
```
etojeDenys, line 34: remove all comments

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,390 @@
...
+  return (
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {todos.length > 0 && (
+            <button
+              type="button"
+              className={`todoapp__toggle-all ${toggleAll ? 'active' : ' '}`}
```
etojeDenys, line 282: let's use classNames package for such cases

etojeDenys (changes requested):
The clear completed button should delete todos with animation

src/api/todos.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const addTodo = ({ userId, title, completed }: Omit<TodoType, 'id'>) => {
+  return client.post<TodoType>('/todos', { userId, title, completed });
+};
+
+export const toggleTodoCompleteState = (todoId: number, completed: boolean) => {
+  return client.patch<TodoType>(`/todos/${todoId}`, { completed });
+};
+
+export const updateTodoTitle = (todoId: number, title: string) => {
```
denys-danyliuk, line 20: you don't need to create separate methods for this, use one

denys-danyliuk (changes requested):

src/api/todos.ts
AngerDESTROYS: @denys-danyliuk but it takes different props i dont know how to make it with one

denys-danyliuk: if you have questions you should as them in fe_chat instead of requesting review again, don't pass separate properties, just pass data object where this properties may be

denys-danyliuk (changes requested):
if you have questions you should as them in fe_chat instead of requesting review again, don't pass separate properties, just pass data object where this properties may be

denys-danyliuk (approved):

PR #839, opened 2023-09-29
solution

Vira-V:
[DEMO LINK](https://Vira-V.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoFilter.tsx
+++ b/src/components/TodoFilter.tsx
@@ -1,0 +1,44 @@
+import cn from 'classnames';
+import { Filter } from '../types/Filters';
+
+type TodoFilterProps = {
+  filter: Filter;
+  setFilter: (newFilter: Filter) => void;
+};
+
+export const TodoFilter: React.FC<TodoFilterProps> = (
+  { filter, setFilter },
+) => {
+  return (
+    <nav className="filter" data-cy="Filter">
+      <a
+        href="#/"
+        className={cn('filter__link', { selected: filter === 'all' })}
+        data-cy="FilterLinkAll"
+        onClick={() => setFilter('all')}
+      >
+        All
+      </a>
+
+      <a
+        href="#/active"
+        className={cn('filter__link',
+          { selected: filter === 'active' })}
+        data-cy="FilterLinkActive"
+        onClick={() => setFilter('active')}
+      >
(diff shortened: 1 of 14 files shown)


src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,44 @@
...
+type TodoFilterProps = {
+  filter: Filter;
+  setFilter: (newFilter: Filter) => void;
+};
+
+export const TodoFilter: React.FC<TodoFilterProps> = (
+  { filter, setFilter },
+) => {
+  return (
+    <nav className="filter" data-cy="Filter">
```
choeqq, line 13:
```suggestion
) => (
    <nav className="filter" data-cy="Filter">
```

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+    <button
+      type="button"
+      className="todoapp__clear-completed"
+      data-cy="ClearCompletedButton"
+      onClick={() => todos.forEach(todo => {
+        if (todo.completed) {
+          handleDelete(todo);
+        }
+      })}
```
choeqq, line 33: this should be a separate function

src/components/TodoTemporary.tsx
```diff
@@ -0,0 +1,46 @@
+import cn from 'classnames';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  temporaryTodo: Todo | null;
+};
+
+export const TodoTemporary = ({ temporaryTodo }: Props) => {
+  if (temporaryTodo !== null) {
```
choeqq, line 9:
```suggestion
  if (temporaryTodo) {
```

choeqq, line 9: I think this if statement should handle !temporaryTodo then render something when it's not present

choeqq (changes requested):
After chaning completed's todo name it becomes active, please make sure todo remain in the same state as before editing

Please consider separating logic into hooks / global state, try to have components as lean as possible

choeqq (changes requested):

szymonpachucki (approved):
LGTM

choeqq (approved):
LGTM ✅

PR #846, opened 2023-10-01
 first solution with issues

vensecue:
[DEMO LINK](https://vensecue.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,208 @@
 /* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import {
+  USER_ID, addTodo, deleteTodo, getTodos, updateTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header/Header';
+import { List } from './components/List/List';
+import { Footer } from './components/Footer/Footer';
+import { handleError } from './handleError';
+import { ErrorMessage } from './types/ErrorMessage';
+import { Filter } from './types/Filter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [title, setTitle] = useState('');
+  const [filter, setFilter] = useState<Filter>('All');
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isDeleting, setIsDeleting] = useState<boolean>(false);
+  const [isToggling, setIsToggling] = useState<boolean>(false);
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,24 +1,208 @@
...
+        return false;
+      }
+
+      return true;
+    });
+
+    return visible;
+  };
+
+  const handleSubmit = (event: { preventDefault: () => void }) => {
```
choeqq, line 46: try to use Event type: https://bobbyhadz.com/blog/typescript-react-onsubmit-event-type

src/App.tsx
choeqq, line 1: I think you should consider splitting your component's logic into multiple custom hooks - App component grew really big, since you're passing a lot of functions and states to children components, maybe consider global state with context? 🤔 

that't not mandatory, but I think it would be a nice practice

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,61 @@
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+import { Filter } from '../../types/Filter';
+
+type FooterProps = {
+  counter: number;
+  filter: Filter;
+  setFilter: (filter: Filter) => void;
```
choeqq, line 8: you can copy type of setState function, when you hover over that function in VSC, it will be more specific than void

choeqq (changes requested):
If I mark Todo as completed only this one todo should start loading - in your case every single todo from the list starts loading - please try to fix it

choeqq (changes requested):
Waiting for any updates @vensecue 👀

choeqq (changes requested):
Waiting for any updates @vensecue 👀

choeqq (changes requested):
Waiting for any updates @vensecue 👀

src/components/Item/TodoItem.tsx
```diff
@@ -0,0 +1,111 @@
...
+        onTodoUpdate();
+      } catch {
+        handleError(setErrorMessage, ErrorMessage.noUpdateTodo);
+      }
+    }
+  };
+
+  const handleBlur = () => {
+    setIsEditing(false);
+  };
```
pawel-peksa, line 49: On blur we should:
- Do nothing only when title has not changed.
- Update title when it has changed
- Remove Todo when title left empty
This logic is missing here

src/components/Item/TodoItem.tsx
```diff
@@ -0,0 +1,111 @@
...
+      )}
+      {errorMessage && <div className="error-message">{errorMessage}</div>}
+      <button
+        type="button"
+        className="todo__remove"
+        data-cy="TodoDelete"
+        onClick={() => handleDelete(todo.id)}
+      >
+        ×
+      </button>
```
pawel-peksa, line 101: Should be hidden when isEditing is true

pawel-peksa (changes requested):
Good job, It looks like the App is almost working as expected even though a number of tests still fails.
I've added two comments regarding the improvement of app functionality.

pawel-peksa (approved):
Application works as expected. Good job! 🥳

PR #847, opened 2023-10-02
Solution

Nightycake:
[DEMO LINK](https://Nightycake.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,155 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useMemo, useState } from 'react';
+import { Todo } from './types/Todo';
+import {
+  addTodo, deleteTodo, getTodos, updateTodos,
+} from './api/todos';
+import { TodoFilter } from './components/TodoFilter';
+import { TodoList } from './components/TodoList';
+import { Header } from './components/Header';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Errors } from './types/Errors';
+import { Filters } from './types/Filters';
 
-const USER_ID = 0;
+const USER_ID = 11558;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMesssage, setErrorMesssageState] = useState<string>('');
+  const [filter, setFilter] = useState<Filters>(Filters.All);
+  const [title, setTitle] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isDisable, setIsDisable] = useState<boolean>(false);
+  const [loadingItems, setLoadingItems] = useState<number[]>([]);
+
+  const setErrorMesssage = (message: string) => {
+    if (message) {
+      setErrorMesssageState(message);
+      setTimeout(() => setErrorMesssageState(''), 3000);
+    } else {
+      setErrorMesssageState('');
+    }
+  };
+
+  const handleTodoClick = (todoId: number) => {
+    const updatedTodos = todos.map((todo) => {
+      if (todo.id === todoId) {
+        return {
+          ...todo,
+          completed: !todo.completed,
+        };
+      }
+
+      return todo;
+    });
+
+    setTodos(updatedTodos);
+    const clickedTodo = todos.find((todo) => todo.id === todoId);
+
+    if (clickedTodo) {
+      updateTodos(clickedTodo);
+    }
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID).then(setTodos)
+      .catch(() => setErrorMesssage(Errors.load));
+  }, []);
+
+  const filteredTodos = useMemo(() => [...todos].filter(({ completed }) => {
+    if (filter === Filters.All) {
+      return true;
+    }
+
+    return (filter === Filters.Active)
+      ? !completed
+      : completed;
+  }), [todos, filter]);
+
+  const removeTodo = (todoId: number) => {
+    setLoadingItems(current => [...current, todoId]);
+
+    deleteTodo(todoId)
+      .then(() => setTodos(current => current
+        .filter(todo => todo.id !== todoId)))
+      .catch(() => setErrorMesssage(Errors.delete))
+      .finally(() => setLoadingItems(current => current
+        .filter((id:number) => id !== todoId)));
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!title.trim()) {
+      return setErrorMesssage(Errors.title);
+    }
+
+    const newTodo = {
+      id: 0,
+      userId: USER_ID,
+      title: title.trim(),
+      completed: false,
+    };
+
+    setTempTodo(newTodo);
+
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,24 +1,155 @@
...
+
+    if (clickedTodo) {
+      updateTodos(clickedTodo);
+    }
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID).then(setTodos)
+      .catch(() => setErrorMesssage(Errors.load));
+  }, []);
```
choeqq, line 56: how about loading state for fetching all todos?

you can turn it off on `.finally`

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,50 @@
...
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${todos.filter(({ completed }) => !completed).length} items left`}
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+
+        {Object.values(Filters).map((filterType:Filters) => (
```
choeqq, line 25: is your prettier working correctly?

choeqq (changes requested):
Few suggestions:
1. Make sure toggle all works
2. make sure that your todo editor input styles are the same as main input styles

choeqq (changes requested):
Toggle looks great, but currently editing todo doesn't work - it just disappears instead of being applied to todo name

src/App.tsx
pawel-peksa: +1

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,25 @@
+export const ErrorNotification: React.FC<{ errorMesssage: string,
+  setErrorMesssage: (val: string) => void }> = ({
+  errorMesssage,
+  setErrorMesssage,
+}) => {
+  if (errorMesssage) {
+    setTimeout(() => setErrorMesssage(''), 3000);
```
pawel-peksa, line 7: Code duplication. Let's try to clear errorMessage only in one setTimeout

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,141 @@
...
+        >
+          <input
+            data-cy="TodoEditInput"
+            type="text"
+            className="todo__title-field"
+            value={editedTitle}
+            onChange={(e) => setEditedTitle(e.target.value)}
+            onBlur={handleTitleBlur}
+            // eslint-disable-next-line jsx-a11y/no-autofocus
+            autoFocus
```
pawel-peksa, line 109: You can use ref only to focus it on startup

pawel-peksa (changes requested):
Good job so far! Couple of more things need fixing.

choeqq (changes requested):
Waiting for any updates @Nightycake 👀

Nightycake:
@choeqq I tried to apply everything listed, but in TodoItem I had small problem with loading after editing a todo so I used local state and that worked. Maybe I missing something I dont see?

src/types/Filters.ts
```diff
@@ -0,0 +1,5 @@
+export enum Filters {
```
mbruhler, line 1: it's better to use `const enum`. There is less code overhead.

See the difference here https://phuoc.ng/collection/this-vs-that/const-enum-vs-enum/

src/App.tsx
```diff
@@ -1,24 +1,183 @@
...
+
+  const handleToggleAll = () => {
+    const areAllTodosCompleted = todos.every((todo) => todo.completed);
+    const updatedTodos = todos.map((todo) => ({
+      ...todo,
+      completed: !areAllTodosCompleted,
+    }));
+
+    setTodos(updatedTodos);
+    updatedTodos.forEach((updatedTodo) => {
```
mbruhler, line 138: Here also you don't need to pass full Todo to PATCH

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const TodoFilter: React.FC<Props> = ({
+  todos,
+  clearCompletedTodos,
+  selectedFilter,
+  onSelectedFilter,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${todos.filter(({ completed }) => !completed).length} items left`}
```
mbruhler, line 20: Counter is updated even if there is error

mbruhler (changes requested):
Hi!
There are still some issues with your code.
My note on `PATCH` relates to all update requests. Please review your code and change your code where applicable

Also please review tests, especially your app doesn't handle errors properly (f.e. updates counter when error occurs, when you check the todo)

There is more in this case, if you'll have problems with test please attend to Q&A and review them with mentor.

src/types/Filters.ts
Nightycake: Can't use const enum in this case.

choeqq (changes requested):
Waiting for any updates @Nightycake 👀

src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+  };
+
+  const clearCompletedTodos = () => {
+    // eslint-disable-next-line max-len
+    const completedTodoIds = todos.filter(({ completed }) => completed).map(({ id }) => id);
+
+    completedTodoIds.forEach((id) => removeTodo(id));
+  };
+
+  const handleToggleAll = () => {
```
pawel-peksa, line 130: One of the tests is failing because we don't set loading state to true here. 
We should show loader immediately after pressing a toggle button, before we actually get the response from the server.

src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+    }));
+
+    setTodos((prevTodos) => {
+      updatedTodos.forEach((updatedTodo) => {
+        if (updatedTodo.completed !== prevTodos.find((t) => t.id
+        === updatedTodo.id)?.completed) {
+          updateTodos(updatedTodo)
+            .catch(() => setErrorMesssage(Errors.update));
+        }
+      });
```
pawel-peksa, line 144: I'm not sure if this is the best solution here. This might be the reasone why tests:
```
 3) should become not active after toggling a todo
 4) should become not active
 5) should not be active
 6) should not become active after toggling a todo
 ```
 are failing.

File: src/components/TodoItem.tsx [hunk 10]
```diff
@@ -0,0 +1,160 @@
...
+    if (isEditing) {
+      window.addEventListener('keydown', handleKeyPress);
+    } else {
+      window.removeEventListener('keydown', handleKeyPress);
+    }
+
+    return () => {
+      window.removeEventListener('keydown', handleKeyPress);
+    };
+  }, [isEditing, editedTitle, todo]);
```
pawel-peksa, line 86: In general this is not the best wait to listen for a keyDown events.
Please use standard JSX/React approach with a `onKeyDown` event handler which is put inside JSX element.

pawel-peksa (changes requested):
Almost there 💪 There is still a number of tests failing. Try to fix some of them. If you need more support please reach out to mentors via Slack channel or joing Q&A session.

choeqq (changes requested):
Please make sure to adjust your code according to the comments reviewers left, and you will be good to go ✅

aleksandra-gajda:
...

pawel-peksa (approved):
All the comments have been resolved. Application works as expected! 🥳🥳🥳
Good job!!!

PR #848, opened 2023-10-02
Develop

ivannaromanovych:
[DEMO LINK](https://ivannaromanovych.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,204 +1,279 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useMemo, useState } from 'react';
+import cn from 'classnames';
 
-const USER_ID = 0;
+import { SelectFilter } from './utils/SelectFilter';
+import { Todo } from './types/Todo';
+import { ErrorMessage } from './utils/ErrorMessages';
+import * as ApiServices from './api/todos';
+import { TodoHeader } from './components/TodoHeader';
+import { TodoRow } from './components/TodoRow';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterOption, setFilterOption] = useState(SelectFilter.All);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [processingTodoIds, setProcessingTodoIds] = useState<number[]>([]);
+  const isEveryCompleted = useMemo(() => todos.every(todo => todo.completed),
+    [todos]);
+  const isAnyCompleted = useMemo(() => todos.some(todo => todo.completed),
+    [todos]);
+  const activeTodosCounter = useMemo(() => {
+    const activeTodos = todos.filter(todo => !todo.completed).length;
+
+    return activeTodos;
+  }, [todos]);
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
+      switch (filterOption) {
+        case SelectFilter.Active: return !todo.completed;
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,24 +1,280 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
roman-mirzoian, line 2: Try to rewrite the code so that it is not necessary to disable the linter

src/App.tsx
```diff
@@ -1,24 +1,280 @@
...
+    }
+
+    setTempTodo({
+      id: 0,
+      title: newTodoTitle,
+      userId: 0,
+      completed: false,
+    });
+
+    // eslint-disable-next-line consistent-return
```
roman-mirzoian, line 71: Try to rewrite the code so that it is not necessary to disable the linter

src/App.tsx
```diff
@@ -1,24 +1,280 @@
...
+      })
+      .finally(() => {
+        setProcessingTodoIds(prevState => prevState.filter(id => id !== todo.id));
+      });
+  };
+
+  const handleClearCompletedTodos = () => {
+    todos
+      .filter(todo => todo.completed)
+      .forEach(todo => handleTodoDelete(todo.id));
```
roman-mirzoian, line 156: No need to do extra filtering, this condition can be transferred to the middle of the `forEach` method

src/App.tsx
```diff
@@ -1,24 +1,280 @@
...
+  };
+
+  const handleToggleAll = () => {
+    if (isEveryCompleted) {
+      todos.forEach(handleTodoToggle);
+    } else {
+      todos
+        .filter(todo => !todo.completed)
+        .forEach(handleTodoToggle);
```
roman-mirzoian, line 165: The same about extra filtering.

src/App.tsx
```diff
@@ -1,24 +1,280 @@
...
+    todos
+      .filter(todo => todo.completed)
+      .forEach(todo => handleTodoDelete(todo.id));
+  };
+
+  const handleToggleAll = () => {
+    if (isEveryCompleted) {
+      todos.forEach(handleTodoToggle);
+    } else {
+      todos
```
roman-mirzoian, line 163:
```suggestion
    if (isEveryCompleted) {
      return todos.forEach(handleTodoToggle);
    }
    todos
```

src/App.tsx
```diff
@@ -1,24 +1,280 @@
...
+          ))}
+          {tempTodo && (
+            <TodoRow
+              todo={tempTodo}
+              isProcessing
+            />
+          )}
+        </section>
+
+        {/* Hide the footer if there are no todos */}
```
roman-mirzoian, line 200: Do not leave unnecessary comments.

src/App.tsx
```diff
@@ -1,24 +1,280 @@
...
+        {todos.length > 0 && (
```
roman-mirzoian, line 201:
```suggestion
        {!!todos.length && (
```

src/components/TodoRow.tsx
```diff
@@ -0,0 +1,126 @@
...
+            type="button"
+            className="todo__remove"
+            data-cy="TodoDelete"
+            onClick={onTodoDelete}
+          >
+            ×
+          </button>
+        </>
+      )}
+      {/* overlay will cover the todo while it is being updated */}
```
roman-mirzoian, line 112: Do not leave unnecessary comments.

roman-mirzoian (changes requested):
Good work, but some comments on the code and the application itself:
1. If we try to create a todo with an empty name, we will see an error, but after that the input becomes inactive and we will not be able to add anything new, only after reloading the page:
https://gyazo.com/bc0b7e5aaf5f416faa31f3a517b631d1
2. If we exit the edit mode using the Escape key, the new name will not be saved, but the next time we enter the edit mode, we will see the next state from the previous entry into the edit mode:
https://gyazo.com/1f59971ebddccd3694d2814a556551d0

PR #861, opened 2023-10-10
Solution

Reject-4444:
https://reject-4444.github.io/react_todo-app-with-api/

i know, that not all tests is passed, but if to look on layout, everything is working

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,208 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import { UserWarning } from './UserWarning';
+import { Filter, Todo } from './types/Todo';
+import * as todosService from './api/todos';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Notification } from './components/Notification';
+import { TempTodo } from './components/TempTodo';
 
-const USER_ID = 0;
+const USER_ID = 11641;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [todoTitle, setTodoTitle] = useState('');
+  const [filterBy, setFilterBy] = useState<Filter>(Filter.all);
+  const [hasCompleted, setHasCompleted] = useState(false);
+  const [quantity, setQuantity] = useState(0);
+  const [hideNotification, setHideNotification] = useState(true);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isFocused, setIsFocused] = useState(true);
+
+  useEffect(() => {
+    const filteredTodos = todos.filter((todo) => todo.completed);
+
+    setQuantity(todos.length - filteredTodos.length);
+  }, [todos]);
+
+  useEffect(() => {
+    setErrorMessage('');
+    todosService
+      .getTodos(USER_ID)
+      .then((data) => {
+        setTodos(data);
+      })
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
+      });
+  }, []);
+
+  const filteredTodos = useMemo(() => {
+    const filteredByTodos = todos.length
+      ? todos.filter((todo) => {
+        switch (filterBy) {
+          case Filter.active:
+            return !todo.completed;
+          case Filter.completed:
+            return todo.completed;
+          default:
+            return true;
+        }
+      })
+      : [];
+
+    return filteredByTodos;
+  }, [filterBy, todos]);
+
+  const deleteTodo = (todoId: number) => {
+    setErrorMessage('');
+    setIsFocused(false);
+
+    return todosService
+      .deleteTodo(todoId)
+      .then(() => {
+        setTodos((currentTodos) => currentTodos
+          .filter((todo) => todo.id !== todoId));
+      })
+      .catch((error) => {
+        setErrorMessage('Unable to delete a todo');
+        throw error;
+      })
+      .finally(() => {
+        setIsFocused(true);
+      });
+  };
+
+  const createNewTodo = ({ title, completed, userId }: Omit<Todo, 'id'>) => {
+    setErrorMessage('');
+
+    return todosService
+      .addTodo({ title, completed, userId })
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,24 +1,208 @@
...
+    setQuantity(todos.length - filteredTodos.length);
+  }, [todos]);
+
+  useEffect(() => {
+    setErrorMessage('');
+    todosService
+      .getTodos(USER_ID)
+      .then((data) => {
+        setTodos(data);
+      })
```
denys-danyliuk, line 41:
```suggestion
      .then(setTodos})
```
this can be shortened

src/App.tsx
```diff
@@ -1,24 +1,208 @@
...
+
+  useEffect(() => {
+    const filteredTodos = todos.filter((todo) => todo.completed);
+
+    setQuantity(todos.length - filteredTodos.length);
+  }, [todos]);
+
+  useEffect(() => {
+    setErrorMessage('');
+    todosService
```
denys-danyliuk, line 37:
```suggestion
    setErrorMessage('');
    
    todosService
```

src/App.tsx
```diff
@@ -1,24 +1,208 @@
...
+        switch (filterBy) {
+          case Filter.active:
+            return !todo.completed;
+          case Filter.completed:
+            return todo.completed;
+          default:
+            return true;
+        }
+      })
+      : [];
```
denys-danyliuk, line 59: simplify this logic

denys-danyliuk (changes requested):
why toggle all button is active if not all todos are completed? also it doesn't work at all
Знімок екрана 2023-10-10 о 15 31 36

edition of the todos is broken, when I write something, then press escape button and again doubleclick on a todo old text is saved

https://github.com/mate-academy/react_todo-app-with-api/assets/106310796/642809fe-28e0-46e6-ac2c-e858bd19cc39

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+          .filter((todo) => todo.id !== todoId));
+      })
+      .catch((error) => {
+        setErrorMessage('Unable to delete a todo');
+        throw error;
+      })
+      .finally(() => {
+        setIsFocused(true);
+      });
+  };
```
Andriy-Shymkiv, line 84: consider to use async/await instead of then/catch. it is more readable imho
```suggestion
  const deleteTodo = async (todoId: number) => {
  setErrorMessage('');
  setIsFocused(false);

  try {
    await todosService.deleteTodo(todoId);
    const updatedTodos = currentTodos => currentTodos.filter(todo => todo.id !== todoId);
    setTodos(updatedTodos);
  } catch (error) {
    setErrorMessage('Unable to delete a todo');
  } finally {
    setIsFocused(true);
  }
};
```

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+    const hasCompletedTodos = filteredTodos.some((todo) => todo.completed);
+
+    setHasCompleted(hasCompletedTodos);
+  };
+
+  useEffect(() => {
+    const hasCompletedTodos = filteredTodos.some((todo) => todo.completed);
+
+    setHasCompleted(hasCompletedTodos);
+  }, [filteredTodos, hasCompleted]);
```
Andriy-Shymkiv, line 133: `hasCompleted` it is the same as `quantity === 0`, isn't it?

Andriy-Shymkiv: if so, then you don't need this effect and this state at all, but I'm not sure, please check

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+
+  useEffect(() => {
+    setHideNotification(!errorMessage.length);
+  }, [errorMessage]);
+
+  useEffect(() => {
+    const hasAllCompleted = filteredTodos.every(todo => todo.completed);
+
+    setAllTodosCompleted(hasAllCompleted);
+  }, [filteredTodos, allTodosCompleted]);
```
Andriy-Shymkiv, line 143: why you need this effect? please consider to remove it because it looks unnecessary

src/components/Header.tsx
```diff
@@ -0,0 +1,106 @@
...
+                reset();
+              })
+              .catch(() => setErrorMessage('Unable to add a todo'))
+              .finally(() => {
+                setIsSubmitting(false);
+                setTempTodo(null);
+                setIsFocused(true);
+              });
+          }
+        }}
```
Andriy-Shymkiv, line 89: move this function outside of event for more readability. also consider to use async/await

src/components/Notification.tsx
```diff
@@ -0,0 +1,52 @@
...
+  error,
+  hideNotification,
+  setHideNotification,
+}) => {
+  useEffect(() => {
+    if (!hideNotification) {
+      setTimeout(() => {
+        setHideNotification(true);
+      }, 3000);
+    }
```
Andriy-Shymkiv, line 20: you can just `setError('')`, why so complicated?

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    if (makeTodosComplete === MakeTodosCompleted.not && checked) {
+      setIsClicked(true);
+      setChecked(false);
+      updateTodo({ ...todo, completed: false })
+        .then(() => setIsClicked(false))
+        .finally(() => setMakeTodosComplete(MakeTodosCompleted.begin));
+    }
+  }, [makeTodosComplete]);
```
Andriy-Shymkiv, line 114: update deps

Andriy-Shymkiv (changes requested):

src/App.tsx
Reject-4444: I need this effect because we set the result of these effect into state which i use in button that toggles all todos.  We need to know if all todos are completed and we click on button, all todos should be uncompleted. If they are all uncompleted and we click in button, all todos should be completed. And if some of buttons aren't completed and we click on button, all buttons should be completed. To do this logic, i need to know about this

src/components/TodoItem.tsx
Reject-4444: i don't need to update depts because if i do it, the code won't work. (we don't need to track another depts, this effect should work only if makeTodosComplete will change

src/App.tsx
Reject-4444: quantity shows us how many todos aren't completed. We use it to show this amount in the left bottom of app.
state hasCompleted gives us possibility to disable button clear all completed

PR #868, opened 2023-10-15
problems with tests

w315:
[DEMO LINK](https://w315.github.io/react_todo-app-with-api/)

Everything seems to be working fine, but there are several similar errors in the tests that I don't understand how to solve


volodymyr-soltys97 (changes requested):
Good job 👍 
When I edit the name after the change I get this error, this needs to be fixed
image

PR #885, opened 2023-10-25
Todo with API

JulyaPetrovskaya:
[DEMO LINK](https://JulyaPetrovskaya.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,7 +7,6 @@ import { TodoList } from './components/TodoList/TodoList';
 import { TodoFooter } from './components/TodoFooter/TodoFooter';
 import { Header } from './components/Header/Header';
 
-
 export const App: React.FC = () => {
   const {
     todos,
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,140 +1,20 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useState, useEffect, useMemo } from 'react';
+import React from 'react';
 import classNames from 'classnames';
+import { TodosContext, USER_ID } from './TodosContext';
 import { UserWarning } from './UserWarning';
-import { Todo } from './types/Todo';
-import { StatusFilter } from './types/Filter';
-import * as todosServices from './api/todos';
 import { TodoList } from './components/TodoList/TodoList';
 import { TodoFooter } from './components/TodoFooter/TodoFooter';
 import { Header } from './components/Header/Header';
 
-const USER_ID = 11587;
 
 export const App: React.FC = () => {
-  const [todos, setTodos] = useState<Todo[]>([]);
-  const [isloadingTodo, setIsLoadingTodo] = useState<number[]>([]);
-  //  Этот массив отслеживает состояние загрузки для каждой
-  //  задачи (TODO) в виде идентификаторов (id). По сути,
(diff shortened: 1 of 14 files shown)


xiimiik (changes requested):
this button should work for both sides
image

on toggle any of those buttons, i cant add new TODO, because input is disabled
(on edit the same)
image

add some favicon
image

JulyaPetrovskaya:
моя favicon відображується у мене локально, але в demo - чомусь - ні!? не знаю як виправити інші помилки в тестах 
image

src/App.tsx
```diff
@@ -1,24 +1,64 @@
...
+            {tempTodo && (
+              <TodoItem
+                todo={tempTodo}
+              />
+            )}
+
+            {!!todos.length && (
+              <TodoFooter />
+            )}
+          </>
```
etojeDenys, line 43: you have double length check here

src/TodosContext.tsx
```diff
@@ -0,0 +1,229 @@
...
+    const { completed } = todo;
+
+    const newTodo = {
+      ...todo,
+      completed: !completed,
+    };
+
+    updateTodo(newTodo)
+      .then(() => {})
+      .catch(() => {});
```
etojeDenys, line 186: looks like then and catch are redundant here

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { useEffect, useRef, useCallback } from 'react';
+import classNames from 'classnames';
+import { TodosContext } from '../../TodosContext';
+
+export const Header: React.FC = () => {
+  const {
+    todos,
+    title,
```
etojeDenys, line 8: do you need to save this field in context?  I suggest to you it will be better for the performance and readability  if you create the title state here, and pass this title as an argument to addTodo

etojeDenys (changes requested):

xiimiik (changes requested):
not fixed
Снимок экрана 2023-10-26 в 14 14 41

src/TodosContext.tsx
```diff
@@ -0,0 +1,224 @@
...
+type Props = {
+  children: React.ReactNode;
+};
+
+export const TodosProvider: React.FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoadingTodo, setIsLoadingTodo] = useState<number[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [statusFilter, setStatusFilter] = useState(Filter.ALL);
+  const [statusResponse, setStatusResponse] = useState(false);
```
loralevitska, line 42: Rename all boolean variables according [tips](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig)

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
```
loralevitska, line 4: Remove comment. Try it write obvious code that explain all we need without comments)
Fix all cases

loralevitska (approved):
Well done) Check my comments for improvement

image

PR #887, opened 2023-10-26
react_todo-app-with-api

IhorVerba:
[DEMO LINK](https://IhorVerba.github.io/react_todo-app-with-api/)

 3 tests are pending


etojeDenys (changes requested):
let's fix it

 looks different
image

volodymyr-soltys97 (approved):
Nice work 🔥

PR #888, opened 2023-10-28
solution

drhtka:
https://drhtka.github.io/react_todo-app-with-api/


src/TodosContext.tsx
```diff
@@ -0,0 +1,63 @@
...
+  setIsProcessing: () => {},
+});
+
+type Props = {
+  children: React.ReactNode;
+};
+
+export const TodoProvider: React.FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isChecked, setIsChecked] = useState<boolean>(false);
```
SanyaBratashchuk, line 27: there is no reason to keep data in state if it can be calculated based on existing ones

SanyaBratashchuk (approved):

src/TodosContext.tsx
drhtka: I did it this way to keep the data in a state so that its changes were synchronized between all components.

PR #906, opened 2023-11-08
made Todo

sintax1s:
[DEMO LINK](https://sintax1s.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoItem.tsx
+++ b/src/components/TodoItem.tsx
@@ -1,0 +1,207 @@
+import cn from 'classnames';
+import {
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import { Todo } from '../types/Todo';
+import { TodoContext } from '../TodoContext';
+import { ErrorMessage } from '../types/ErrorMessage';
+import { removeTodo, updateTodo } from '../api/todos';
+import { Key } from '../types/Key';
+
+type Props = {
+  todo: Todo
+};
+
+export const TodoItem: React.FC<Props> = ({ todo }) => {
+  const {
+    todos,
+    tempTodo,
+    setTodos,
+    setError,
+    isClearCompleted,
+    setIsUpdating,
+    isUpdating,
+    toggledTodos,
+    setToggledTodos,
+    isToggleAllClicked,
+    isEveryTodoCompleted,
+    inputRef,
+  } = useContext(TodoContext);
+
+  const [
+    deletingTodo,
+    setDeletingTodo,
+  ] = useState<Todo | undefined>(undefined);
+
+  const [updatingTodo, setUpdatingTodo] = useState<number | null>(null);
+  const [updatedTitle, setUpdatedTitle] = useState(todo.title);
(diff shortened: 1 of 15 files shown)


src/components/TodoItem.tsx
```diff
@@ -0,0 +1,207 @@
...
+      setUpdatingTodo(null);
+      setIsUpdating(false);
+    }
+  };
+
+  return (
+    <div
+      data-cy="Todo"
+      className={cn('todo', { completed: todo.completed })}
+      onDoubleClick={() => handleDoubleClick(todo.id)}
```
mpohorenyi, line 141: In this and the following examples, you do not need to directly pass the `todo.id` as an argument to your handle function. You can use `todo.id` in your functions directly.

src/types/DefaultValue.ts
```diff
@@ -0,0 +1,25 @@
+import { ErrorMessage } from './ErrorMessage';
+import { Status } from './Status';
+import { Todo } from './Todo';
+
+export interface DefaultValue {
+  todos: Todo[];
+  setTodos: (value: Todo[] | ((value: Todo[]) => Todo[])) => void;
```
mpohorenyi, line 7: You can type your setter using the default type it has:
`React.Dispatch<React.SetStateAction<Todo[]>>`

mpohorenyi (approved):
Great job, well done! 
Pay attention to the comments.

src/components/TodoItem.tsx
sintax1s: Ty, good point

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+  const handleTodoToggle = useCallback(() => {
+    setToggledTodos(todo.id);
+    setIsToggled(true);
+
+    updateTodo(todo.id, { completed: !todo.completed })
+      .then(() => {
+        const updatedTodos = todos.map((t) => {
+          if (t.id === todo.id) {
+            return { ...t, completed: !t.completed };
```
vasylmiskiv, line 67: The same, try to write more appropriate name of the element

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,214 @@
...
+        className={cn(
+          'modal overlay',
+          {
+            'is-active': todo.id === tempTodo?.id
+            || todo.id === deletingTodo?.id
+            || (isClearCompleted && todo.completed)
+            || (isUpdating && updatingTodo === todo.id)
+            || (isToggled && toggledTodos === todo.id)
+            || (isEveryTodoCompleted && isToggleAllClicked && todo.completed)
+            || (!isEveryTodoCompleted && isToggleAllClicked && !todo.completed),
```
vasylmiskiv, line 205: Here is not a good idea to write the logic in the jsx. Would be better to do it separately

vasylmiskiv (approved):
Looks fine but pay attention to some comments

sintax1s:
> Looks fine but pay attention to some comments

Thank you, done

PR #911, opened 2023-11-09
add todo app solution

o-drozzdyk:
[DEMO LINK](https://o-drozzdyk.github.io/react_todo-app-with-api/)
Can't pass 1 test but app seems to work as expected

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,136 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { Header } from './components/Header';
+import { Todo } from './types/Todo';
+import {
+  USER_ID, deleteTodo, getTodos, updateTodo,
+} from './api/todos';
+import { Filter, Footer } from './components/Footer';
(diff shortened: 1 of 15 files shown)


src/api/todos.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const addTodo = ({ userId, title, completed }: Omit<Todo, 'id'>) => {
+  return client.post<Todo>('/todos', { userId, title, completed });
+};
+
+export const deleteTodo = (id: number) => {
+  return client.delete(`/todos/${id}`);
+};
+
+export const updateTodo = (id: number, todo: Todo) => {
```
YaroslavYarynych, line 18: I think you don't need to have id here. You can use todo.id for your url path.

src/App.tsx
```diff
@@ -1,24 +1,136 @@
...
+        .filter(todo => todo.id !== todoId)))
+      .catch(() => {
+        setErrorMessage('Unable to delete a todo');
+      })
+      .finally(() => {
+        setLoadingTodosIds(current => current.filter(id => id !== todoId));
+      });
+  }
+
+  function updateTodoData(id: number, updatedTodo: Todo) {
```
YaroslavYarynych, line 60: I think you can use updatedTodo.id when you need id.

YaroslavYarynych (approved):
Good job! Also you can use Context for making your states global and with it you won`t have a lot of props. But it's up to you.

PR #914, opened 2023-11-10
solution react_todo-app-with-api task

ivaZaiets:
[DEMO LINK](https://ivaZaiets.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,264 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+import { Todo } from './types/Todo';
+
+import { USER_ID } from './helpers/userId';
+import * as dataOperations from './api/todos';
+
+import { TodoList } from './components/TodoList/TodoList';
+import { TodoError } from './components/TodoError/TodoError';
+import { TodoFooter } from './components/TodoFooter/TodoFooter';
+import { Header } from './components/Header/Header';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const [query, setQuery] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isTodoSelected, setIsTodoSelected] = useState('');
+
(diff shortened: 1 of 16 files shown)


tetyana-pol (approved):
Good job, but
in general, I think you should have used the Context, then you would have had less props

src/App.tsx
```diff
@@ -1,24 +1,264 @@
...
+import { TodoFooter } from './components/TodoFooter/TodoFooter';
+import { Header } from './components/Header/Header';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const [query, setQuery] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isTodoSelected, setIsTodoSelected] = useState('');
```
polosanya, line 21: The name is a bit confusing. Sounds like a boolean

src/App.tsx
```diff
@@ -1,24 +1,264 @@
...
+
+  const [isTodoLoading, setIsTodoLoading] = useState<Todo | null>(null);
```
polosanya, line 23: Fim similar cases
https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

src/App.tsx
```diff
@@ -1,24 +1,264 @@
...
+        setErrorMessage('Unable to update a todo');
+        throw error;
+      })
+      .finally(() => setIsTodoLoading(null));
+  };
+
+  let copyOfTodos = [...todos];
+
+  switch (isTodoSelected) {
+    case 'all':
```
polosanya, line 200: Consider using Enum instead of the hardcoded values

src/App.tsx
```diff
@@ -1,24 +1,264 @@
...
+        });
+      })
+      .catch((error) => {
+        setErrorMessage('Unable to update a todo');
+        throw error;
+      })
+      .finally(() => setIsTodoLoading(null));
+  };
+
+  let copyOfTodos = [...todos];
```
polosanya, line 197: Filter is a non-mutating method, so there is actually no need to create a copy

src/components/TodoError/TodoError.tsx
```diff
@@ -0,0 +1,35 @@
...
+  errorMessage: string,
+};
+
+export const TodoError: React.FC<Props> = ({
+  onErrorMessage,
+  errorMessage,
+}) => {
+  return (
+    <>
+      <div
```
polosanya, line 16: React fragment is redundant here as you already have a container (div)

polosanya (approved):
Good job! Left a few comments that might be useful)

PR #921, opened 2023-11-12
add todo app api solution

Kuper18:
Hi, here is my solution. Functionality is working, but I have one failed test.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,212 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState } from 'react';
 
-const USER_ID = 0;
+import { Todo } from './types/Todo';
+import { deleteTodos, getTodos, updateTodos } from './api/todos';
+import { Header } from './Components/Header';
+import { TodoList } from './Components/TodoList';
+import { Footer } from './Components/Footer';
+import { ErrorNotification } from './Components/ErrorNotification';
+import { FilterParams } from './types/FilteredParams';
+import { UpdatingData } from './types/UpdatingData';
 
-export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+const USER_ID = 11853;
+
+function filterBy(todos: Todo[], filterValue: string) {
+  let filteredTodos = todos;
+
+  switch (filterValue) {
+    case FilterParams.ACTIVE:
+      filteredTodos = filteredTodos.filter(todo => !todo.completed);
+      break;
+
+    case FilterParams.COMPLETED:
+      filteredTodos = filteredTodos.filter(todo => todo.completed);
+      break;
+
+    case FilterParams.All:
+    default:
+      return filteredTodos;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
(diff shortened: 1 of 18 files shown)


src/App.tsx
```diff
@@ -1,24 +1,212 @@
...
+import { TodoList } from './Components/TodoList';
+import { Footer } from './Components/Footer';
+import { ErrorNotification } from './Components/ErrorNotification';
+import { FilterParams } from './types/FilteredParams';
+import { UpdatingData } from './types/UpdatingData';
 
-export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+const USER_ID = 11853;
```
ivaZaiets, line 12: Will be better use this variable (```USER_ID```) in file 'helpers'

src/Components/Footer/Footer.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const isSomeCompletedTodo = todos.some(todo => todo.completed);
+  const activeTodos = todos.filter(todo => !todo.completed);
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeTodos.length} items left`}
+      </span>
+
+      <nav className="filter" data-cy="Filter">
```
ivaZaiets, line 28: You can use this logic in another component where you can write only code with select. And then use this component in component ```Footer```

src/Components/Header/Header.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const isUnableAddTodo = useRef(true);
+
+  useEffect(() => {
+    textInput.current?.focus();
+  }, [isDisableInput]);
+
+  const onSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const newTodo = {
```
ivaZaiets, line 39: If you write your constant with userId in another file 'helpers' you can export this constant instead ```11853```. I think it's will be better

src/Components/Header/Header.tsx
```diff
@@ -0,0 +1,108 @@
...
+          type="button"
+          className={cn('todoapp__toggle-all', {
+            active: isCompletedTodos,
+          })}
+          data-cy="ToggleAllButton"
+          onClick={makeAllTodosCompleted}
+        />
+      )}
+
+      <form onSubmit={onSubmit}>
```
ivaZaiets, line 94: In my opinion, will be better write form in another component and write code which related to form in that component also

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,159 @@
...
+      {isActiveInput ? (
+        <form onSubmit={onSubmit}>
+          <input
+            data-cy="TodoTitleField"
+            type="text"
+            className="todo__title-field"
+            placeholder="Empty todo will be deleted"
+            value={newTodoTitle}
+            onChange={(event) => setNewTodoTitle(event.target.value)}
+            onBlur={onBblur}
```
ivaZaiets, line 126: I think you can use in ```onBlur``` onSubmit too, this function have the same logic

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,159 @@
...
+            // eslint-disable-next-line jsx-a11y/no-autofocus
+            autoFocus
```
ivaZaiets, line 128: Will be better use useRef hook, autofocus unreliable

ivaZaiets (changes requested):
Good job! Let's improve your code a bit more

StasSokolov (approved):
Looks like masterpiece🔥

PR #922, opened 2023-11-12
solution

MVoutch:
[DEMO LINK](https://MVoutch.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,97 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useRef, useState } from 'react';
+import { Todo } from './types/Todo';
+import { TodosContext } from './components/TodosContext';
+import { TodosFilter } from './types/TodosFilter';
+import { TodosHeader } from './components/TodosHeader';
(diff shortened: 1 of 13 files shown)


tilugdmytro (changes requested):
Great job no doubts!

Please, just add some changes)

src/App.tsx
```diff
@@ -1,24 +1,99 @@
...
+
+const useFilter = (todos: Todo[], filter: string) => {
+  return todos.filter(todo => {
+    switch (filter) {
+      case TodosFilter.active:
+        return !todo.completed;
+      case TodosFilter.completed:
+        return todo.completed;
+      default:
+        return true;
```
polosanya, line 28: You can move switch out of the filter callback to avoid calling it if the case is default

src/components/TodosItem.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+  const handleChange = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    event.preventDefault();
+
+    switch (event.key) {
+      case 'Escape':
+        resetChange();
+
+        break;
+      case 'Enter':
```
polosanya, line 95: Would be great to use Enum instead of hardcoded values

polosanya (approved):
Well done!

PR #925, opened 2023-11-13
pre alpha without tests

Clover247:
[DEMO LINK](https://clover247.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -1,0 +1,136 @@
+import {
+  useCallback,
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
+import cn from 'classnames';
+import { TodoContext, USER_ID } from '../TodoContext';
+import { addTodo, editTodo } from '../api/todos';
+import { Error } from '../types/Error';
+
+export const Header: React.FC = () => {
+  const {
+    todos,
+    inputRef,
+    setTempTodo,
+    setTodos,
+    setError,
+    tempTodo,
+    setToggledTodos,
+    isEveryTodoCompleted,
+    setIsToggleAllClicked,
+  } = useContext(TodoContext);
+
+  const [inputValue, setInputValue] = useState('');
+  const [isToggleAll, setIsToggleAll] = useState(false);
+
+  useEffect(() => {
+    setIsToggleAll(isEveryTodoCompleted);
+  }, [todos, isEveryTodoCompleted]);
+
+  const onToggleAll = useCallback(() => {
+    setIsToggleAllClicked(true);
+
+    if (isToggleAll) {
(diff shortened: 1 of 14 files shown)


ElenaGrasovskaya (approved):
Works like a charm! I've got the same crazy lists of imported stuff from the context :)

src/components/Header.tsx
```diff
@@ -0,0 +1,138 @@
...
+          )}
+          data-cy="ToggleAllButton"
+          aria-label="ToggleAll"
+          onClick={onToggleAll}
+        />
+      )}
+
+      <form onSubmit={handleSubmit}>
+        <input
+          disabled={tempTodo !== null}
```
StasSokolov, line 124:
```suggestion
          disabled={tempTodo}
```

can be simplify, should work same

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,70 @@
...
+          type="button"
+          className="delete"
+          aria-label="hide error"
+          onClick={closeNotification}
+        />
+        {error && (
+          <>
+            { error }
+          </>
+        )}
```
StasSokolov, line 65:
```suggestion
        {error || ''}
```

can be a bit simplify

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,99 @@
...
+          Active
+        </a>
+
+        <a
+          href="#/completed"
+          className={cn(
+            'filter__link', { selected: selectedType === TodoStatus.Completed },
+          )}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setSelectedType(TodoStatus.Completed)}
```
StasSokolov, line 81: looks like we can iterate these elements using `.map()`

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,213 @@
...
+    setToggledTodos,
+    isToggleAllClicked,
+    isEveryTodoCompleted,
+    inputRef,
+  } = useContext(TodoContext);
+
+  const [
+    deletingTodo,
+    setDeletingTodo,
+  ] = useState<Todo | undefined>(undefined);
```
StasSokolov, line 37: let's use `null` everywhere

StasSokolov (changes requested):
Almost done, let's improve your code a bit more🔥

src/components/Header.tsx
Clover247: it's not because we take 'null' like a basic parameter for useState in context

StasSokolov (approved):
Looks like masterpiece, i'm approving it🔥

PR #928, opened 2023-11-13
Solution ver1

hrysta:
[DEMO LINK](https://hrysta.github.io/react_todo-app-with-api/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3407,9 +3407,9 @@
       }
     },
     "@mate-academy/eslint-config-react-typescript": {
-      "version": "1.0.11",
+      "version": "1.0.12",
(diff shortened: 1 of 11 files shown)


polosanya (changes requested):
It would be great to hide the error if the problem is fixed (or after a few seconds)
image
Also, there is bug when switching between tabs. Please fix it
delete-bug

dsfreedom (approved):
Everything seems fine to me) good job!

StasSokolov (changes requested):
Almost done, let's fix only few minor comments🔥

StasSokolov (changes requested):
Hey, i don't see any changes after review((

StasSokolov (approved):
Looks good to me, i'm approving it🔥

PR #931, opened 2023-11-13
Completed the task

Maksim114488:
https://Maksim114488.github.io/react_todo-app-with-api/


Stim1 (approved):
Looks good to me ! 👍🏼 

A few recommendations that I think would improve your code :

- it would be nice to use interfaces instead of types and maybe it's better to put them in separate files.
- split app component into more components for better code structure
- it would be better to remove unnecessary comments

PR #934, opened 2023-11-14
Solution without animation.

Stim1:
[DEMO LINK](https://Stim1.github.io/react_todo-app-with-api/)
 
###  **I'll add animations and fix the loader issue later**

I don't understand why I have a problem with the tests, locally alll test passes.

image

### DIFF
--- a/src/TodoFilter.tsx
+++ b/src/TodoFilter.tsx
@@ -12,7 +12,7 @@ export const TodoFilter: React.FC<TodoFilterProps> = ({
       {Object.values(FilterBy).map(filterType => (
         <a
           key={filterType}
-          href={`#/${filterType === FilterBy.All && filterType}`}
+          href={`#/${filterType}`}
           onClick={handleFilterClick(filterType)}
           className={cn('filter__link', { selected: filterBy === filterType })}
           data-cy={`FilterLink${filterType}`}


src/App.tsx
```diff
@@ -1,24 +1,264 @@
...
+  const [editingId, setEditingId] = useState<number | null>(null);
+  const [editText, setEditText] = useState('');
+
+  useEffect(() => {
+    const fetchTodos = async () => {
+      try {
+        const todosFromServer = await TodoService.getTodos(USER_ID);
+
+        setTodos(todosFromServer);
+        setFilteredTodos(todosFromServer);
```
StasSokolov, line 34: i think `filteredTodos` is redundant, we can operate only one state with todos

src/TodoFilter.tsx
```diff
@@ -0,0 +1,41 @@
...
+      </a>
+      <a
+        href="#/completed"
+        onClick={handleFilterClick(FilterBy.Completed)}
+        className={cn('filter__link',
+          { selected: filterBy === FilterBy.Completed })}
+        data-cy="FilterLinkCompleted"
+      >
+        Completed
+      </a>
```
StasSokolov, line 38: not critical but looks like we can iterate theses elements using `.map`

StasSokolov (changes requested):
Overall looks good to me, let's improve your code a bit more🔥

Also i can't open demo link
Снимок экрана 2023-11-16 в 10 30 13

StasSokolov (approved):
Looks good to me, i'm approving it🔥

PR #937, opened 2023-11-14
add task solution

andrew17752985:
[demo](https://andrew17752985.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoItem/TodoItem.tsx
+++ b/src/components/TodoItem/TodoItem.tsx
@@ -1,0 +1,124 @@
+import React, { useEffect, useRef, useState } from 'react';
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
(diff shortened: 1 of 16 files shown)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,124 @@
...
+      {isEditing ? (
+        <form onSubmit={handleSubmit}>
+          <input
+            data-cy="TodoTitleField"
+            type="text"
+            ref={field}
+            className="todo__title-field"
+            placeholder="Empty todo will be deleted"
+            defaultValue={todo.title}
+            onChange={event => setNewTitle(event.target.value)}
```
volodymyr-soltys97, line 89: You can move this logic to the helper function and use it here

volodymyr-soltys97 (approved):
Nice work 🔥

PR #941, opened 2023-11-16
added the first solution to the task

annabasenyuk:
[DEMO LINK](https://annabasenyuk.github.io/react_todo-app-with-api/) 
I skipped several tests

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,205 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useState, useEffect } from 'react';
+import cn from 'classnames';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
+import { ErrorMessage } from './types/ErrorMessage';
+import { getTodos, updateTodo } from './api/todos';
+import { TodoHeader } from './components/TodoHeader';
+import { TodoItem } from './components/TodoItem';
+import { TodoFooter } from './components/TodoFooter';
 
-const USER_ID = 0;
+const USER_ID = 11859;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState<boolean | number>(false);
+  const [todosError, setTodosError] = useState(ErrorMessage.Default);
+  const [filter, setFilter] = useState(Filter.All);
+  const [tempTodos, setTempTodos] = useState<Todo | null>(null);
+  const [isHidden, setIsHidden] = useState(false);
+  const [processingTodoId, setProcessingTodoId] = useState<Array<number>>([]);
+  const [filteredTodos, setFilTodos] = useState<Todo[]>([]);
+  const [newTodoLoad, setNewTodoLoad] = useState(-1);
+  const [processDelete, setProcDelete] = useState(false);
+
+  const loadTodos = async () => {
+    setIsLoading(true);
+    try {
+      const todosData = await getTodos(USER_ID);
+
+      setTodos(todosData);
+      setFilTodos(todosData);
+    } catch (error) {
+      setTimeout(() => {
+        setTodosError(ErrorMessage.UnableToLoadTodos);
+      }, 3000);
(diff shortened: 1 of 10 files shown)


src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  useEffect(() => {
+    if (focusedInput.current) {
+      focusedInput.current.focus();
+    }
+  }, [tempTodos, todos.length]);
+
+  const isActiveButton = todos.every(todo => todo.completed);
+
+  const addTodoHandler = async (event: React.FormEvent) => {
```
Kuper18, line 41: would be better to call this function something like this `handlerSubmitData`

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,122 @@
...
+      }, 3000);
+
+      throw new Error('Some error');
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  const onTodoToggleHandler = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
```
Kuper18, line 84: why do you need it here? When you click the button, it doesn't cause any default browser behavior

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,183 @@
...
+      setIsEditing(false);
+      setNewTitle(todo.title);
+
+      return;
+    }
+
+    if (event.key === 'Enter') {
+      handleSubmit(event);
+    }
+  };
```
Kuper18, line 109: use useEffect like component did mount and will unmount to remove this event

Kuper18 (changes requested):
Good job! But your app doesn't work properly, after removing all todos I couldn't add new todo, you should figure it out how to fix the code!

src/App.tsx
```diff
@@ -1,24 +1,205 @@
...
+          return todo.completed;
+
+        case Filter.All:
+        default:
+          return true;
+      }
+    });
+
+    setFilTodos(newTodos);
+  }, [todos, filter]);
```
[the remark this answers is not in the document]
StasSokolov: agree with @Kuper18 , let's use only one state with todos

src/App.tsx
```diff
@@ -1,24 +1,208 @@
...
+              <TodoItem
+                todo={todo}
+                key={todo.id}
+                todos={filteredTodos}
+                setTodos={setTodos}
+                isLoading={isLoading}
+                setIsLoading={setIsLoading}
+                setTodosError={setTodosError}
+                onTodoUpdate={handleUpdate}
+                isNew={(todo.id === newTodoLoad) && true}
```
StasSokolov, line 150:
```suggestion
                isNew={todo.id === newTodoLoad}
```

should work same but simplify, check all places

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,111 @@
...
+          href="#/active"
+          className={cn(
+            'filter__link', { selected: filter === Filter.Active },
+          )}
+          data-cy="FilterLinkActive"
+          onClick={() => setFilter(Filter.Active)}
+        >
+          Active
+        </a>
+
```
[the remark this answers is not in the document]
StasSokolov: check this comment from your mate also please

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,128 @@
...
+        setTodosError(ErrorMessage.TitleShouldNotBeEmpty);
+      }, 3000);
+
+      return;
+    }
+
+    setIsLoading(true);
+    let id;
+
+    if (todos.length === 0) {
```
StasSokolov, line 54:
```suggestion
    if (!todos.length) {
```

can be a bit simplify

StasSokolov (changes requested):
Good job🔥

First of all check please demo link, i can't add todos
Снимок экрана 2023-11-20 в 15 54 45

annabasenyuk:
I found out one problem with id. When adding todos that contain the number 3 in the id, a 500 server error is thrown. This causes my ids to start with 4, but it doesn't completely solve the problem

StasSokolov (changes requested):
Almost done, some issue still exist

User see error when refresh page
Снимок экрана 2023-11-27 в 11 37 36

Can't add todos, check please
Снимок экрана 2023-11-27 в 11 38 38

PR #949, opened 2023-11-21
Develop1

Liubomyr19:
[DEMO LINK](https://Liubomyr19.github.io/react_todo-app-with-api/)


IvanFesenko (changes requested):
Please re-check your app
https://github.com/mate-academy/react_todo-app-with-api/assets/61904995/3bfc8537-cb31-4489-ba5d-85e229887136

src/App.tsx
```diff
@@ -1,24 +1,234 @@
...
+      ))
+      .catch(() => {
+        showError(ErrorMessage.deleteError);
+      })
+      .finally(() => setProcessingIds([]));
+  };
+
+  const handleUpdateTodo = (todo: Todo) => {
+    setProcessingIds(current => [...current, todo.id]);
+    const currentTodo = todos.filter(toDo => toDo.id === todo.id)[0];
```
TarasHoliuk, line 73: you use `filter` like `find`.
```suggestion
    const currentTodo = todos.find(toDo => toDo.id === todo.id);
```

TarasHoliuk: or even `findIndex`

src/App.tsx
```diff
@@ -1,24 +1,234 @@
...
+    const indexOfTodo = todos.indexOf(currentTodo);
+
+    updateTodo(todo)
+      .then(() => setTodos(
```
TarasHoliuk, line 77: you may use `map` and update todo in state by id

src/App.tsx
```diff
@@ -1,24 +1,234 @@
...
+    setTempTodo({ ...newTodo, id: 0 });
+
+    setProcessingIds(current => [...current, 0]);
+    createTodo(newTodo)
+      .then(newToDo => {
+        setTodos(currentTodos => [...currentTodos, newToDo]);
+        resetField();
+      })
+      .catch(() => {
+        setTempTodo(null);
```
TarasHoliuk, line 140: redundant, it's in finally

src/App.tsx
```diff
@@ -1,24 +1,234 @@
...
+    resetField();
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(data => {
+        const todosData = data as Todo[];
+
+        setTodos(todosData);
+      })
```
TarasHoliuk, line 157: based on this you don't need it:
image

but anyway you can do it like this:
```suggestion
      .then(data => {
        setTodos(data as Todo[]);
      })
```

src/components/TodosFillter/TodosFillter.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { Filter } from '../../types/Fillter';
+
+type Props = {
+  todos: Todo[];
+  filterBy: Filter;
+  setFilterBy: React.Dispatch<React.SetStateAction<Filter>>;
+  clearCompleted: () => void
+};
+
+export const TodosFilter: React.FC<Props> = (
```
TarasHoliuk, line 13: It's better to rename component to `Footer`, since there is not only filter

TarasHoliuk (changes requested):
Hey! Very good job

Left a few comments - they are really small, but some things look a bit strange, so let's improve them
P.S. try to fix linter error

src/api/todos.ts
```diff
@@ -0,0 +1,42 @@
...
+  return client.post<Todo>(
+    '/todos',
+    {
+      id,
+      title,
+      completed,
+      userId,
+    },
+  );
+};
```
DarkMistyRoom, line 22:
```suggestion
export const createTodo = (
  todo: Todo,
) => {
  return client.post<Todo>(
    '/todos', todo
  );
};
```

DarkMistyRoom: why don't do it this way

src/App.tsx
DarkMistyRoom: not fixed
image

DarkMistyRoom (approved):
Good job! Check the comments

PR #950, opened 2023-11-21
dev

nikkias:
[DEMO LINK](https://nikkias.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -1,0 +1,22 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+export const addTodo = ({ title, completed, userId }: Omit<Todo, 'id'>) => {
+  return client.post<Todo>('/todos', { title, completed, userId });
+};
+
+export const deleteTodo = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const updatedTodo = ({
+  id, title, completed, userId,
+}: Todo) => {
+  const trimmedTitle = title.trim();
+
+  return client.patch<Todo>(`/todos/${id}`, { trimmedTitle, completed, userId });
+};
--- a/src/components/Error.tsx
+++ b/src/components/Error.tsx
@@ -1,0 +1,40 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, { useContext } from 'react';
+import cn from 'classnames';
+import { TodosContext } from './TodosContext';
+
+export const Error: React.FC = () => {
+  const { errorMessage, setErrorMessage } = useContext(TodosContext);
+  let isErrorHidden = true;
+
+  if (errorMessage !== '') {
+    isErrorHidden = false;
+    setTimeout(() => {
+      setErrorMessage('');
+      isErrorHidden = true;
+    }, 3000);
+  }
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={cn(
+        'notification',
+        'is-danger',
+        'is-light',
+        { hidden: isErrorHidden },
+      )}
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={() => {
+          setErrorMessage('');
+          isErrorHidden = true;
+        }}
+      />
+      {errorMessage}
+    </div>
+  );
+};
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -1,0 +1,91 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import cn from 'classnames';
+import { TodosContext } from './TodosContext';
+import * as todoService from '../api/todos';
+
+export const Header = () => {
+  const [inputValue, setInputValue] = useState('');
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const {
+    todos,
+    setTodos,
+    setErrorMessage,
+    USER_ID,
+  } = useContext(TodosContext);
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, []);
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value);
+  };
+
+  const handlePressEnter = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter') {
+      if (inputValue.trim() === '') {
+        setErrorMessage('Title should not be empty');
+      } else {
+        todoService.addTodo({
+          title: inputValue,
+          completed: false,
+          userId: USER_ID,
+        }).then(newTodo => {
+          setTodos([...todos, newTodo]);
+        }).catch(() => setErrorMessage('Unable to add a todo'));
+
+        setInputValue('');
+      }
+    }
+  };
+
+  const handleChangeToggle = () => {
+    const completedValue = todos.some(todo => !todo.completed);
+
+    const updatedTodos = todos.map(todo => {
+      return ({
+        ...todo,
+        completed: completedValue,
+      });
+    });
+
+    setTodos(updatedTodos);
+  };
+
+  return (
+    <header className="todoapp__header">
+      <form onSubmit={(event) => event.preventDefault()}>
+        {todos.length > 0 && (
+          <button
+            type="button"
+            className={cn(
+              'todoapp__toggle-all',
+              { active: todos.every(todo => todo.completed) },
+            )}
+            data-cy="ToggleAllButton"
+            onClick={handleChangeToggle}
+          />
+        )}
+
+        <input
+          data-cy="NewTodoField"
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          value={inputValue}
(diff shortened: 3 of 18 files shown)


TarasHoliuk:
When there is no `Clear completed` button, filter buttons "jump" to the right side:

image

Display filter always in the center. Also, there is a redundant white line under the footer - try to get rid of it.

src/api/todos.ts
```diff
@@ -0,0 +1,22 @@
...
+};
+
+export const deleteTodo = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const updatedTodo = ({
+  id, title, completed, userId,
+}: Todo) => {
+  const trimmedTitle = title.trim();
```
TarasHoliuk, line 19: it's better to prepare all data before. This function shouldn't manipulate data, just send request (single responsibility principle)

src/components/Error.tsx
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, { useContext } from 'react';
+import cn from 'classnames';
+import { TodosContext } from './TodosContext';
+
+export const Error: React.FC = () => {
+  const { errorMessage, setErrorMessage } = useContext(TodosContext);
+  let isErrorHidden = true;
```
TarasHoliuk, line 8: generally `isErrorHidden === !errorMessage`, so you can get rid of `isErrorHidden` variable and always use `!errorMessage`, don't you?

src/components/TodosContext.tsx
```diff
@@ -0,0 +1,61 @@
...
+      .then(setTodos)
+      .catch(() => setErrorMessage('Unable to load todos'));
+  }, []);
+
+  const value = useMemo(() => ({
+    todos,
+    setTodos,
+    filterOption,
+    setFilterOption,
+    USER_ID,
```
TarasHoliuk, line 51:
```suggestion
    userId: USER_ID,
```
generally, you can just write it out of component and export/import wherever you need
But indeed it _could_ be a dynamic variable, so you can keep it in the context, but change the object key to `camelCase`

src/components/TodoList.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { TodoItem } from './TodoItem';
+import { TodosContext } from './TodosContext';
+import * as todoService from '../api/todos';
+import { Todo } from '../types/Todo';
+import { FilterOption } from '../types/FilterOption';
+
+export const TodoList: React.FC = () => {
+  const { todos, filterOption, setErrorMessage } = useContext(TodosContext);
+
+  const visibleTodo = todos.filter(todo => {
```
TarasHoliuk, line 11:
```suggestion
  const visibleTodos = todos.filter(todo => {
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,51 @@
...
+      default:
+        return true;
+    }
+  });
+
+  const deleteTodo = (id: number) => {
+    return todoService.deleteTodo(id)
+      .catch((error) => {
+        setErrorMessage('Unable to delete a todo');
+        throw error;
```
TarasHoliuk, line 27: do not throw the error again - it will lead to the page error, you already caught it and handled, no need to throw it again

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,178 @@
...
+  useState,
+} from 'react';
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+import { TodosContext } from './TodosContext';
+
+type Props = {
+  todo: Todo,
+  deleteTodo: (id: number) => Promise<unknown>,
+  updatedTodo: (newTodo: Todo) => Promise<unknown>,
```
TarasHoliuk, line 15:
```suggestion
  updateTodo: (newTodo: Todo) => Promise<Todo>,
```
it returns todo (you may check it in the network tab in devTools also)
update**d**Todo sounds like it's an object of todo that was updated, function should contain verb in the name

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,178 @@
...
+};
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  deleteTodo,
+  updatedTodo,
+}) => {
+  const [editValue, setEditValue] = useState(todo.title);
+  const [editTodosId, setEditTodosId] = useState(0);
```
TarasHoliuk, line 24:
```suggestion
  const [editingTodoId, setEditTodosId] = useState(0);
```
and fix `set` accordingly
it's 1 todo which is in `editing` state at the moment

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  const updateTitle = (newTitle: string) => {
+    const trimmedTitle = newTitle.trim();
+
+    if (!trimmedTitle) {
+      deleteTodo(todo.id)
+        .then(() => {
+          const newTodos = todos.filter((t) => t.id !== todo.id);
+
+          setTodos(newTodos);
```
TarasHoliuk, line 43:
```suggestion
          setTodos(currentTodos => currentTodos.filter((t) => t.id !== todo.id));
```
it's better to do it in this way if it's possible: https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,178 @@
...
+    if (event.key === 'Enter') {
+      updateTitle(editValue.trim());
+    } else if (event.key === 'Escape') {
+      setEditValue(todo.title.trim());
+      setEditTodosId(0);
+    }
+  };
+
+  const handleOnBlueEditValue = () => {
+    updateTitle(editValue.trim());
```
TarasHoliuk, line 98: you shouldn't send the request if nothing was changed

src/components/Header.tsx
```diff
@@ -0,0 +1,91 @@
...
+        completed: completedValue,
+      });
+    });
+
+    setTodos(updatedTodos);
+  };
+
+  return (
+    <header className="todoapp__header">
+      <form onSubmit={(event) => event.preventDefault()}>
```
TarasHoliuk, line 65:
```suggestion
      <form onSubmit={handlePressEnter}>
```
and you may omit `onKeyPress={handlePressEnter}` in input

TarasHoliuk:
Try to add a loader during actions like in [the working example](https://mate-academy.github.io/react_todo-app-with-api/): 
[Screencast from 21-11-23 20:38:44.webm](https://github.com/mate-academy/react_todo-app-with-api/assets/86923089/d5d470d7-661f-43d1-979e-2e5556533674)

TarasHoliuk (changes requested):
Well done :+1: 
Let's add some fixes

src/components/Error.tsx
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, { useContext, useEffect } from 'react';
+import cn from 'classnames';
+import { TodosContext } from './TodosContext';
+
+export const Error: React.FC = () => {
+  const { errorMessage, setErrorMessage } = useContext(TodosContext);
+
+  useEffect(() => {
+    if (errorMessage !== '') {
```
TarasHoliuk, line 10:
```suggestion
    if (errorMessage) {
```
Just FYI: string will be automatically transformed into boolean

src/components/TodoList.tsx
```diff
@@ -0,0 +1,50 @@
...
+      });
+  };
+
+  const updatedTodo = (newTodo: Todo) => {
+    return todoService.updatedTodo(newTodo)
+      .catch((error) => {
+        setErrorMessage('Unable to update a todo');
+        throw error;
+      });
+  };
```
TarasHoliuk, line 36: Hey! I see you've fixed such issues exactly in the places I marked

But when you see some comments you should fix all similar cases:
image
image

src/components/TodosContext.tsx
TarasHoliuk, line 51: not fixed

src/api/todos.ts
TarasHoliuk: not fixed
image

src/components/TodoItem.tsx
TarasHoliuk: not fixed:
image

src/components/Header.tsx
TarasHoliuk: not critical, but better to do it in the described way. Ignored:

TarasHoliuk (changes requested):
You've fixed only a few comments, and silently ignored most of them. If you have questions/the comments are not clear to you/need help - feel free to ask in the chat. Do not ignore comments silently

src/api/todos.ts
```diff
@@ -0,0 +1,41 @@
...
+const prepareTodoData = (todo: Todo) => {
+  const {
+    id,
+    title,
+    completed,
+    userId,
+  } = todo;
+  const trimmedTitle = title.trim();
+
+  if (!id || !trimmedTitle || userId === undefined) {
```
loralevitska, line 13:
```suggestion
  if (!id || !trimmedTitle || !userId) {
```

src/components/Main.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { TodoList } from './TodoList';
+
+export const Main: React.FC = () => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
```
loralevitska, line 6: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots and `ul` paddings

loralevitska (approved):
Good job! Check my comments for improvement

- something went wrong with this styles, fix it according [working example](https://mate-academy.github.io/react_todo-app-with-api/)
image

PR #958, opened 2023-11-24
add task solution

MariaSnegireva:
[DEMO LINK](https://MariaSnegireva.github.io/react_todo-app-with-api/)


etojeDenys (changes requested):
you demo link doesn't work, try to fix that

sTorba24 (changes requested):
Good work! Lets fix bug you have

after adding 1 item i can't focus on input
image

sTorba24 (approved):

PR #959, opened 2023-11-24
task solution

Anmonnn:


volodymyr-soltys97 (changes requested):
When I click on the checkbox here then the case is deleted immediately, but it should be deleted only when clicking on the cross
image
image

src/App.tsx
```diff
@@ -1,24 +1,245 @@
...
+  const [filterType, setFilterType] = useState('all');
+  const [typeError, setTypeError] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    // Функция для загрузки данных с сервера
```
loralevitska, line 25: Remove comment. Try to write obvious code with meaningful variable and function naming. Fix all cases

src/App.tsx
```diff
@@ -1,24 +1,245 @@
...
+      id: Math.floor(Math.random() * 1000),
+      title: inputValue,
+      completed: false,
+      userId: 11910,
+    };
+
+    setTempTodo(obj);
+
+    try {
+      if (inputValue.trim().length === 0) {
```
loralevitska, line 72:
```suggestion
      if (!inputValue.trim()) {
```

loralevitska (changes requested):
Almost perfect) Let's improve your code

loralevitska (approved):

PR #972, opened 2023-12-05
solution

KseniyaGonchar:
https://KseniyaGonchar/.github.io/react_todo-app-with-api/

### DIFF
--- a/src/components/TodoCard.tsx
+++ b/src/components/TodoCard.tsx
@@ -1,0 +1,132 @@
+import React, { useEffect, useRef, useState } from 'react';
+import cn from 'classnames';
+import { Todo } from '../types/Todo';
+import { updateTodoItems } from '../api/todos';
+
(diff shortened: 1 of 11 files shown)


src/components/TodoCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+      title: todo.title,
+      userId: USER_ID,
+      completed: !todo.completed,
+    };
+    const newTodos = [...todos];
+    const index = newTodos.findIndex(currentTodo => (
+      currentTodo.id === newTodo.id));
+
+    newTodos.splice(index, 1, newTodo);
+
```
IvanFesenko, line 72: use Array.filter() instead

IvanFesenko (changes requested):
Good work. Check my one minor comment

IvanFesenko (approved):
Well done

PR #974, opened 2023-12-06
todo with api

valeriaoleynik:
[DEMO LINK](https://valeriaoleynik.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoItem/TodoItem.tsx
+++ b/src/components/TodoItem/TodoItem.tsx
@@ -63,10 +63,6 @@ export const TodoItem: React.FC<Props> = ({ todo }) => {
 
   const handleDoubleClick = () => {
     setIsEdit(true);
-
-    setTimeout(() => {
-      titleField.current?.focus();
-    }, 1);
   };
 
   return (


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+      setEditTitle(title);
+    }
+  };
+
+  const handleDoubleClick = () => {
+    setIsEdit(true);
+
+    setTimeout(() => {
+      titleField.current?.focus();
+    }, 1);
```
SanyaBratashchuk, line 69: why do you need this 1ms timeout?

SanyaBratashchuk (changes requested):
Implement tempTodo on add(check [working example](https://mate-academy.github.io/react_todo-app-with-api/))

src/components/TodoItem/TodoItem.tsx
valeriaoleynik: It's my mistake, timeout is not needed here


plus tempTodo was implemented in the function addTodo in the TodosContext component

SanyaBratashchuk (approved):

PR #975, opened 2023-12-07
Solution

VKdrvtsv:
[DEMO LINK](https://VKdrvtsv.github.io/react_todo-app-with-api/)


src/utils/GlobalContext.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { useState, useEffect } from 'react';
+import debounce from 'lodash.debounce';
```
VitaliyBondarenko1982, line 2: for fix this linter issue  - commit and push changes package.json file. even better commit and push all files with changes.
Screenshot from 2023-12-07 18-11-36

src/App.tsx
```diff
@@ -1,24 +1,91 @@
...
 
 export const App: React.FC = () => {
+  const {
+    error,
+    setError,
+    todoList,
+    setTodoList,
+    USER_ID,
+  } = useContext(PageContext);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>(todoList);
```
VitaliyBondarenko1982, line 22: why do you need state for filteredTodos?
in context you alreade has state for all todos.
it is should be enough. just filter todos be active filter where you need it.

VitaliyBondarenko1982 (changes requested):
Looks good.
Please check comments.

volodymyr-soltys97 (approved):
Great work 🔥

PR #981, opened 2023-12-11
Solution

IrynaK-a:
[DEMO LINK](https://IrynaK-a.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -5,6 +5,7 @@ import { FilterStatus } from '../types/FilterStatus';
 import { DispatchContext, StateContext } from './TodosProvider';
 import { deleteTodo } from '../api/todos';
 import { ErrorMessage } from '../types/ErrorMessage';
+import { LoadingStatus } from '../types/LoadingStatus';
 
 export const Footer = () => {
   const { filteredBy, todos } = useContext(StateContext);
@@ -20,7 +21,10 @@ export const Footer = () => {
   const unCompletedTodosCount = todos.length - completedTodosCount;
 
   const clearCompletedTodos = async () => {
-    dispatch({ type: 'shoulDeleteCompleted' });
+    dispatch({
+      type: 'shouldLoading',
+      payload: LoadingStatus.Completed,
+    });
 
     todos.forEach(async todo => {
       if (todo.completed) {
@@ -36,10 +40,14 @@ export const Footer = () => {
             type: 'error',
             payload: ErrorMessage.Deleting,
           });
+        } finally {
+          dispatch({
+            type: 'shouldLoading',
+            payload: LoadingStatus.None,
+          });
         }
       }
     });
-    dispatch({ type: 'shoulDeleteCompleted' });
   };
 
   const filterTodos = (item: FilterStatus) => {
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,96 @@
+import cn from 'classnames';
+import { useContext } from 'react';
+
+import { FilterStatus } from '../types/FilterStatus';
+import { DispatchContext, StateContext } from './TodosProvider';
+import { deleteTodo } from '../api/todos';
+import { ErrorMessage } from '../types/ErrorMessage';
+
+export const Footer = () => {
+  const { filteredBy, todos } = useContext(StateContext);
+  const dispatch = useContext(DispatchContext);
+
+  const filteredStatus = Object.values(FilterStatus);
+
+  const completedTodosCount = todos.filter(
+    todo => todo.completed,
+  )
+    .length || 0;
+
+  const unCompletedTodosCount = todos.length - completedTodosCount;
+
+  const clearCompletedTodos = async () => {
+    dispatch({ type: 'shoulDeleteCompleted' });
+
+    todos.forEach(async todo => {
+      if (todo.completed) {
+        try {
+          await deleteTodo(todo.id);
+
+          dispatch({
+            type: 'deleteTodo',
+            payload: todo.id,
+          });
+        } catch (error) {
+          dispatch({
+            type: 'error',
+            payload: ErrorMessage.Deleting,
+          });
+        }
+      }
+    });
+    dispatch({ type: 'shoulDeleteCompleted' });
+  };
+
+  const filterTodos = (item: FilterStatus) => {
+    dispatch({
+      type: 'filter',
+      payload: item,
+    });
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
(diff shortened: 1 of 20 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { ErrorMessage } from '../types/ErrorMessage';
+import { LoadingStatus } from '../types/LoadingStatus';
+
+export const Footer = () => {
+  const { filteredBy, todos, newTodoInputRef } = useContext(StateContext);
+  const dispatch = useContext(DispatchContext);
+
+  const filteredStatus = Object.values(FilterStatus);
+
+  const completedTodosCount = todos.filter(
```
Viktor-Kost, line 16: wrap those operation where you use filter or find (or sth like this) with useMemo hook

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,223 @@
...
+    }
+  };
+
+  const closeErrorMessage = () => {
+    dispatch({ type: 'error', payload: { error: ErrorMessage.None } });
+  };
+
+  useEffect(() => {
+    if (errorMessage) {
+      setTimeout(() => dispatch(
```
Viktor-Kost, line 143: when you use setTimeout or setInterval at useEffect you should clear this operations with `return () => {....}`

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,262 @@
...
+              }}
+            >
+              <input
+                data-cy="TodoTitleField"
+                type="text"
+                className="todo__title-field"
+                placeholder="Empty todo will be deleted"
+                value={newTitle}
+                onChange={e => setNewTitle(e.target.value)}
+                onBlur={handleNewTitleSubmit}
```
Viktor-Kost, line 240: we shouldn't submit value on blur. Since it means that user'd like not to save new data (the same as `escape`). Save data only when user submit new title

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,262 @@
...
+                ×
+              </button>
+            </>
+          )
+          : (
+            <form
+              onSubmit={e => {
+                e.preventDefault();
+                handleNewTitleSubmit();
+              }}
```
Viktor-Kost, line 231: move to an own handler function

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,262 @@
...
+      dispatch({
+        type: 'deleteTodo',
+        payload: { id },
+      });
+    } catch (error) {
+      dispatch({
+        type: 'error',
+        payload: { error: ErrorMessage.Deleting },
+      });
+      setNewTitle(title);
```
Viktor-Kost, line 74: as a question - why do you need to set a new title on error?

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,262 @@
...
+      }
+
+      const updatedTodo = {
+        ...todo,
+        title: newTitle.trim(),
+      };
+
+      if (!newTitle.trim()) {
+        try {
+          await deleteTodo(id);
```
Viktor-Kost, line 100: i'd rather not delete todo if we have no value. You can throw an error that field shouldn't be empty or something and check `!newTitle.trim()` before your create `updatedTodo `

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  const handleKeyUp = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === 'Escape') {
+      setIsUpdatingTitle(false);
+      setNewTitle(title);
+    }
+  };
+
+  useEffect(() => {
+    switch (true) {
```
Viktor-Kost, line 151: hm..it's not obvious solution to work with loading like this. you'd better setIsLoading(true/false) during async operations

Viktor-Kost (changes requested):

src/components/TodoItem.tsx
IrynaK-a: I am a little confused because of the requirements we have:

Please tell me what I should do?

src/components/TodoItem.tsx
IrynaK-a: Sorry, but it's in requirement:

src/components/TodoItem.tsx
IrynaK-a: Oh, it's probable just a typo)

src/components/Footer.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { FilterStatus } from '../types/FilterStatus';
+import { DispatchContext, StateContext } from './TodosProvider';
+import { deleteTodo } from '../api/todos';
+import { ErrorMessage } from '../types/ErrorMessage';
+
+export const Footer = () => {
+  const { filteredBy, todos, newTodoInputRef } = useContext(StateContext);
+  const dispatch = useContext(DispatchContext);
+
+  const filteredStatus = Object.values(FilterStatus);
```
IvanFesenko, line 13: Not mistake, but looks like no need to re create this variable on each render, so it should be moved out of component

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,221 @@
...
+  } = useContext(StateContext);
+  const dispatch = useContext(DispatchContext);
+
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const completedTodosCount = useMemo(() => (
+    todos.filter(
+      todo => todo.completed,
+    ).length
+  ), [todos]);
+
```
IvanFesenko, line 32: Try do not repeat your self. Is possible to pas this value usign pros? Footer component has same logic

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const [title, setTitle] = useState('');
+  const [isFocus, setIsFocus] = useState(false);
+  let isCompletedAll = completedTodosCount === todos.length;
+
+  const toggleAll = () => {
```
IvanFesenko, line 37: Bad practice update values like this

src/components/TodoItem.tsx
IvanFesenko: Agree with comment

src/components/TodoList.tsx
```diff
@@ -0,0 +1,35 @@
...
+          case FilterStatus.Completed:
+            return todo.completed;
+
+          default:
+            return todo;
+        }
+      })
+      : []
+
+  ), [filteredBy, todos]);
```
IvanFesenko, line 26: No need call filter in case FilterStatus is All

IvanFesenko (approved):
Check my comments

PR #983, opened 2023-12-14
Develop

aniashev:
[DEMO LINK](https://aniashev.github.io/react_todo-app-with-api/)

Functionally and visually everything seems to be working as expected, however, many tests do not pass.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,220 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
 import { UserWarning } from './UserWarning';
+import { Error } from './Error';
+import { Header } from './Header';
+import { Todo } from './types/Todo';
+import { TodoList } from './TodoList';
+import { Footer } from './Footer';
+import * as postService from './api/todos';
+import { Status } from './types/Status';
 
-const USER_ID = 0;
+export const USER_ID = 11894;
+
+function getVisibleTodos(todos: Todo[], newStatus: Status) {
+  return todos.filter(({ completed }) => {
+    switch (newStatus) {
+      case Status.ACTIVE:
+        return !completed;
+      case Status.COMPLETED:
+        return completed;
+      default:
+        return true;
+    }
+  });
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [todoTitle, setTodoTitle] = useState('');
+  const [deletingTodoId, setDeletingTodoId] = useState<number | null>(null);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const [filter, setFilter] = useState<Status>(Status.ALL);
+  const visibleTodos = getVisibleTodos(todos, filter);
+
+  const areAllTodosCompleted = useMemo(() => {
+    return todos.every(todo => todo.completed);
+  }, [todos]);
+
+  const clearError = useCallback(() => {
+    const timer = setTimeout(() => {
+      setErrorMessage('');
+    }, 3000);
+
+    return () => clearTimeout(timer);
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,24 +1,226 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
```
loralevitska, line 5:
```suggestion
import React, {
  useCallback, 
  useEffect, 
  useMemo, 
  useState,
} from 'react';
```

src/App.tsx
```diff
@@ -1,24 +1,226 @@
...
+
+function getVisibleTodos(todos: Todo[], newStatus: Status) {
+  return todos.filter(({ completed }) => {
+    switch (newStatus) {
+      case Status.ACTIVE:
+        return !completed;
+      case Status.COMPLETED:
+        return completed;
+      default:
+        return true;
```
loralevitska, line 25:
```suggestion
    switch (newStatus) {
      case Status.ACTIVE:
        return !completed;
        
      case Status.COMPLETED:
        return completed;
        
      default:
        return true;
```

src/App.tsx
```diff
@@ -1,24 +1,226 @@
...
+            setFilter={setFilter}
+            currentFilter={filter}
+            clearCompletedTodos={clearCompletedTodos}
+          />
+        )}
+
+      </div>
+
+      {/* Notification is shown in case of any error */}
+      {/* Add the 'hidden' class to hide the message smoothly */}
```
loralevitska, line 222: Remove comment. Try to write obvious code with meaningful variable and function naming. Fix all cases

src/Error.tsx
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export const Error: React.FC<Props> = ({
+  errorMessage,
+  setErrorMessage = () => {},
+}) => {
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={`notification is-danger is-light has-text-weight-normal ${errorMessage ? '' : 'hidden'}`}
```
loralevitska, line 13: Use [classnames](https://www.npmjs.com/package/classnames) library to add classes conditionally

src/TodoList.tsx
```diff
@@ -0,0 +1,58 @@
...
+  deletingTodoId,
+  isLoading,
+  setIsLoading,
+  tempTodo,
+  toggleTodo,
+  updateTodoList,
+  setErrorMessage,
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
```
loralevitska, line 30: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots and `ul` paddings

loralevitska (changes requested):

- fix disable input after editing todo 
image
- try to pass all tests 
image

aniashev:
1) I am not sure what is meant by: "fix disable input after editing todo". Could you please elaborate? When I am checking now, it seems to be working the same as the example.
2) No matter what I do tests are not passing, even though the app works as it should. Not sure how to fix it(

VitaliyBondarenko1982 (changes requested):
Looks good.
But found a few mismatches with working example
- check styles for todo Item. no visible line between items
actual:
actual

design:
Screenshot from 2023-12-17 19-29-30
- during editing todo no visible loader on todo which in process. also no visible loader during removing or editing multiply todo. loader should be visible on each todo which in process.
Please compare with [working example](https://mate-academy.github.io/react_todo-app-with-api/)

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+            deleteTodo={deleteTodo}
+            deletingTodoId={deletingTodoId}
+            deletingTodoIds={deletingTodoIds}
+            toggleTodo={toggleTodo}
+            updateTodoList={updateTodoList}
+            setErrorMessage={setErrorMessage}
+          />
+        )}
+
+        {(todos.length > 0 || tempTodo) && (
```
volodymyr-soltys97, line 233:
```suggestion
        {(!!todos.length || tempTodo) && (
```

volodymyr-soltys97 (approved):
Great work 🔥

PR #986, opened 2023-12-17
add solution

purpleFade:
[DEMO LINK](https://purplefade.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,68 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback, useContext, useEffect, useMemo,
+} from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { Footer } from './components/Footer/Footer';
+import { getTodos } from './api/todos';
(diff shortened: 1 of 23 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Now the field does not change, I enter new data, but it is not saved.
image

VitaliyBondarenko1982 (changes requested):
Almost done.
Suggest fix some mismatches with  [working example](https://mate-academy.github.io/react_todo-app-with-api/)
- when todo in process (adding, removing, updating) - should be visible loader on todo. if updating or removing multiply todos - loader should be visible on each todo.
- do not show error message when todo deleting instead updating (in case empty title)
todo

IvanFesenko (changes requested):
Good job. 
Check those issues:
2023-12-18_15h58_26
https://github.com/mate-academy/react_todo-app-with-api/assets/61904995/2e88c532-efc0-417a-92f8-82d8655f8224
it's imposible to make updates after first time

VitaliyBondarenko1982 (approved):
Looks good.
Approve this time.
Just still not visible loader on any todo which in process. You can fix it by yourself without re-request.

PR #988, opened 2023-12-19
add task solution

ruslanagnatenko:
[DEMO LINK](https://ruslanagnatenko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,88 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+} from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { getTodos } from './api/todos';
+import { Footer } from './components/Footer';
+import { prepareTodos } from './helpers';
+import { Header } from './components/Header';
+import { ErrorType } from './types/ErrorType';
+import { ErrorNotification } from './components/ErrorNotification';
+import { TodoList } from './components/TodoList';
+import { AppContext } from './AppContext';
+import { USER_ID } from './userId/userId';
 
 export const App: React.FC = () => {
+  const {
+    todos,
+    setTodos,
+    status,
+    setStatus,
+    error,
+    setError,
+  } = useContext(AppContext);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => setError(ErrorType.cantLoadTodos));
+  }, [setTodos, setError]);
+
+  const filterTodos = useCallback(() => {
+    return prepareTodos({
+      todos,
+      status,
+    });
+  }, [todos, status]);
+
+  const todosOnPage = useMemo(() => filterTodos(),
+    [filterTodos]);
+
+  const uncompletedTodosCount = todos.filter(todo => (
+    !todo.completed
+  )).length;
+
+  const isSomeTodosCompleted = todos.some(
+    todo => todo.completed,
+  );
+
+  const isEveryTodosCompleted = todosOnPage.every(
+    todo => todo.completed,
+  );
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <Header
+          isEveryTodosCompleted={isEveryTodosCompleted}
+        />
+
+        <TodoList
+          todos={todosOnPage}
+        />
+
+        {todos.length > 0 && (
+          <Footer
+            uncompletedTodosCount={uncompletedTodosCount}
+            status={status}
+            setStatus={setStatus}
+            isSomeTodosCompleted={isSomeTodosCompleted}
+          />
+        )}
+
+        {error && todos.length > 0 && (
+          <ErrorNotification />
+        )}
+      </div>
+    </div>
   );
 };
--- a/src/AppContext.tsx
+++ b/src/AppContext.tsx
@@ -1,0 +1,278 @@
+import React, {
+  createContext,
+  useCallback,
+  useMemo,
+  useState,
+} from 'react';
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+import * as todoService from './api/todos';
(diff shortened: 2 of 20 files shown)


src/AppContext.tsx
```diff
@@ -0,0 +1,278 @@
...
+      completed: false,
+    })
+      .then(newTodo => {
+        setTodoTitle('');
+        setTimeout(() => {
+          setTodos(currentTodos => [...currentTodos, newTodo]);
+        }, 500);
+      })
+      .catch(() => {
+        setSelectedTodoIds(ids => ids.filter(
```
mykhailoyurkovskyi, line 189: You do it in finally block

mykhailoyurkovskyi (approved):
Looks good for me✔👌

src/App.tsx
```diff
@@ -1,24 +1,88 @@
...
+  }, [setTodos, setError]);
+
+  const filterTodos = useCallback(() => {
+    return prepareTodos({
+      todos,
+      status,
+    });
+  }, [todos, status]);
+
+  const todosOnPage = useMemo(() => filterTodos(),
```
BudnikOleksii, line 41:
```suggestion
  const todosOnPage = useMemo(() => prepareTodos({
      todos,
      status,
    }),
```
`filterTodos` is redundant in this case, you can do it in `todosOnPage` and avoid creating a wrapper for `prepareTodos`

src/App.tsx
```diff
@@ -1,24 +1,88 @@
...
 
 export const App: React.FC = () => {
+  const {
+    todos,
+    setTodos,
+    status,
+    setStatus,
+    error,
+    setError,
+  } = useContext(AppContext);
```
BudnikOleksii, line 26: You can create a custom hook to avoid passing context to useContext function every time
```typescript
const useAppContext = () => useContext(AppContext);
```

src/App.tsx
```diff
@@ -1,24 +1,88 @@
...
+        {todos.length > 0 && (
+          <Footer
+            uncompletedTodosCount={uncompletedTodosCount}
+            status={status}
+            setStatus={setStatus}
+            isSomeTodosCompleted={isSomeTodosCompleted}
+          />
+        )}
+
+        {error && todos.length > 0 && (
```
BudnikOleksii, line 82: If there are no todos an error still can occur, for example when loading todos

src/AppContext.tsx
```diff
@@ -0,0 +1,292 @@
...
+            setTodos(currentTodos => currentTodos
+              .filter(post => post.id !== todoId));
+          }, 500);
+        })
+        .catch(() => {
+          setSelectedTodoIds(ids => {
+            ids.splice(ids.indexOf(todoId), 1);
+
+            return ids;
+          });
```
BudnikOleksii, line 87: - You need to do it in finally block only as you should remove loaders in both cases
- don't use splice + indexOf, use filter instead

src/AppContext.tsx
```diff
@@ -0,0 +1,292 @@
...
+          setError(ErrorType.cantDeleteTodo);
+          setTimeout(() => setError(null), 2000);
```
BudnikOleksii, line 89: You set the same timeout every time when error occurs, create a function
```typescript
const handleError = (errorMessage: ErrorType) => {
  setError(errorMessage);
  setTimeout(() => setError(null), 2000);
};
```

src/AppContext.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+      setSelectedTodoIds(currentIds => (
+        [...currentIds, ...completedTodos
+          .map(completedTodo => completedTodo.id),
+        ]
+      ));
+
+      completedTodos.map(
+        completedTodo => todoService.deleteTodo(completedTodo.id),
+      );
```
BudnikOleksii, line 112: What if requests fail? You can wait for a response with `Promise.all` or `Promise.allSettled`

src/AppContext.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+      todoService.setCompleted(updatedTodo)
+        .then(todo => {
+          setTodos(currentTodos => {
+            const newTodos = [...currentTodos];
+            const index = newTodos.findIndex(newTodo => (
+              newTodo.id === updatedTodo.id
+            ));
+
+            newTodos.splice(index, 1, todo);
```
BudnikOleksii, line 138: Just use filter instead, fix it in all places

src/AppContext.tsx
BudnikOleksii: > You do it in finally block

Good point, fix it in all places

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,125 @@
...
+export const TodoItem: React.FC<Props> = React.memo(
+  ({ todo }) => {
+    const {
+      deleteTodo,
+      selectedTodoIds,
+      handleToggleCompleted,
+      updateTodo,
+    } = useContext(AppContext);
+
+    const [editing, setEditing] = useState(false);
```
BudnikOleksii, line 24: Boolean variables should have prefix `is` or `has`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,125 @@
...
+              placeholder="Empty todo will be deleted"
+              ref={inputRef}
+              value={newTitle}
+              onChange={(e) => setNewTitle(e.target.value)}
+              onKeyUp={(e) => {
+                if (e.key === 'Escape') {
+                  setEditing(false);
+                  setNewTitle(todo.title);
+                }
+              }}
```
BudnikOleksii, line 93: Create a handler instead of passing an anonim function to `onKeyUp`

BudnikOleksii (changes requested):
Looks good 🔥 
But could be even better, left some suggestions

src/AppContext.tsx
```diff
@@ -0,0 +1,297 @@
...
+          setTodos(currentTodos => {
+            const newTodos = [...currentTodos];
+            const index = newTodos.findIndex(newTodo => (
+              newTodo.id === updatedTodo.id
+            ));
+
+            newTodos[index] = todo;
+
+            return newTodos;
+          });
```
BudnikOleksii, line 152:
```suggestion
          setTodos(currentTodos => {
            return currentTodos .map(newTodo=> newTodo.id === updatedTodo.id ? todo : newTodo);
          });
```
Use `map` instead

src/AppContext.tsx
```diff
@@ -0,0 +1,297 @@
...
+};
+
+export const AppProvider: React.FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState<Status>(Status.all);
+  const [selectedTodoIds, setSelectedTodoIds] = useState<number[]>([]);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [error, setError] = useState<ErrorType | null>(null);
+  const [loadind, setLoading] = useState(false);
```
BudnikOleksii, line 73: Boolean variables should have prefix `is` or `has`

BudnikOleksii (approved):
Nicely done 🚀

PR #993, opened 2023-12-21
added task solution

Yevhen-Kolosov:
[DEMO LINK](https://Yevhen-Kolosov.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,345 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, {
+  Dispatch,
+  FormEvent,
+  SetStateAction,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 
-const USER_ID = 0;
+import { TodoMain } from './components/TodoMain';
+import { TodoHeader } from './components/TodoHeader';
+import { FilterType } from './types/FilterType';
+import { TodoNotification } from './components/TodoNotification';
+import { TodoFooter } from './components/TodoFooter';
+import { Todo } from './types/Todo';
+import {
+  addTodo,
+  deleteTodo,
+  getTodos,
+  patchTodo,
+} from './api/todos';
+import { ErrorType } from './types/ErrorType';
 
-export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+const USER_ID = 12031;
+
+const filterTodos = (sourceTodos: Todo[], filterBy: FilterType) => {
+  let filteredTodos: Todo[];
+
+  switch (filterBy) {
+    case FilterType.Active:
+      filteredTodos = sourceTodos.filter(todo => !todo.completed);
+      break;
+
+    case FilterType.Completed:
+      filteredTodos = sourceTodos.filter(todo => todo.completed);
+      break;
+
+    default:
+      filteredTodos = [...sourceTodos];
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+  return filteredTodos;
+};
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+export const App: React.FC = () => {
+  const [
+    selectedFilter,
+    setSelectedFilter,
+  ] = useState<FilterType>(FilterType.All);
+
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [selectedTodoIds, setSelectedTodoIds] = useState<number[]>([]);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const [
+    currentError,
+    setCurrentError,
+  ] = useState<ErrorType>(ErrorType.NoError);
+
+  const timerId = useRef<NodeJS.Timeout>();
+
+  const onError = (error: ErrorType) => {
+    if (timerId.current) {
+      clearTimeout(timerId.current);
+    }
+
+    setCurrentError(error);
+    timerId.current = setTimeout(
+      () => setCurrentError(ErrorType.NoError),
+      3000,
+    );
+  };
+
+  const handleSubmit = async (
+    event: FormEvent<HTMLFormElement>,
+    newTodoText: string,
+    setNewTodoText: Dispatch<SetStateAction<string>>,
+  ) => {
+    event.preventDefault();
+
+    if (!newTodoText.trim()) {
+      onError(ErrorType.EmptyTitle);
+
+      return;
+    }
+
+    const newTodo = {
+      id: 0,
+      userId: USER_ID,
+      title: newTodoText.trim(),
+      completed: false,
+    };
+
+    try {
+      setTempTodo(() => newTodo);
+      await addTodo(newTodo)
+        .then((todo) => setTodosFromServer(
+          prevTodos => ([...prevTodos, todo]),
+        ))
+        .then(() => {
+          setNewTodoText('');
+        });
+    } catch {
+      onError(ErrorType.UnableToAdd);
+    } finally {
+      setTempTodo(() => null);
+
+      setTimeout(() => {
+        if (inputRef.current) {
+          inputRef.current.focus();
+        }
+      }, 100);
+    }
+  };
+
+  const handleSingleTodoDelition = async (todoId: number) => {
+    try {
+      setSelectedTodoIds(prevIds => [...prevIds, todoId]);
+
+      await deleteTodo(todoId).then(() => {
+        setTodosFromServer(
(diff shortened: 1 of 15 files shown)


src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { Dispatch, SetStateAction } from 'react';
+import cn from 'classnames';
+import { FilterType } from '../../types/FilterType';
+
+type Props = {
+  activeTodosCount: number,
+  isAnyTodoCompleted: boolean,
+  selectedFilter: FilterType,
+  setSelectedFilter: Dispatch<SetStateAction<FilterType>>,
```
Anton-Kuchmasov, line 9: nice to see Context in your solution)

Anton-Kuchmasov (approved):
nice job you did!

src/App.tsx
```diff
@@ -1,24 +1,345 @@
...
+    .filter(todo => todo.completed)
+    .map(todo => todo.id);
+
+  const handleCompletedTodosDelition = async () => {
+    const processedIds: number[] = [];
+
+    try {
+      setSelectedTodoIds(prev => [...prev, ...completedTodoIds]);
+
+      completedTodoIds.forEach(async (todoId) => {
```
BudnikOleksii, line 160: `async` doesn't have any effect in `forEach`. If you want to proceed a batch of requests in the same time use `Promise.all`.
If you want to do them one by one you can do it with `forof`. Example with forof:
```javascript
for (const todoId of completedTodoIds) {
  await deleteTodo(todoId);
}
```

src/App.tsx
```diff
@@ -1,24 +1,345 @@
...
+    } finally {
+      setTimeout(
+        () => {
+          setSelectedTodoIds(
+            ids => ids.filter(id => !completedTodoIds.includes(id)),
+          );
+
+          setTodosFromServer(
+            todos => todos.filter(todo => !processedIds.includes(todo.id)),
+          );
```
BudnikOleksii, line 177: Remove todos from an array you should do only in positive case instaed of finally method

src/App.tsx
```diff
@@ -1,24 +1,345 @@
...
+      setSelectedTodoIds(prevIds => [...prevIds, updatedTodo.id]);
+
+      await patchTodo(updatedTodo)
+        .then((todoFromResponse: Todo) => {
+          const todos = todosFromServer;
+          const indexOfTodo = todos.findIndex(
+            todo => todo.id === todoFromResponse.id,
+          );
+
+          todos[indexOfTodo] = todoFromResponse;
```
BudnikOleksii, line 199: Use `map` method instead

src/App.tsx
```diff
@@ -1,24 +1,345 @@
...
+        })
+        .catch(() => onError(ErrorType.UnableToUpdate));
+    } catch {
+      onError(ErrorType.UnableToUpdate);
+    } finally {
+      setSelectedTodoIds(
+        ids => {
+          const idsCopy = [...ids];
+
+          idsCopy.splice(idsCopy.indexOf(updatedTodo.id), 1);
```
BudnikOleksii, line 211: Get rid of `splice`/`indexOf` in your code

src/App.tsx
```diff
@@ -1,24 +1,345 @@
...
+            ...todo,
+            completed: true,
+          };
+
+          await patchTodo(updatedTodo).then((todoFromResponse: Todo) => {
+            const indexOfTodo = todos.findIndex(
+              td => td.id === todoFromResponse.id,
+            );
+
+            todos[indexOfTodo] = todoFromResponse;
```
BudnikOleksii, line 253: The same mistakes, fix them in all places

src/components/TodoFooter/TodoFooter.tsx
BudnikOleksii: > nice to see Context in your solution)

It's not `Context` 👀

src/components/TodoMain/TodoItem.tsx
```diff
@@ -0,0 +1,144 @@
...
+      return;
+    }
+
+    try {
+      updateTodo(updatedTodo);
+      setIsBeingEdited(false);
+    } catch {
+      onError(ErrorType.UnableToUpdate);
+      setIsBeingEdited(true);
+    }
```
BudnikOleksii, line 73: - If you handle errors in the `updateTodo` function you don't need to do the same here.
- `setIsBeingEdited(false);` will be executed before than async function fihish
- `setIsBeingEdited(false)` you do it in both cases, operations that you do anyway should be placed in finally block.

So looks like it will be enough:
```javascript
await updateTodo(updatedTodo);
setIsBeingEdited(false);
```

BudnikOleksii (changes requested):
Nice start 👍 
Left comments to improve your solution, check them

ksyvvkaa:
stickgfdser

src/App.tsx
BudnikOleksii: > `async` doesn't have any effect in `forEach`. If you want to proceed a batch of requests in the same time use `Promise.all`. If you want to do them one by one you can do it with `forof`. Example with forof:
> 
> ```js
> for (const todoId of completedTodoIds) {
>   await deleteTodo(todoId);
> }
> ```

Not fixed, a few links about async in forEach:
https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971
https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop

src/App.tsx
BudnikOleksii: > Remove todos from an array you should do only in positive case instaed of finally method

Not fixed

src/App.tsx
```diff
@@ -1,24 +1,283 @@
...
+          if (td.id === todoFromResponse.id) {
+            return todoFromResponse;
+          }
+
+          return td;
+        }),
+      );
+    } catch {
+      onError(ErrorType.UnableToUpdate);
+      throw new Error(ErrorType.UnableToUpdate);
```
BudnikOleksii, line 185: You don't need to throw an error as it's already shown for the user

src/App.tsx
```diff
@@ -1,24 +1,283 @@
...
+  const activeTodoIds = activeTodos.map(todo => todo.id);
+  const activeTodosCount = activeTodos.length;
+
+  const handleToggleAll = () => {
+    if (activeTodosCount) {
+      setSelectedTodoIds(ids => [...ids, ...activeTodoIds]);
+
+      activeTodos.forEach(async (todo) => {
+        try {
+          const updatedTodo = {
```
BudnikOleksii, line 215: Fix `forEach` with `async` usage in all places

BudnikOleksii (changes requested):
Good work, it's almost done! Just a few more fixes that should be done

src/App.tsx
Yevhen-Kolosov: I've read the articles you've sent before and those refer to a situation, when calllback function (that is passed to forEach) doesn't have a proper inner structure async (try-catch blocks are missing) hence they can't handle promises. But in my case, the async structure is followed so the promises resolve as expected KEKWait (2)

src/App.tsx
Yevhen-Kolosov: The error throw is needed as the function is being reused in another handler that needs to catch an error (`onSumbitTitle` in `TodoItem.tsx`) KEKWait (2)

src/App.tsx
roman-mirzoian: Not fixed 😢

src/App.tsx
```diff
@@ -1,24 +1,276 @@
...
+          if (td.id === todoFromResponse.id) {
+            return todoFromResponse;
+          }
+
+          return td;
+        }),
+      );
+    } catch {
+      onError(ErrorType.UnableToUpdate);
+      throw new Error();
```
roman-mirzoian, line 179: There is no point in using an empty `Error`.

src/App.tsx
roman-mirzoian: Not fixed 😢 
@Yevhen-Kolosov those articles generally advise against using `forEach` with asynchronous code

src/App.tsx
```diff
@@ -1,24 +1,276 @@
...
+
+          await updateTodo(updatedTodo);
+        } catch {
+          onError(ErrorType.UnableToUpdate);
+        }
+      });
+    } else {
+      setSelectedTodoIds(ids => [...ids, ...completedTodoIds]);
+
+      todosFromServer.forEach(async (todo) => {
```
roman-mirzoian, line 222: async forEach 😢

src/components/TodoMain/TodoMain.tsx
```diff
@@ -0,0 +1,45 @@
...
+          />
+        );
+      })}
+
+      {tempTodo && (
+        <TodoItem
+          todo={tempTodo}
+          selectedTodoIds={selectedTodoIds}
+          onDelete={() => {}}
+          updateTodo={async () => {}}
```
roman-mirzoian, line 39: If you make these fields optional, you won't need to use unnecessary empty functions.

src/types/ErrorType.ts
```diff
@@ -0,0 +1,8 @@
+export enum ErrorType {
+  NoError = '',
```
roman-mirzoian, line 2: If there are no errors, just set the error state to `null`, no need to store extra enum

roman-mirzoian (changes requested):
There are still some unresolved issues 😢

BudnikOleksii (approved):
👍

PR #994, opened 2023-12-21
Develop

Anton-Kuchmasov:
[DEMO LINK](https://Anton-Kuchmasov.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,168 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useMemo, useState } from 'react';
+import { UserWarning } from './components/UserWarning/UserWarning';
+import { client } from './utils/fetchClient';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+import { Header } from './components/Header/Header';
+import { Footer } from './components/Footer/Footer';
+import { Errors } from './components/Errors/Errors';
+import { Status } from './types/Status';
+import { ErrorSpec } from './types/ErrorSpec';
+import { filterTodos } from './helpers/filterTodos';
+import * as TodoService from './services/todo';
 
-const USER_ID = 0;
+const USER_ID = 12021;
+const ADDED_URL = `/todos?userId=${USER_ID}`;
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [hasErrors, setHasErrors] = useState(false);
+  const [error, setError] = useState<ErrorSpec | null>(null);
+  const [status, setStatus] = useState<Status>(Status.ALL);
+  const [uncompletedTodosCount, setUncompletedTodosCount] = useState<number>(0);
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodo, setLoadingTodo] = useState<Todo | null>(null);
+
+  const filteredTodos = useMemo(
+    () => filterTodos(todosFromServer, status),
+    [status, todosFromServer],
+  );
+
+  useEffect(() => {
+    client
+      .get<Todo[]>(ADDED_URL)
+      .then((todos) => {
+        setTodosFromServer(todos);
+
+        if (todos.length === 0) {
+          setError(ErrorSpec.EMPTY_TITLE);
+        }
+      })
+      .catch(() => setError(ErrorSpec.NOT_LOADED));
+  }, []);
+
+  useEffect(() => {
+    const completedCount = todosFromServer.filter(
+      (todo) => !todo.completed,
+    ).length;
+
+    setUncompletedTodosCount(completedCount);
+  }, [todosFromServer, status, filteredTodos]);
+
+  const handleStatus = (newStatus: Status) => {
+    setStatus(newStatus);
+  };
+
+  const cleanErrors = () => {
+    setHasErrors(false);
+  };
+
+  const addTodo = (title: string) => {
+    setIsInputDisabled(true);
+    const preparedTitle = title.trim();
+
+    if (!preparedTitle) {
+      setError(ErrorSpec.EMPTY_TITLE);
+      setHasErrors(true);
+      setIsInputDisabled(false);
+
+      return;
+    }
+
+    const newTodo: Todo = {
+      id: 0,
+      userId: USER_ID,
+      title,
+      completed: false,
+    };
+
+    TodoService.createTodo(newTodo)
+      .then((responce) => {
+        setTempTodo(responce);
+        setTodosFromServer([...todosFromServer, responce]);
+      })
+      .catch(() => {
+        setError(ErrorSpec.NOT_ADDED);
+        setIsInputDisabled(false);
+      })
+      .finally(() => {
+        setIsInputDisabled(false);
+        setTempTodo(null);
+      });
+  };
+
+  const removeTodo = (id: number) => {
+    setIsInputDisabled(true);
+    setLoadingTodo(filteredTodos.find((todo) => todo.id === id) || null);
+
+    TodoService.deleteTodo(id)
+      .then(() => setLoadingTodo(null))
+      .catch(() => {
+        setError(ErrorSpec.NOT_DELETED);
+        setIsInputDisabled(false);
+      })
+      .finally(() => {
+        setTodosFromServer((prev) => prev.filter((todo) => todo.id !== id));
+        setIsInputDisabled(false);
+      });
+  };
+
+  const clearCompleted = () => {
+    filteredTodos.forEach(element => {
+      if (element.completed) {
+        removeTodo(element.id);
+      }
+    });
+    setUncompletedTodosCount(0);
+    setHasErrors(false);
+  };
+
+  useEffect(() => {
+    setLoadingTodo(
+      filteredTodos.find((todo) => todo.id === loadingTodo?.id) || null,
+    );
+  }, [filteredTodos, loadingTodo]);
+
   if (!USER_ID) {
+    setHasErrors(true);
+
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <Header
+          onSubmit={addTodo}
+          onInput={cleanErrors}
+          inputDisabled={isInputDisabled}
+          hasErrors={hasErrors}
+        />
+
+        {todosFromServer.length > 0 && (
+          <>
+            <TodoList
+              todos={filteredTodos}
+              tempTodo={tempTodo}
+              deleteTodo={removeTodo}
+              isProcessing={loadingTodo}
+            />
+            <Footer
+              onStatus={handleStatus}
+              status={status}
+              completedCount={uncompletedTodosCount}
+              handleClear={clearCompleted}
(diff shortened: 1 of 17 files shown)


mickhakk (approved):

src/components/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+type Props = {
+  error: ErrorSpec | null;
+};
+
+export const ErrorMessage: React.FC<Props> = ({ error }) => {
+  return (
+    <span>
+      {ErrorText(error)}
+      <br />
```
artemforsoff, line 12: It’s better not to use this `<br />` tag
Because if you need some kind of indentation, then CSS is used for this

src/components/Errors/Errors.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className="notification is-danger is-light has-text-weight-normal"
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        aria-label="none"
```
artemforsoff, line 19: `<button>none</button>` used for what?
the value for the aria-attribute should answer the question what this button does
this button closes the error, so `aria-label="Hide error"` will be better

src/components/Errors/Errors.tsx
artemforsoff, line 1: the component is named in the plural, but only one error is displayed in the component, although it is assumed that there is more than one

src/components/Errors/Errors.tsx
```diff
@@ -0,0 +1,26 @@
+import { ErrorSpec } from '../../types/ErrorSpec';
+import { ErrorMessage } from '../ErrorMessage/ErrorMessage';
```
artemforsoff, line 2: I wouldn't move the logic to a separate component for picking an error

src/components/Errors/Errors.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+type Props = {
+  error: ErrorSpec | null;
+  closeError: () => void;
```
artemforsoff, line 6: if a function is called on some event, then the prop should be named with the prefix `on`
like this `onHideError`

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,72 @@
+import classNames from 'classnames';
+import React from 'react';
+import { Status } from '../../types/Status';
+
+type Props = {
+  onStatus: (status: Status) => void;
```
artemforsoff, line 6: the name does not say when the function will be called
therefore, it is better to rename it to `onChangeStatus`

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,72 @@
...
+  status: Status;
+  completedCount: number;
+  isClearNeeded: boolean;
+  handleClear: () => void;
```
artemforsoff, line 10: `onClearCompleted`

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+type Props = {
+  todos: Todo[];
+  tempTodo: Todo | null;
+  deleteTodo: (id: number) => void;
+  isProcessing: Todo | null;
+  onToggleCompleted: (todo: Todo) => void;
+  isEditing: Todo | null;
+  onEditTodo: (todo: Todo | null) => void;
+  handleSaveTodo: (updatedTitle: string, todo: Todo) => void;
```
artemforsoff, line 13: `onSaveTodo`

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,55 @@
...
+          onEditTodo={onEditTodo}
+          handleSaveTodo={handleSaveTodo}
+        />
+      ))}
+
+      {!!tempTodo && (
+        <TodoItem
+          key={tempTodo.id}
+          todo={tempTodo}
+          deleteTodo={() => {}}
```
artemforsoff, line 45: it's not required to do better

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,108 @@
+import React, { useEffect, useRef, useState } from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type TodoItemProps = {
+  todo: Todo;
+  deleteTodo: (id: number) => void;
```
artemforsoff, line 7: prefix `on`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,108 @@
...
+  isProcessing: Todo | null;
+  onToggleCompleted: (todo: Todo) => void;
+  isEditing: Todo | null;
+  onEditTodo: (todo: Todo | null) => void;
+  handleSaveTodo: (titleUpdated: string, todo: Todo) => void;
```
artemforsoff, line 12: prefix `on` same as before

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+import { ErrorSpec } from './types/ErrorSpec';
+import { filterTodos } from './helpers/filterTodos';
+import * as TodoService from './services/todo';
 
-const USER_ID = 0;
+const USER_ID = 12021;
+const ADDED_URL = `/todos?userId=${USER_ID}`;
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
```
artemforsoff, line 18: it is not necessary to indicate where the data came from

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+  const areAllCompleted = uncompletedTodosCount === 0;
+
+  const filteredTodos = useMemo(
+    () => filterTodos(todosFromServer, status),
+    [status, todosFromServer],
+  );
+
+  useEffect(() => {
+    client
+      .get<Todo[]>(ADDED_URL)
```
artemforsoff, line 37: better move to `src/services/todo.ts`

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+import { filterTodos } from './helpers/filterTodos';
+import * as TodoService from './services/todo';
 
-const USER_ID = 0;
+const USER_ID = 12021;
+const ADDED_URL = `/todos?userId=${USER_ID}`;
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [hasErrors, setHasErrors] = useState(false);
```
artemforsoff, line 19: it is not necessary to create a separate state to determine whether there is an error or not,
since you have an error state that stores a `ErrorSpec | null`, so we can determine whether there is an error or not by the presence of a string like this:
`const hasError = Boolean(error);`

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+  const [error, setError] = useState<ErrorSpec | null>(null);
+  const [status, setStatus] = useState<Status>(Status.ALL);
+  const [uncompletedTodosCount, setUncompletedTodosCount] = useState<number>(0);
```
artemforsoff, line 22: you can use `useMemo`

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+    [status, todosFromServer],
+  );
+
+  useEffect(() => {
+    client
+      .get<Todo[]>(ADDED_URL)
+      .then((todos) => {
+        setTodosFromServer(todos);
+
+        if (todos.length === 0) {
```
artemforsoff, line 41: if the server responded without an error and returned an empty array of entities, this is not an error

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+
+    const newTodo: Todo = {
+      id: 0,
+      userId: USER_ID,
+      title,
+      completed: false,
+    };
+
+    TodoService.createTodo(newTodo)
+      .then((responce) => {
```
artemforsoff, line 84: `responSe`
I suggest installing [Code Spell Checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker)

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+        setTempTodo(responce);
```
artemforsoff, line 85: there is no point in adding to this state because at the end you will set `null`

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+        setTodosFromServer([...todosFromServer, responce]);
+      })
+      .catch(() => {
+        setError(ErrorSpec.NOT_ADDED);
+        setIsInputDisabled(false);
```
artemforsoff, line 90: there is no point in changing to this state because at the end you will change

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+    setLoadingTodo(filteredTodos.find((todo) => todo.id === id) || null);
+
+    TodoService.deleteTodo(id)
+      .then(() => setLoadingTodo(null))
+      .catch(() => {
+        setError(ErrorSpec.NOT_DELETED);
+        setIsInputDisabled(false);
+      })
+      .finally(() => {
+        setTodosFromServer((prev) => prev.filter((todo) => todo.id !== id));
```
artemforsoff, line 109: this logic should be in the `then` block
because if I delete and the request returns an error, the entity will be removed from the screen

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+      });
+  };
+
+  const clearCompleted = () => {
+    filteredTodos.forEach((element) => {
+      if (element.completed) {
+        removeTodo(element.id);
+      }
+    });
+    setUncompletedTodosCount(0);
```
artemforsoff, line 120: requests for deletion may fall and therefore you cannot know for sure that this value will be `0`

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+    TodoService.updateTodo(todo)
+      .then(() => {
+        setLoadingTodo(todo);
+      })
+      .catch(() => {
+        setError(ErrorSpec.NOT_UPDATED);
+        setLoadingTodo(null);
+      })
+      .finally(() => {
+        const todosAfterUpdate = filteredTodos.map(
```
artemforsoff, line 138: this logic should be in the `then` block
because if I change title and the request returns an error, the entity will be changed on the screen

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+    const todoUpdated = { ...todo, completed: !todo.completed };
+
+    updateTodo(todoUpdated);
+  };
+
+  const onToggleAll = () => {
+    const toggledTodos = todosFromServer.map((todo) => (areAllCompleted
+      ? {
+        ...todo,
+        completed: !todo.completed,
```
artemforsoff, line 157: `completed: areAllCompleted ? !todo.completed : false,`

src/App.tsx
```diff
@@ -1,24 +1,246 @@
...
+  };
+
+  useEffect(() => {
+    setLoadingTodo(
+      filteredTodos.find((todo) => todo.id === loadingTodo?.id) || null,
+    );
+  }, [filteredTodos, loadingTodo]);
+
   if (!USER_ID) {
+    setHasErrors(true);
```
artemforsoff, line 201: you can't call like that
the setter call must be in some function or event handler

artemforsoff (changes requested):
good
I left comments, please pay attention to them
and not implemented:

_1. show the Unable to update a todo notification in case of API error
2. the status should be changed on success;
3. update the todo title on success;
4. show Unable to update a todo in case of API error;_

Anton-Kuchmasov:
> good I left comments, please pay attention to them and not implemented:
> 
> _1. show the Unable to update a todo notification in case of API error 2. the status should be changed on success; 3. update the todo title on success; 4. show Unable to update a todo in case of API error;_

what do you mean?
when i try to update ToDo title or toggle the 'completed', it works

src/components/ErrorMessage/ErrorMessage.tsx
Anton-Kuchmasov: it was into basic code of that task

src/App.tsx
Anton-Kuchmasov: installed, thx a lot)

src/App.tsx
Anton-Kuchmasov: it should be in case of showing loader while response didn't finish yet

roman-mirzoian (changes requested):
Fix all problems from the previous review.

roman-mirzoian (approved):
Well done 👍

PR #995, opened 2023-12-21
solution

OstinM:
- [DEMO LINK](https://OstinM.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,184 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback,
+  useEffect, useMemo, useState,
+} from 'react';
 import { UserWarning } from './UserWarning';
+import { TodoList } from './components/TodoList/TodoList';
+import { Header } from './components/Header/Header';
+import { Footer } from './components/Footer/Footer';
+import * as todoService from './api/todos';
+import { Todo } from './types/Todo';
+import { FilteredBy, filteredTodoList } from './helpers';
+import { Errors } from './components/Errors/Errors';
+import { ErrorType } from './types/ErorTypes';
 
-const USER_ID = 0;
+const USER_ID = 12036;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState<FilteredBy>(FilteredBy.DefaultType);
+  const [errorType, setErorType] = useState<ErrorType | null>(null);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [isLoadingTodo, setIsLoadingTodo] = useState<Todo | null>(null);
+
+  const todosToView = useMemo(
+    () => filteredTodoList(todos, filterBy),
+    [todos, filterBy],
+  );
+
+  const deleteTodo = useCallback(
+    async (todoId: number) => {
+      try {
+        const todoToRemove = todos.find((todo) => todo.id === todoId) || null;
+
+        setIsLoadingTodo(todoToRemove);
+
+        await todoService.removeTodo(todoId);
+
+        setTodos((currentTodos) => currentTodos.filter(
+          (todo) => todo.id !== todoId,
+        ));
+      } catch (error) {
+        setErorType(ErrorType.DeleteError);
+      }
+    },
+    [todos, setIsLoadingTodo, setTodos, setErorType],
+  );
+
+  const addTodo = useCallback(
+    async (todoTitle: string) => {
+      try {
+        setIsLoading(true);
+        const newTempTodo: Todo = {
+          id: 0,
+          userId: USER_ID,
+          title: todoTitle,
+          completed: false,
+        };
+
+        const addedTodo = await todoService.addTodo({
+          title: todoTitle,
+          completed: false,
+          userId: USER_ID,
+        });
+
+        setTempTodo(newTempTodo);
+
+        setTodos((currentTodos) => [...currentTodos, addedTodo]);
+
+        setIsLoading(false);
+      } catch (error) {
+        setErorType(ErrorType.AddError);
+      } finally {
+        setTempTodo(null);
+        setNewTodoTitle('');
+        setIsLoading(false);
+      }
+    },
+    [setIsLoading, setTempTodo, setTodos, setErorType, setNewTodoTitle],
+  );
+
+  const updateTodo = useCallback(
+    async (id: number, todo: Todo) => {
+      try {
+        setIsLoading(true);
+
+        await todoService.updateTodo(id, { ...todo });
+
+        setTodos((currentTodos) => {
+          const localUpdatedTodo = [...currentTodos];
+          const index = localUpdatedTodo.findIndex((t) => t.id === id);
+
+          localUpdatedTodo.splice(index, 1, todo);
+
+          return localUpdatedTodo;
+        });
+      } catch (error) {
+        setErorType(ErrorType.UpdateError);
+      } finally {
+        setIsLoading(false);
+      }
+    },
+    [setIsLoading, setTodos, setErorType],
+  );
+
+  const clearCompleted = () => {
+    todosToView.forEach(todo => {
+      if (todo.completed) {
+        deleteTodo(todo.id);
+      }
+    });
+
+    setErorType(null);
+  };
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        setErorType(null);
+        const todosFromServer = await todoService.getTodos(USER_ID);
+
+        setTodos(todosFromServer);
+      } catch (error) {
+        setErorType(ErrorType.LoadError);
+      }
+    };
+
+    fetchData();
+  }, []);
+
+  const closeError = useCallback(() => setErorType(null), []);
+  const isAllTodosCompleted = todos.every(todo => todo.completed);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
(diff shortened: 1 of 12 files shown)


yefimov77 (approved):
Good Job!!

src/App.tsx
```diff
@@ -1,24 +1,186 @@
...
+const USER_ID = 12036;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState<FilteredBy>(FilteredBy.DefaultType);
+  const [errorType, setErorType] = useState<ErrorType | null>(null);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [isLoadingTodo, setIsLoadingTodo] = useState<Todo | null>(null);
```
BudnikOleksii, line 26: Why do you need two loading states and in one you store todo?
It looks strange, you can store an array of todos ids that currently loading to show loaders for them

src/App.tsx
```diff
@@ -1,24 +1,186 @@
...
+          completed: false,
+        };
+
+        const addedTodo = await todoService.addTodo({
+          title: todoTitle,
+          completed: false,
+          userId: USER_ID,
+        });
+
+        setTempTodo(newTempTodo);
```
BudnikOleksii, line 69:
```suggestion
        setTempTodo(newTempTodo);
        
        const addedTodo = await todoService.addTodo({
          title: todoTitle,
          completed: false,
          userId: USER_ID,
        });
```
You should show temp todo before the request succeeded, as now you show it only after you have a new todo and then delete it

src/App.tsx
```diff
@@ -1,24 +1,186 @@
...
+
+        setTodos((currentTodos) => [...currentTodos, addedTodo]);
+
+        setIsLoading(false);
+      } catch (error) {
+        setErorType(ErrorType.AddError);
+      } finally {
+        setTempTodo(null);
+        setNewTodoTitle('');
+        setIsLoading(false);
```
BudnikOleksii, line 79: - You remove the loader in both cases, so it should be only in finally block
- Also, when I'm adding the todo I see loader on all todos, but it should be only for the temp todo

src/App.tsx
```diff
@@ -1,24 +1,186 @@
...
+    },
+    [setIsLoading, setTempTodo, setTodos, setErorType, setNewTodoTitle],
+  );
+
+  const updateTodo = useCallback(
+    async (id: number, todo: Todo) => {
+      try {
+        setIsLoading(true);
+
+        await todoService.updateTodo(id, { ...todo });
```
BudnikOleksii, line 90:
```suggestion
        const updatedTodos = await todoService.updateTodo(id, { ...todo });
```
You should store and then set updated todo from the server, as maybe there were other changes with it too.

src/App.tsx
```diff
@@ -1,24 +1,186 @@
...
+
+        setTodos((currentTodos) => {
+          const localUpdatedTodo = [...currentTodos];
+          const index = localUpdatedTodo.findIndex((t) => t.id === id);
+
+          localUpdatedTodo.splice(index, 1, todo);
+
+          return localUpdatedTodo;
```
BudnikOleksii, line 98: Use `map` instead of `findIndex` and `splice`

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,87 @@
...
+  }, [newTodoTitle]);
+
+  const handleUpdateTodoStatus = () => {
+    let newStatus : boolean;
+
+    if (todos.some(todo => !todo.completed)) {
+      newStatus = true;
+    } else {
+      newStatus = false;
+    }
```
BudnikOleksii, line 53:
```suggestion
    const newStatus = todos.some(todo => !todo.completed);
```
You can simplify

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,169 @@
...
+        ...todo,
+        completed: !todo.completed,
+      });
+    } finally {
+      onProcessing(null);
+    }
+  };
+
+  const updateTodoTitle = () => {
+    if (updateTitle.trim() !== '' && updateTitle !== todo.title) {
```
BudnikOleksii, line 50:
```suggestion
    if (!updateTitle.trim() && updateTitle !== todo.title) {
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,169 @@
...
+        title: updateTitle,
+      });
+
+      onEdit(null);
+      setIsHidenButton(false);
+      setUpdateTitle('');
+    } else if (!updateTitle) {
+      onDelete(todo.id);
+      onEdit(null);
+    } else {
```
BudnikOleksii, line 62: Avoid `else if`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,169 @@
...
+          <div
+            data-cy="TodoLoader"
+            className={cn('modal overlay', {
+              'is-active': isLoading,
+            })}
+          >
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div>
+        )}
```
BudnikOleksii, line 166: You don't need two similar loaders and two states to show loaders

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const TodoList: React.FC<Props> = ({
+  todos,
+  onDelete,
+  todoTemp,
+  isProcessing,
+  onUpdate,
+  isLoading,
+  onProcessing,
+}) => {
+  const [isEdited, setIsEdited] = useState<number | null>(null);
```
BudnikOleksii, line 24: Why do you keep this state here? You use it only in `TodoItem`

src/helpers.tsx
```diff
@@ -0,0 +1,48 @@
...
+    case FilteredBy.All:
+    default:
+      return filteredTodos;
+  }
+}
+
+export const getErrorMessage = (error: ErrorType | null) => {
+  switch (error) {
+    case ErrorType.LoadError:
+      return 'Unable to load todos';
```
BudnikOleksii, line 31: You can store messages directly in the enum so you don't need this function
```typescript
export enum ErrorType {
  LoadError = 'Unable to load todos',
  // other errors
}
```

BudnikOleksii (changes requested):
Good work 👍 
But there are some fixes are required. And some functionality doesn't work properly:
- by default `All` filter should be selected:
image
- When adding todo I see the loader on each todo, not on the temp one
- If I use button toggle all after clear completed there are no loaders on todos
https://www.loom.com/share/d98e57ee80574db8b5a59de88767e795?sid=bdddbaee-8d4b-48b3-a96d-51153d6b34b3

src/App.tsx
```diff
@@ -1,24 +1,185 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
roman-mirzoian, line 2: Try to rewrite the code so that it is not necessary to disable the linter

src/components/Errors/Errors.tsx
```diff
@@ -0,0 +1,46 @@
...
+        clearTimeout(timerRef.current);
+      }
+    };
+  }, [error, onClose]);
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className="notification is-danger is-light has-text-weight-normal"
+
```
roman-mirzoian, line 34:
```suggestion
```

src/components/Header/Header.tsx
roman-mirzoian: Not fixed 😢

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,142 @@
...
+      className={cn('todo', { completed: todo.completed })}
+      key={todo.id}
+    >
+      <label className="todo__status-label">
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+          onChange={() => handleUpdateTodoStatus()}
```
roman-mirzoian, line 90:
```suggestion
          onChange={handleUpdateTodoStatus}
```

src/App.tsx
```diff
@@ -1,24 +1,185 @@
...
+          />
+        )}
+        {todos.length > 0 && (
+          <Footer
+            onFilter={setFilterBy}
+            todos={todosToView}
+            filterBy={filterBy}
+            onClear={clearCompleted}
+          />
+        )}
```
roman-mirzoian, line 174: These rendering conditions can be combined into a single condition.

roman-mirzoian (changes requested):

src/App.tsx
```diff
@@ -1,24 +1,185 @@
...
+    },
+    [setTempTodo, setTodos, setErorType, setNewTodoTitle],
+  );
+
+  const updateTodo = useCallback(
+    async (id: number, todo: Todo) => {
+      try {
+        setProcessingTodoIds((currentIds) => [...currentIds, id]);
+
+        const updatedTodo = await todoService.updateTodo(id, { ...todo });
```
BudnikOleksii, line 91:
```suggestion
        const updatedTodo = await todoService.updateTodo(id, todo);
```

src/App.tsx
```diff
@@ -1,24 +1,185 @@
...
+
+        setTodos((currentTodos) => currentTodos.map(
+          (t) => (t.id === id ? updatedTodo : t),
```
BudnikOleksii, line 94: Avoid one-letter variables and parameters

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const handleUpdateTodoStatus = () => {
+    const newStatus = todos.some(todo => !todo.completed);
+
+    todos.forEach(todo => {
+      onUpdate(todo.id, {
+        ...todo,
+        completed: newStatus,
+      });
+    });
```
BudnikOleksii, line 52: If some of the todos are completed and a new status is completed, you don't need to send requests for them as they already have the same status.

BudnikOleksii (approved):
Nicely done 🔥

PR #997, opened 2023-12-21
Solution

yefimov77:
[DEMO LINK](https://yefimov77.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -25,7 +25,7 @@ export const App: React.FC = () => {
   const [progressStatus, setProgressStatus] = useState<Progress>(Progress.All);
   const [errorMessage, setErrorMessage] = useState<ErrorType | null>(null);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
-  const [globalLoading, setGlobalLoading] = useState(false);
+  const [loadingIds, setLoadingIds] = useState<number[]>([]);
 
   useEffect(() => {
     getTodos(USER_ID)
@@ -116,7 +116,7 @@ export const App: React.FC = () => {
           setErrorMessage={setErrorMessage}
           updateTodos={updateTodos}
           todos={todosFromServer}
-          setGlobalLoading={setGlobalLoading}
+          setLoadingIds={setLoadingIds}
         />
 
         <TodoList
@@ -125,7 +125,7 @@ export const App: React.FC = () => {
           addTodo={addTodo}
           tempTodo={tempTodo}
           updateTodos={updateTodos}
-          globalLoading={globalLoading}
+          loadingIds={loadingIds}
         />
 
         {todosFromServer.length > 0 && (
@@ -135,7 +135,7 @@ export const App: React.FC = () => {
             isAnyCompleted={isAnyCompleted}
             todos={todosFromServer}
             onClear={removeTodo}
-            setGlobalLoading={setGlobalLoading}
+            setLoadingIds={setLoadingIds}
           />
         )}
 
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -8,7 +8,7 @@ interface Props {
   isAnyCompleted: boolean,
   todos: Todo[],
   onClear: (todoId: number) => void,
-  setGlobalLoading: Dispatch<SetStateAction<boolean>>,
+  setLoadingIds:Dispatch<SetStateAction<number[]>>
 }
 
 export const Footer:React.FC<Props> = ({
@@ -17,10 +17,11 @@ export const Footer:React.FC<Props> = ({
   isAnyCompleted,
   todos,
   onClear,
-  setGlobalLoading,
+  setLoadingIds,
 }) => {
   const ClearAllCompleted = async () => {
-    setGlobalLoading(true);
+    setLoadingIds((
+    ) => todos.filter(todo => todo.completed).map(todo => todo.id));
 
     try {
       await Promise.all(
@@ -31,7 +32,7 @@ export const Footer:React.FC<Props> = ({
           }),
       );
     } finally {
-      setGlobalLoading(false);
+      setLoadingIds([]);
     }
   };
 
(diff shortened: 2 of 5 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className="filter__link selected"
+          data-cy="FilterLinkAll"
+          onClick={() => setProgress(Progress.All)}
+        >
+          All
```
BoblykNazar, line 40: You can  replaced 'All', ect.  by enum Progress.

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,127 @@
...
+          className="todo__status"
+          checked={todo.completed}
+          onChange={toggleComplete}
+        />
+      </label>
+
+      {isEditing
+        ? (
+          <form
+            onSubmit={e => {
```
BoblykNazar, line 79: In my opinion, it's better to take this to the separate handler

BoblykNazar (approved):
Nicely done! :bowtie:

src/App.tsx
```diff
@@ -1,24 +1,153 @@
...
+        setErrorMessage(ErrorType.Todo);
+      }
+    },
+    [],
+  );
+
+  const updateTodos = useCallback(
+    async (updatedTodo: Todo) => {
+      try {
+        const updatedTodoFromServer = (await updateTodo(updatedTodo)) as Todo;
```
BudnikOleksii, line 90: To avoid `as` you can path generic on your `updeteTodo` function
```typescript
return client.patch<Todo>(`/todos/${id}`, data);
```

src/App.tsx
```diff
@@ -1,24 +1,153 @@
...
+
+        setTodosFromServer((currentTodos) => {
+          const newTodos = [...currentTodos];
+          const index = newTodos.findIndex(
+            (todo) => todo.id === updatedTodo.id,
+          );
+
+          newTodos.splice(index, 1, updatedTodoFromServer);
+
+          return newTodos;
```
BudnikOleksii, line 100: Use `map` method instead

src/components/Todo/Todo.tsx
BudnikOleksii: > In my opinion, it's better to take this to the separate handler

Good point

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,127 @@
...
+              e.preventDefault();
+
+              if (title.length === 0) {
```
BudnikOleksii, line 82:
```suggestion
              if (!title.length) {
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,41 @@
...
+  tempTodo: TodoType | null,
+  updateTodos: (updatedTodo: TodoType) => void;
+}
+
+export const TodoList: React.FC<Props> = (
+  {
+    todos, onDelete, tempTodo, updateTodos,
+  },
+) => {
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
```
BudnikOleksii, line 18: Why do you need this state here if you use it only in `Todo` component?

src/types/ErrorEnum.ts
```diff
@@ -0,0 +1,7 @@
+export enum ErrorType {
+  Loading = 'Unable to load todos',
+  Title = 'Title should not be empty',
+  Todo = 'Unable to add a todo',
```
BudnikOleksii, line 4: Maybe the name should be `Add` to match the other?

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions. Also, you need to show loaders when toggle all or remove completed too, implement this feature

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,84 @@
...
+export const Footer:React.FC<Props> = ({
+  itemsLeft,
+  setProgress,
+  isAnyCompleted,
+  todos,
+  onClear,
+  setGlobalLoading,
+}) => {
+  const ClearAllCompleted = async () => {
+    setGlobalLoading(true);
```
BudnikOleksii, line 23: You should show loaders only for todos that you plan to update or delete. Because now I see a loader on all todos, when only a few will be deleted

As you need to show multiple loaders at the same time you can store in the state an array of ids of todos that currently loading. So in the todo component, you can check if the current todo id is in the array, and show the loader.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,104 @@
...
+      setErrorMessage(ErrorType.AddTodo);
+    } finally {
+      setIsLoading(false);
+
+      setTitle('');
+    }
+  };
+
+  const toggleAll = async () => {
+    setGlobalLoading(true);
```
BudnikOleksii, line 61: The same

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,122 @@
...
+        onClick={handleDelete}
+      >
+        ×
+      </button>
+
+      <div
+        data-cy="TodoLoader"
+        className={cn(
+          'modal overlay',
+          { 'is-active': isLoading || todo.id === 0 || globalLoading },
```
BudnikOleksii, line 114: And then check here if in the loading ids array is present current todo id. Something like this:
```typescript
const [loadingTodosIds, SetLoadingTodosIds] = useState<number>([]);
```
```suggestion
          { 'is-active': loadingTodosIds.includes(todo.id) },
```

BudnikOleksii (changes requested):
Good progress, just one moment that should be fixed

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,85 @@
...
+    setLoadingIds((
+    ) => todos.filter(todo => todo.completed).map(todo => todo.id));
+
+    try {
+      await Promise.all(
+        todos
+          .filter(todo => todo.completed)
+          .map(async todo => {
+            await onClear(todo.id);
+          }),
```
BudnikOleksii, line 32:
```suggestion
          .map(todo => onClear(todo.id)),
```
You await requests with `Promise.all`, so you can just create an array of promises and wait for the result like you do in `toggleAll` function

BudnikOleksii (approved):
Nicely done 🚀

PR #1001, opened 2023-12-23
Solution

HannieA:
https://hanniea.github.io/react_todo-app-with-api/

### DIFF
--- a/src/components/Notification/Notification.tsx
+++ b/src/components/Notification/Notification.tsx
@@ -1,0 +1,33 @@
+import React, { useContext, useState } from 'react';
+import classNames from 'classnames';
+
+import { TodoContext } from '../TodoContext';
+
+export const Notification: React.FC = () => {
+  const { errorMessage } = useContext(TodoContext);
+  const [isHidden, setIsHidden] = useState(false);
+
+  setTimeout(() => {
+    setIsHidden(true);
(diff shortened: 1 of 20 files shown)


src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,33 @@
...
+    setIsHidden(true);
+  }, 3000);
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={classNames('notification '
+            + 'is-danger '
+            + 'is-light '
+            + 'has-text-weight-normal',
```
sTorba24, line 20:
```suggestion
      className={classNames('notification is-danger is-light has-text-weight-normal'
       
```

sTorba24: also you can add classes just splitting by comma https://www.npmjs.com/package/classnames
classNames('a', 'b', { hidden: isHidden }, 'any-class' )

sTorba24 (changes requested):
Nicely done! Lets improve it

1. loader appears only on last element when i click on `change status all`
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

sTorba24 (approved):

PR #1002, opened 2023-12-23
first try

Mantserov:
[DEMO LINK](https://Mantserov.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,226 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useEffect, useMemo, useRef, useState,
+} from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
+import * as todosService from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+import { TodoFooter } from './components/TodoFooter/TodoFooter';
+import { Status } from './types/Status';
+import { ErrorNotification } from './components/ErrorNotification';
 
-const USER_ID = 0;
+const USER_ID = 12042;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [status, setStatus] = useState<Status>(Status.all);
+  const [todoTitle, setTodoTitle] = useState<string>('');
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const titleField = useRef<HTMLInputElement>(null);
+
+  const handleError = (errorMessage: string) => {
+    setError(errorMessage);
+    setTimeout(() => {
+      setError(null);
+    }, 3000);
+  };
+
+  useEffect(() => {
+    if (titleField.current) {
+      titleField.current.focus();
+    }
+
+    todosService.getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => (handleError('Unable to load todos')));
+  }, []);
+
+  const filteredTodos = useMemo(() => {
+    switch (status) {
+      case Status.all: {
+        return todos;
+      }
+
+      case Status.active: {
+        return todos.filter(todo => !todo.completed);
+      }
+
+      case Status.completed: {
+        return todos.filter(todo => todo.completed);
+      }
+
+      default: {
+        return todos;
+      }
+    }
+  }, [todos, status]);
+
+  const createTodo = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (!todoTitle) {
+      handleError('Title should not be empty');
+    }
+
+    if (todoTitle) {
+      const tempTodoId = 0;
+
+      setTempTodo({
+        id: tempTodoId,
+        userId: USER_ID,
+        title: todoTitle,
+        completed: false,
+      });
+
+      setLoadingTodoIds(curr => [...curr, tempTodoId]);
+
+      todosService.addTodo({
+        userId: USER_ID, title: todoTitle, completed: false,
+      })
+        .then(newTodo => {
+          setTodos(currentTodos => {
+            const maxId = Math.max(0, ...currentTodos.map(todo => todo.id));
+            const id = maxId + 1;
+
+            return [...currentTodos, { ...newTodo, id }];
+          });
+          setTodoTitle('');
+        })
+        .catch(() => handleError('Unable to add a todo'))
+        .finally(() => {
+          setTempTodo(null);
+          setLoadingTodoIds(curr => curr.filter(id => id !== tempTodoId));
+        });
+    }
+  };
+
+  const deleteTodo = (todoId: number) => {
+    setLoadingTodoIds(curr => [...curr, todoId]);
+
+    todosService.deleteTodo(todoId)
+      .then(() => setTodos(
+        currentTodos => currentTodos.filter(todo => todo.id !== todoId),
+      ))
+      .catch(() => handleError('Unable to delete a todo'))
+      .finally(() => {
+        setLoadingTodoIds(curr => curr.filter(id => id !== todoId));
+      });
+  };
+
+  const clearTodos = () => {
+    todos.map(todo => {
+      if (todo.completed) {
+        todosService.deleteTodo(todo.id);
+      }
+
+      return 1;
+    });
+
+    setTodos(
+      currentTodos => currentTodos.filter(todo => todo.completed !== true),
+    );
+  };
+
+  const updateTodo = (newTodo: Todo): Promise<void> => {
+    setLoadingTodoIds(curr => [...curr, newTodo.id]);
+
+    return todosService.updateTodo(newTodo)
+      .then((todoFromServer) => setTodos(
+        currentTodos => currentTodos.map((todo) => {
+          const updatedTodo = todoFromServer as Todo;
+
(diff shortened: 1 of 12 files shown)


src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,68 @@
...
+          Completed
+        </a>
+      </nav>
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        onClick={() => onClear()}
+        disabled={visibleTodos.every(todo => todo.completed === false)}
```
ruslanagnatenko, line 62: `todo => !todo.completed`

src/App.tsx
```diff
@@ -1,24 +1,216 @@
...
+    const todoUpdated = { ...todo, completed: !todo.completed };
+
+    updateTodo(todoUpdated);
+  };
+
+  const toggleAllTodos = () => {
+    todos.map(todo => {
+      updateTodoComplete(todo);
+
+      return 1;
```
ruslanagnatenko, line 146: no need to return 1

src/App.tsx
```diff
@@ -1,24 +1,216 @@
...
+      });
+  };
+
+  const updateTodoComplete = (todo: Todo) => {
+    const todoUpdated = { ...todo, completed: !todo.completed };
+
+    updateTodo(todoUpdated);
+  };
+
+  const toggleAllTodos = () => {
```
ruslanagnatenko, line 142: your function just change the completed property of todo, so every completed todo becomes uncompleted and every uncompleted becomes completed (check the demo), fix this, checking if there's some uncompleted todo in todos array, if it is, then make all uncompleted todos completed, in other case make every todo uncompleted
image 14

src/App.tsx
```diff
@@ -1,24 +1,216 @@
...
+      .then(() => setTodos(
+        currentTodos => currentTodos.filter(todo => todo.id !== todoId),
+      ))
+      .catch(() => handleError('Unable to delete a todo'))
+      .finally(() => {
+        setLoadingTodoIds(curr => curr.filter(id => id !== todoId));
+      });
+  };
+
+  const clearTodos = () => {
```
ruslanagnatenko, line 100: you forgot to add completed todos ids to state of loadingTodoIds while clearing

src/App.tsx
```diff
@@ -1,24 +1,216 @@
...
+      });
+
+      setLoadingTodoIds(curr => [...curr, tempTodoId]);
+
+      todosService.addTodo({
+        userId: USER_ID, title: trimmedTitle, completed: false,
+      })
+        .then(newTodo => {
+          setTodos(currentTodos => {
+            return [...currentTodos, { ...newTodo }];
```
ruslanagnatenko, line 75: no need to destructure newTodo, just `[...currentTodos, newTodo]`

ruslanagnatenko (changes requested):
Good, but need to add some changes to be done

src/App.tsx
```diff
@@ -1,24 +1,222 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [status, setStatus] = useState<Status>(Status.all);
+  const [todoTitle, setTodoTitle] = useState<string>('');
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const titleField = useRef<HTMLInputElement>(null);
```
roman-mirzoian, line 24:
```suggestion
  const titleRef = useRef<HTMLInputElement>(null);
```

src/App.tsx
```diff
@@ -1,24 +1,222 @@
...
+          if (todo.id === updatedTodo.id) {
+            return updatedTodo;
+          }
+
+          return todo;
+        }),
+      ))
+      .catch(() => {
+        handleError('Unable to update a todo');
+        throw new Error();
```
roman-mirzoian, line 127: There is no point in throwing an empty error, because it will not carry any information

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,68 @@
...
+  onClear: () => void
+}
+
+export const TodoFooter: React.FC<Props> = ({
+  allTodos, visibleTodos, status, onStatus, onClear,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${allTodos.filter(todo => !todo.completed).length} items left`}
```
roman-mirzoian, line 19: Such a record is very unreadable, put the calculation of the number in a separate variable

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,142 @@
...
+}) => {
+  const [isEditing, setIsEditing] = useState(false);
+  const [editedTodoTitle, setEditedTodoTitle] = useState('');
+
+  const handleEdit = () => {
+    setIsEditing(true);
+    setEditedTodoTitle(todo.title);
+  };
+
+  const EditedTitleField = useRef<HTMLInputElement>(null);
```
roman-mirzoian, line 28:
```suggestion
  const editedTitleRef = useRef<HTMLInputElement>(null);
```

src/helpers.tsx
```diff
@@ -0,0 +1,24 @@
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
+
+export const filteredTodos = (todos: Todo[], status: Status) => {
+  const preparedTodos = [...todos];
+
+  switch (status) {
+    case Status.all: {
+      return preparedTodos;
+    }
```
roman-mirzoian, line 10: This case returns the same as the `default` block, so you should either combine the cases or remove this case

src/types/Status.ts
```diff
@@ -0,0 +1,5 @@
+export enum Status {
+  all = 'all',
+  active = 'active',
+  completed = 'completed',
+}
```
roman-mirzoian, line 5: It is customary to write the names of enum values either in all capital letters or with a capital letter
```suggestion
export enum Status {
  All = 'all',
  Active = 'active',
  Completed = 'completed',
}
```

roman-mirzoian (changes requested):
Also, the Demo is currently not working correctly:
https://gyazo.com/154b342a5da1158ac4ad445fccf00c05

(Optional) Not all GitHub tests pass validation.

src/App.tsx
Mantserov: The thing is that without this throwing my handleEditSumbit function doesn't work correctly when it comes to showing an error message. To be more precise — the catch of this function doesn't work correctly when I'm trying to edit Todo in the offline mode.

Mantserov:
> (Optional) Not all GitHub tests pass validation.

I think there is something wrong with this tests to be honest :D

src/App.tsx
BudnikOleksii: > no need to return 1

Agree, doesn't make sense to return something if you don't use a new array. You can use forEach instead

src/App.tsx
```diff
@@ -1,24 +1,222 @@
...
+
+  const updateTodoComplete = (todo: Todo) => {
+    const todoUpdated = { ...todo, completed: !todo.completed };
+
+    updateTodo(todoUpdated);
+  };
+
+  const toggleAllTodos = () => {
+    if (todos.every(todo => todo.completed)) {
+      return todos.map(todo => updateTodoComplete(todo));
```
BudnikOleksii, line 142: The same, you don't need returned array

src/App.tsx
```diff
@@ -1,24 +1,222 @@
...
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">React Todo App - Add and Delete</a>
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
```
BudnikOleksii, line 165: Create a header component

src/helpers.tsx
```diff
@@ -0,0 +1,21 @@
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
+
+export const filteredTodos = (todos: Todo[], status: Status) => {
+  const preparedTodos = [...todos];
```
BudnikOleksii, line 5: Filter method creates a new array, you don't need to do a copy

BudnikOleksii (changes requested):
Good work 👍 
Just a few minor fixes are required

src/App.tsx
BudnikOleksii: > no need to return 1

Still not fixed 🤷‍♂️  When you don't need a new array, but just iterate through it you can use forEach or forof
And it's not only here, but in every similar case

src/api/todos.ts
```diff
@@ -0,0 +1,22 @@
...
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+export const addTodo = ({
+  userId, title, completed,
+}: Omit<Todo, 'id'>) => {
+  return client.post<Todo>('/todos', {
+    userId, title, completed,
+  });
+};
```
BudnikOleksii, line 14:
```suggestion
export const addTodo = (newTodo: Omit<Todo, 'id'>) => {
  return client.post<Todo>('/todos', newTodo);
};
```

BudnikOleksii (approved):
Approved, but, please, pay attention to what array iteration method to use in each case.
It doesn't make sense to use `map` when you need just iterate through it

PR #1003, opened 2023-12-23
react_todo-app-with-api

vadymboichenkoAcupower:
[DEMO LINK](https://vadymboichenkoAcupower.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,266 @@
+/* eslint-disable jsx-a11y/no-autofocus */
+/* eslint-disable @typescript-eslint/no-unused-expressions */
+/* eslint-disable react-hooks/rules-of-hooks */
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useEffect, useMemo, useRef, useState,
+} from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import {
+  addTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+} from './api/todos';
+import { TodoList } from './components/TodoList/TodoList';
+import { Status } from './types/Status';
+import { Errors } from './types/Errors';
+import { TodosFilter } from './components/TodoFilter/TodoFilter';
 
-const USER_ID = 0;
+const USER_ID = 7023;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredType, setFilteredType] = useState(Status.All);
+  const [, setTodo] = useState<Todo | null>(null);
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<Errors | null>(null);
+  const [title, setTitle] = useState<string>('');
+
+  const titleRef = useRef<HTMLInputElement | null>(null);
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
+
+  const handleToggleAll = async () => {
+    const allCompleted = todos.every(todo => todo.completed === true);
+
+    if (allCompleted) {
+      const updatedTodos = todos.map(todo => ({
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,24 +1,273 @@
+/* eslint-disable jsx-a11y/no-autofocus */
+/* eslint-disable @typescript-eslint/no-unused-expressions */
+/* eslint-disable react-hooks/rules-of-hooks */
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
Andriy-Shymkiv, line 5: it is a lot of disabled eslint rules. not a good approach

src/App.tsx
```diff
@@ -1,24 +1,273 @@
...
+import { Errors } from './types/Errors';
+import { TodosFilter } from './components/TodoFilter/TodoFilter';
 
-const USER_ID = 0;
+const USER_ID = 7023;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredType, setFilteredType] = useState(Status.All);
+  const [, setTodo] = useState<Todo | null>(null);
```
Andriy-Shymkiv, line 28: why did you need this? what's the purpose?

src/App.tsx
```diff
@@ -1,24 +1,273 @@
...
+    setTitle(event.target.value);
+  };
+
+  const handleUpdateTodo = (todo: Todo) => {
+    updateTodo(todo)
+      .then(() => setTodos(current => current.map(t => (t.id === todo.id ? todo : t))))
+      .catch(() => setErrorMessage(Errors.CAN_NOT_UPDATE_TODO));
+  };
+
+  const handleToggleAll = async () => {
```
Andriy-Shymkiv, line 45: no need to use `async` if you don't use `await` inside
```suggestion
  const handleToggleAll = () => {
```

src/App.tsx
```diff
@@ -1,24 +1,273 @@
...
+      .catch(() => {
+        setTitle(title);
+        setErrorMessage(Errors.CAN_NOT_ADD_TODO);
+      })
+      .finally(() => {
+        setTodo(null);
+        setIsInputDisabled(false);
+      });
+
+    setTitle('');
```
Andriy-Shymkiv, line 100: clear title only in success add

Andriy-Shymkiv (changes requested):

src/App.tsx
vadymboichenkoAcupower: this need to save todo while it not saved to the server server. This task was in previous topic

src/App.tsx
anastasiia-tilikina: not fixed

src/App.tsx
```diff
@@ -1,24 +1,274 @@
-/* eslint-disable max-len */
+/* eslint-disable react-hooks/rules-of-hooks */
```
anastasiia-tilikina, line 1: this shouldn't be disabled as well

src/App.tsx
```diff
@@ -1,24 +1,274 @@
...
+import { TodosFilter } from './components/TodoFilter/TodoFilter';
 
-const USER_ID = 0;
+const USER_ID = 7023;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredType, setFilteredType] = useState(Status.All);
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
+  const [, setTodo] = useState<Todo | null>(null);
```
anastasiia-tilikina, line 26: sorry but it's still not clear why do you need this state. what does it store?

anastasiia-tilikina (changes requested):
1. move header, footer and error message to separate components to keep your code clean and readable.
2. implement loader state for totdos while waiting response from server.
3. implement temp todo to show it while new todo is creating (see working version)
4. your toggle all button is always active now
5. it should be 1 item left title in the footer when the is only one todo (remove plural for this case)
6. There is a glitch now with editing todo.  when I press enter it shows an old title at first and a few moments later it shows a new one. pls fix it.

sTorba24 (approved):
GJ 🚀

PR #1008, opened 2023-12-26
solution

emilionila:


emilionila:
[DEMO LINK](https://emilionila.github.io/react_todo-app-with-api/)

src/App.tsx
```diff
@@ -1,24 +1,165 @@
...
+      });
+    } catch (e) {
+      handleError(Errors.UpdateTodo);
+    }
+  };
+
+  const deleteTodo = (todoId: number) => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    setLoadingTodoId(completedTodos.map(todo => todo.id));
```
BudnikOleksii, line 89:
```suggestion
  const deleteTodo = (todoId: number) => {
    setLoadingTodoId(prevIds => [...prevIds, todoId]);
```
Strange logic and behavior for now, when I'm deleting not completed todo I see the loader on completed instead of what I actually want to delete, and the loader doesn't disappear after deleting

src/App.tsx
```diff
@@ -1,24 +1,165 @@
...
+    todosService.deleteTodo(todoId)
+      .then(() => setTodos(
+        currentTodos => currentTodos.filter(todo => todo.id !== todoId),
+      ))
+      .catch(() => handleError(Errors.DeleteTodo));
```
BudnikOleksii, line 94:
```suggestion
      .catch(() => handleError(Errors.DeleteTodo))
      .finally(() => {
        setLoadingTodoId(prevIds => prevIds.filter(id => id !== todoId);
      });
```
To remove the loader add finally block and remove id from the array of loading todo ids

src/App.tsx
```diff
@@ -1,24 +1,165 @@
...
+
+  const handleEditTodo = async (todoId: number, newTitle: string) => {
+    if (!newTitle) {
+      setErrorMessage(Errors.UpdateTodo);
+
+      return;
+    }
+
+    try {
+      setLoadingTodoId([todoId]);
```
BudnikOleksii, line 105: What if currently loading other todos? Use previous state as I mentioned previously

src/App.tsx
```diff
@@ -1,24 +1,165 @@
...
+      const updatedTodo = await todosService.updateTodoTitle(todoId, newTitle);
+
+      setTodos(currentTodos => currentTodos.map(
+        todo => (todoId === todo.id ? updatedTodo : todo),
+      ));
+    } catch {
+      setErrorMessage(Errors.EmptyTitle);
+    } finally {
+      setLoadingTodoId([]);
```
BudnikOleksii, line 114: The same, don't remove loaders for all todos, only for current

src/App.tsx
```diff
@@ -1,24 +1,165 @@
...
+      ))
+      .catch(() => handleError(Errors.DeleteTodo));
+  };
+
+  const handleEditTodo = async (todoId: number, newTitle: string) => {
+    if (!newTitle) {
+      setErrorMessage(Errors.UpdateTodo);
+
+      return;
+    }
```
BudnikOleksii, line 102: You should delete todo instead if newTitle is empty
image

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,107 @@
...
+    if (!inputValue.trim()) {
+      handleError(Errors.EmptyTitle);
+
+      return;
+    }
+
+    setIsDisabledInput(true);
+
+    try {
+      addTodo(inputValue);
```
BudnikOleksii, line 45:
```suggestion
      await addTodo(inputValue);
```
You should wait for the response from the API before clear input

Tip: to have ability to wait for the response from the `addTodo` function it should return the promise.
```typescript
const addTodo = (inputValue: string) => {
    // your code

    return todosService.createTodo(data) // return your api call
      .then().catch().finally() // your logic to handle response from create todo API call
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,107 @@
...
+      setIsDisabledInput(false);
+    }
+  };
+
+  const completeAllTodos = async () => {
+    const statusOfTodos = todos.every(todo => todo.completed);
+
+    const todoIdsToUpdate = todos.map(todo => todo.id);
+
+    setLoadingTodoId(todoIdsToUpdate);
```
BudnikOleksii, line 58: You set loaders in the `updateTodo` function

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,107 @@
...
+      ));
+
+      await Promise.all(updateTodos.map(async todo => {
+        await updateTodo({ ...todo, completed: !todo.completed });
+      }));
+    } catch {
+      setErrorMessage(Errors.UpdateTodo);
+    } finally {
+      setLoadingTodoId([]);
+    }
```
BudnikOleksii, line 72: The same, you can remove loaders in `the `updateTodo` function

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,107 @@
...
+      const updateTodos = todos.filter(todo => (
+        !statusOfTodos ? !todo.completed : todo.completed
+      ));
+
+      await Promise.all(updateTodos.map(async todo => {
+        await updateTodo({ ...todo, completed: !todo.completed });
+      }));
+    } catch {
+      setErrorMessage(Errors.UpdateTodo);
+    } finally {
```
BudnikOleksii, line 70: And an error you set in the `updateTodo` as well

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,132 @@
...
+    handleEditTodo,
+  } = props;
+
+  const [isEditing, setIsEditing] = useState(false);
+  const [editedTitle, setEditedTitle] = useState(todo.title);
+
+  const titleField = useRef<HTMLInputElement>(null);
+
+  const handleToggle = async () => {
+    setLoadingTodoId?.([todo.id]);
```
BudnikOleksii, line 34: The same, you set and remove loader in `updateTodo`, fix it

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,132 @@
...
+    setIsEditing(false);
+  };
+
+  const handleKeyUp = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter') {
+      handleEditTitle();
+    } else if (event.key === 'Escape') {
+      setEditedTitle(todo.title);
+      setIsEditing(false);
+    }
```
BudnikOleksii, line 66:
```suggestion
    }
    
     if (event.key === 'Escape') {
      setEditedTitle(todo.title);
      setIsEditing(false);
    }
```
Avoid `else if`

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,132 @@
+import React, {
+  FC, useEffect, useRef, useState,
+} from 'react';
+import cn from 'classnames';
+import { Todo as TodoType } from '../../types/Todo';
+
+interface Props {
+  todo: TodoType,
+  deleteTodo: (todoId: number) => void,
+  loading: boolean,
```
BudnikOleksii, line 10:
```suggestion
  isLoading: boolean,
```
Boolean variables should have prefix `is` or `has`

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      <TransitionGroup>
+        {todos.map(todo => (
+          <CSSTransition
+            key={todo.id}
+            timeout={300}
+            classNames="item"
+          >
```
BudnikOleksii, line 35: 🔥

BudnikOleksii (changes requested):
Good start 👍 
Some functionality doesn't work as expected, I mentioned in comments what's wrong and left suggestions to fix it.
Make sure that your app works as required

src/components/Todo/Todo.tsx
BudnikOleksii: > The same, you set and remove loader in `updateTodo`, fix it

Not fixed

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const titleChanger = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setEditedTitle(event.target.value);
+  };
+
+  const handleEditTitle = () => {
+    setIsEditing(true);
+
+    if (editedTitle === todo.title) {
+      setIsEditing(false);
+    }
```
BudnikOleksii, line 54:
```suggestion
    if (editedTitle === todo.title) {
      setIsEditing(false);
      
      return;
    }
```
Don't forget to return, as in your case I still see a request and then an error

BudnikOleksii (changes requested):
Great progress 👍 
Almost done, just a few moments should be fixed

roman-mirzoian (approved):
Well done 👍

PR #1011, opened 2023-12-27
Develop

Fluebubble:
[DEMO LINK](https://Fluebubble.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Todo/Todo.tsx
+++ b/src/components/Todo/Todo.tsx
@@ -58,6 +58,9 @@ export const Todo: React.FC<Props> = ({
       } catch (err) {
         handleError(Errors.UnableToDelete);
       } finally {
+        setSelectedTodos(prevTodosIds => prevTodosIds
+          .filter(prevId => prevId !== editableTodo.id));
+
         setEditableTodo(null);
       }
 


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { filterTodos } from '../../utils/helpers';
+
+interface Props {
+  filterType: FilterType,
+  onFilterSelect: (filterT: FilterType) => void,
+  todos: Todo[],
+  clearCompleted: () => void
+}
+
+export const Footer: React.FC<Props> = ({
```
roman-mirzoian, line 14: The `Filter` component copies the functionality of the `Footer` component, what's the point?

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,61 @@
...
+    if (titleField.current) {
+      titleField.current.focus();
+    }
+  }, [todos]);
+
+  return (
+    <header className="todoapp__header">
+      <button
+        type="button"
+        // className="todoapp__toggle-all active"
```
roman-mirzoian, line 35: Do not leave unnecessary comments.

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,78 @@
...
+            handleDelete={() => { }}
+            selectedTodos={selectedTodos}
+            onClick={() => { }}
+            editTodoTitle={() => { }}
+            editableTodo={editableTodo}
+            setEditableTodo={() => { }}
+            handleError={handleError}
+            setTodos={setTodos}
+            setSelectedTodos={() => { }}
+          />
```
roman-mirzoian, line 74: Set the default fields to be equal to the empty function, then you won't need to pass extra empty props

roman-mirzoian (changes requested):
Also, the application itself has a somewhat incorrect behavior:
- if I add spaces between words - they will be cut off when saving:
https://gyazo.com/a01ea40fb30b5ff0a700470d5f6b641b
similar behavior occurs when creating a new record

- this functionality is not implemented
image

src/components/TodoList/TodoList.tsx
Fluebubble: Solved

PR #1015, opened 2023-12-28
Solution

andreysakhno:


src/components/TodoApp/RemoveCompleted/RemoveCompleted.tsx
```diff
@@ -0,0 +1,67 @@
...
+      deleteTodo(id)
+        .then(() => id)
+    ));
+
+    Promise.allSettled(deletePromises)
+      .then(results => {
+        let hasError = false;
+
+        results.forEach(result => {
+          if (result.status === 'fulfilled') {
```
lerastarynets, line 35: consider moving statuses value to an enum and using it here

src/libs/state/reducer.ts
lerastarynets, line 1: better to give a name to your reducer eg todoReducer, commonReducer etc.

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #1019, opened 2023-12-30
first try

SerhiiUnhurian:
[DEMO LINK](https://SerhiiUnhurian.github.io/react_todo-app-with-api/)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${countActiveTodos} item${countActiveTodos > 1 ? 's' : ''} left`}
+      </span>
+
+      <Filter />
+      {/* don't show this button if there are no completed todos */}
+      {countCompletedTodos && (
```
TarasHoliuk, line 32: now when the Clear button doesn't exist filter is not centered - you may use `visibility: hidden` to hide the button but keep the place for it (and disable it if tests fail):

image

src/components/TododsContext/TodosContext.tsx
```diff
@@ -0,0 +1,67 @@
...
+      .then(setTodos)
+      .catch(() => setErrorMessage('Unable to load todos'));
+  }, []);
+
+  const value = useMemo(() => ({
+    todos,
+    setTodos,
+    filterOption,
+    setFilterOption,
+    USER_ID,
```
TarasHoliuk, line 55:
```suggestion
    userId: USER_ID,
```
or just create a separate `constants.ts` file with this exported variable and import it wherever you need

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,110 @@
...
+        })
+        .finally(() => {
+          setTimeout(() => setErrorMessage(''), 3000);
+          setTempTodo(null);
+          setIsSubmiting(false);
+        });
+    }
+  };
+
+  const handleChangeToggle = () => {
```
TarasHoliuk, line 69: it changes only the local state, but doesn't do requests on the server:
[Screencast from 31-12-23 14:08:08.webm](https://github.com/mate-academy/react_todo-app-with-api/assets/86923089/2c1090f7-7e26-4fe6-9ed9-26c0ed4ab193)

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,110 @@
...
+        completed: false,
+        userId: USER_ID,
+      })
+        .then((newTodo: Todo) => {
+          setTodos(currentTodos => [...currentTodos, newTodo]);
+          setTodoTitle('');
+        })
+        .catch((error) => {
+          setErrorMessage('Unable to add a todo');
+          throw error;
```
TarasHoliuk, line 59: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,110 @@
+import classNames from 'classnames';
```
TarasHoliuk, line 1: No need to create a separate folder for the component if there is only 1 file inside

A good example is `TodoItem.tsx` - there is no folder

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const handleDeleteTodo = (todoId: number) => {
+    setIsDeliting(true);
+
+    deleteTodo(todoId)
+      .then(() => setTodos(
+        curentTodos => curentTodos.filter(curtodo => curtodo.id !== todoId),
+      ))
+      .catch((error) => {
+        setErrorMessage('Unable to delete a todo');
+        throw error;
```
TarasHoliuk, line 44: again - do not throw it

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,169 @@
...
+              className="todo__remove"
+              data-cy="TodoDelete"
+              onClick={() => handleDeleteTodo(todo.id)}
+            >
+              ×
+            </button>
+          </>
+        )}
+
+      {(todo.id === 0 || isDeliting || toggled || selectedTodo) && (
```
TarasHoliuk, line 164: currently, you don't show the loader for batch actions (toggle all, clear completed), and use different variables to show the loader. You may create a 1 separate on a higher level with `loadingTodoIds` array (empty by default). Once you do any action for todo add its id to the state mentioned above. When the action is done - remove the id. Inside this component check if in those array exists id of current todo - if yes - show loader (and gray background)

TarasHoliuk (changes requested):
Hi! Very good job! :+1: 

Batch toggle doesn't work, also there are no loader for batch action, + a few minor comments

DarkMistyRoom (approved):

PR #1020, opened 2024-01-01
Solution

Barosz30:
https://barosz30.github.io/react_todo-app-with-api/


mosowiecka (changes requested):
Toggle button should update existing todos instead it looks like it is adding them in a different order. 
An error occurs when setting an empty title.

mbruhler (approved):

PR #1021, opened 2024-01-01
Develop

mosowiecka:
[DEMO LINK](https://mosowiecka.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/NewTodo/NewTodo.tsx
+++ b/src/components/NewTodo/NewTodo.tsx
@@ -1,0 +1,65 @@
+import { FC, useState } from 'react';
+import { Todo } from '../../types/Todo';
+import { Errors } from '../../types/ErrorTypes';
+
+type Props = {
+  onAdd: (todo: Todo) => void,
+  userId: number,
+  setErrorMsg: (errorMsg: Errors | null) => void,
+};
+
+export const NewTodo: FC<Props> = (
+  {
+    onAdd, userId, setErrorMsg,
+  },
+) => {
+  const [inputTitle, setInputTitle] = useState('');
+  const [isDisabled, setIsDisabled] = useState(false);
+
+  const addTodo = (title: string) => {
+    const todo: Todo = {
+      id: 0,
+      userId,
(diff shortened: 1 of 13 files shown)


src/types/ErrorTypes.ts
```diff
@@ -0,0 +1,9 @@
+export type ErrorType = 'load' | 'title' | 'add' | 'delete' | 'update';
+
+export enum Errors {
+  load = 'Unable to load todos',
+  title = 'Title should not be empty',
+  add = 'Unable to add a todo',
+  delete = 'Unable to delete a todo',
+  update = 'Unable to update a todo',
```
danon321, line 8: Names in enum should be in upercase. 
There is no need for ErrorType when u have enum Errors, u can take it from there

danon321 (changes requested):
Enums need u

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,65 @@
...
+  userId: number,
+  setErrorMsg: (errorMsg: Errors | null) => void,
+};
+
+export const NewTodo: FC<Props> = (
+  {
+    onAdd, userId, setErrorMsg,
+  },
+) => {
+  const [inputTitle, setInputTitle] = useState('');
```
danon321, line 16: Try to do it by useRef. Now component is rerendering everytime user type singe letter and we dont need that
https://react.dev/reference/react/useRef

danon321 (changes requested):

danon321:
You should have autofocus in main input on page refresh

danon321:
'Clear complited' works in the opposite way than it should. If i dont have any 'compilted' its shown and if i doesnt its disappear

danon321 (changes requested):
'Clear complited' works in the opposite way than it should. If i dont have any 'compilted' its shown and if i doesnt its disappear. Sorry i didnt see it before

danon321 (approved):

PR #1023, opened 2024-01-03
Solution

MarcinLigmann:


src/components/todolist/todolist.tsx
```diff
@@ -15,18 +15,27 @@ export const TodoList = () => {
   const handleDeleteClick = (id: number) => {
     setError(null);
 
-    setDeletingTask([...deletingTask, id]);
+    // setDeletingTask([...deletingTask, id]);
+    // setDeletingTask((prevState: number[]) => [...prevState, id]);
```
zaura333, line 19: Unnecessary comments can be deleted

src/components/todoform/todoform.tsx
```diff
@@ -0,0 +1,137 @@
...
+        setTaskName('');
+      })
+      .catch(() => setError(ErrorType.add))
+      .finally(() => {
+        setIsAddingTask(false);
+        setTempTodo(null);
+      });
+  };
+
+  const completeAllTasks = () => {
```
zaura333, line 60: Probably could be renamed to something like toggle, cause it's not only completing tasks

src/components/todoItem/todoItem.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+        copy[index] = data;
+
+        setTodos(copy);
+      })
+      .catch(() => setError(ErrorType.update))
+      .finally(() => {
+        setIsAddingTask(false);
+        setIsEdited(null);
+      });
```
zaura333, line 65: I think I'd move this logic to some method (e.g. updateTodo) in the context provider since it's repeated later onBlur, however works like that as well

zaura333 (approved):
Looks good, just a few minor suggestions to consider!

danon321 (approved):

PR #1024, opened 2024-01-03
Develop

YuriyBilskiy:
[DEMO LINK](https://yuriybilskiy.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoHeader.tsx
+++ b/src/components/TodoHeader.tsx
@@ -1,22 +1,83 @@
-import React, { RefObject } from 'react';
+import React, { RefObject, useContext, useState } from 'react';
+import * as postServise from '../api/todos';
+import { TodosContext } from './TodoProvider';
+import { Errors } from '../types/Errors';
 
 type Props = {
-  handleFormSubmit: (event: React.FormEvent<HTMLFormElement>) => void;
   handleChangeInput: (event: React.ChangeEvent<HTMLInputElement>) => void;
   title: string,
-  isSubmiting: boolean,
   inputRef: RefObject<HTMLInputElement>;
-  toggleAll: () => void;
+  setTitle: (valu: string) => void,
 };
 
 export const TodoHeader: React.FC<Props> = ({
-  handleFormSubmit,
   handleChangeInput,
   title,
-  isSubmiting,
   inputRef,
-  toggleAll,
+  setTitle,
 }) => {
+  const [isSubmiting, setIsSubmiting] = useState(false);
+  const {
+    todos,
+    setTodos,
+    setError,
+    USER_ID,
+  } = useContext(TodosContext);
+
+  const handleFormSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!title.trim()) {
+      setError(Errors.ERRORS_EMPTY_TITLE);
+
+      return;
+    }
+
+    setIsSubmiting(true);
+
+    try {
+      const newTodo = await postServise.createTodo({
+        completed: false,
+        title: title.trim(),
+        userId: USER_ID,
+      });
+
+      setTodos(currentTodo => [...currentTodo, newTodo]);
+      setTitle('');
+    } catch (newError) {
+      setError(Errors.UNABLE_ADD);
+      throw newError;
+    } finally {
+      if (inputRef.current) {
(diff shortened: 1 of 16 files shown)


TarasHoliuk:
1. It should show the number of uncompleted todos:
image

2. If there are no completed todos the `Clear completed` button should not be visible:
image

But note that it should keep its own place (in other words filter buttons should always be in the same place - centered).

3. Focus input when double click on todo for editing:

Current result:
image
 
Expected: 
image

4. You should always hide input on blur:
[Screencast from 03-01-24 20:57:54.webm](https://github.com/mate-academy/react_todo-app-with-api/assets/86923089/3e6f9218-6348-400d-b320-8428b880ecb2)

Also, if you've removed all symbols when editing you should delete the todo:
[Screencast from 03-01-24 21:00:04.webm](https://github.com/mate-academy/react_todo-app-with-api/assets/86923089/f6a5555d-2b7c-4de1-8dc5-5a194b6c1eb4)
Check how [the working example](https://mate-academy.github.io/react_todo-app-with-api/) works

5. You should show loader on todo every time you change/add/delete it (including multi-actions like clear completed or toggle all). Ask in the chat if you need help with it.

src/components/TodoList.tsx
```diff
@@ -0,0 +1,152 @@
...
+      return todo;
+    });
+
+    setTodos(completedTodo);
+  };
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {filteredTodo.map((todo) => (
+        <div
```
TarasHoliuk, line 99: I'd recommend you to move `TodoItem` to a separate component

it will simplify a lot of methods - they should accept entire todo object as an argument if needed

src/components/TodoList.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  const handleBlur = () => {
+    setIsEditing(false);
+  };
+
+  const handleKeyPress = async (
+    event: React.KeyboardEvent<HTMLInputElement>,
+    editingId: number,
+  ) => {
+    if (event.key === 'Enter') {
```
TarasHoliuk, line 51: you should also handle `escape` button

src/components/TodoList.tsx
```diff
@@ -0,0 +1,152 @@
...
+          >
+            ×
+          </button>
+
+          <div data-cy="TodoLoader" className="modal overlay">
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div>
+        </div>
+      ))}
```
TarasHoliuk, line 149: you should also show `tempTodo` here with loader when adding new todo (there is more detailed instruction in the previous task)

src/components/TodoProvider.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const value = {
+    todos,
+    setTodos,
+    error,
+    setError,
+    removeTodo,
+    setFilterType,
+    filteredTodo,
+    filterType,
+    USER_ID,
```
TarasHoliuk, line 64:
```suggestion
    userId: USER_ID,
```

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,105 @@
...
+    setIsSubmiting(true);
+
+    try {
+      const newTodo = await postServise.createTodo({
+        completed: false,
+        title: title.trim(),
+        userId: USER_ID,
+      });
+
+      setTodos(currentTodo => [...currentTodo, newTodo]);
```
TarasHoliuk, line 45:
```suggestion
      setTodos(currentTodos => [...currentTodos, newTodo]);
```

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,105 @@
...
+      setTitle('');
+    } catch (newError) {
+      setError(Errors.UNABLE_ADD);
+      throw newError;
```
TarasHoliuk, line 49: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

TarasHoliuk (changes requested):
It works :+1: 
But some moments need to be improved - ask in the chat if you need help or have some questions

TarasHoliuk (approved):
Well done :heavy_check_mark:

PR #1025, opened 2024-01-04
feat: add editing and toggle todo

kate-prakofyeva:
[DEMO LINK](https://kate-prakofyeva.github.io/react_todo-app-with-api/)


bertsu02 (approved):
It seems that editing doesn't work.

kate-prakofyeva:
> It seems that editing doesn't work.

Yes! I forgot about event.preventDefault();
fixed

src/UserWarning.tsx
```diff
@@ -1,22 +1,24 @@
...
-      <b> userId </b>
-      {' '}
-      <a href="https://mate-academy.github.io/react_student-registration">
-        here
-      </a>
-      {' '}
-      and save it in the app
-      {' '}
-      <pre>const USER_ID = ...</pre>
+export const UserWarning: React.FC = () => {
```
danon321, line 3: U dont have to import React. Instead of 'React.FC' u can just put 'FC'

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,42 @@
...
+            <Filter filterItem={key} key={key} />
+          ))}
+      </nav>
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        // className={classNames('todoapp__clear-completed', {
+        //   disabled: !isCompleted
+        // })}
```
danon321, line 33: Remove unused code

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { useTodos } from '../../context';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoItem = ({ todo }: Props) => {
+  const { completed, title, id } = todo;
+
+  const [newtitle, setNewTitle] = useState<string | undefined>(title);
```
danon321, line 16: why it can be undefiend?

src/context/TodosContext.tsx
```diff
@@ -0,0 +1,231 @@
...
+  inProgress: 0,
+  deleteTodoFromServer: () => {},
+  deleteCompletedTodos: () => {},
+  addTodoToServer: async () => false,
+  updateTodoOnServer: async () => {},
+  updateAllTodoOnServer: async () => {},
+});
+
+export const TodosProvider: FC<Props> = ({ children }) => {
+  const [loading, setLoading] = useState<number[]>([]);
```
danon321, line 61: i think better name will be loadingTodos

src/context/TodosContext.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+        await Promise.all(queueOfTodosForActivate);
+      }
+    } catch (error) {
+      setErrors(ErrorType.delete);
+    }
+  };
+
+  useEffect(() => {
+    if (userId) {
```
danon321, line 170: Didnt ur AuthContext doesnt prevent from this?

src/types/ErrorType.ts
```diff
@@ -0,0 +1,7 @@
+export enum ErrorType {
+  load = 'Unable to load todos',
+  title = 'Title should not be empty',
+  add = 'Unable to add a todo',
+  delete = 'Unable to delete a todo',
+  update = 'Unable to update a todo',
```
danon321, line 6: Enums should be in upercase

danon321 (changes requested):

src/UserWarning.tsx
kate-prakofyeva: done

src/components/Footer/Footer.tsx
kate-prakofyeva: done

src/components/TodoItem/TodoItem.tsx
kate-prakofyeva: Because at some stage it was necessary to change the type to 'undefined' and if it is a prop it could be 'undefined'. But now, I see, that, I don't need string | undefined anymore, so done

src/context/TodosContext.tsx
kate-prakofyeva: done

src/context/TodosContext.tsx
kate-prakofyeva: `const AuthContext = createContext<number | null>(null);`
because defaultValue can bee null type
I changed to 0, this is ok?
`const AuthContext = createContext<number>(0);`

src/types/ErrorType.ts
kate-prakofyeva: done

kate-prakofyeva: done

src/context/TodosContext.tsx
danon321: Tbh u are the first person i see trying to do seperate Auth for this. Maybe simpler way will be just check if USER_ID is set in App and if not render component with correct information

danon321 (changes requested):
userId should be checked earlier in component

danon321 (approved):

PR #1026, opened 2024-01-04
solutiooooooooooooon

bertsu02:
[DEMO LINK](https://bertsu02.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,195 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { UserWarning } from './UserWarning';
+import { Form } from './components/Form';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Status } from './types/Status';
+import { getTodos, updateTodo, deleteTodo } from './api/todos';
 
-const USER_ID = 0;
+const USER_ID = 12132;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,24 +1,195 @@
...
+    }, 3000);
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((data) => {
+        setTodos(data);
+      })
+      .catch(() => {
+        showErrorNotification('Unable to load todos');
```
MarcinLigmann, line 39: I think this should be in some enum ErrorType, for example:
```
export enum ErrorType {
  load = 'Unable to load todos',
  title = 'Title should not be empty',
  ...
  }
  ```
  and then use 
  ```
  showErrorNotification(ErrorType.load);
  ```

MarcinLigmann (approved):
the code looks ok, the comment below is a suggestion to make it look better, which I also used in my case.

src/App.tsx
mbruhler: true

mbruhler (changes requested):
Good job so far, unfortunately your app has some state managment problems. 
Check the videos attached to see what's happening

1. when deleting todos, state is not updated properly

https://github.com/mate-academy/react_todo-app-with-api/assets/21124163/ba0410cc-fe00-4460-ac73-328923ac8318

2. Due to this fact, or maybe the reason is somewhere else, whe you check the completed todo checkbox, error is thrown


https://github.com/mate-academy/react_todo-app-with-api/assets/21124163/46119941-c109-4a8a-a3ce-171ab63a451b

Fix this in order to complete the task and for further reviews

danon321 (changes requested):
This doesnt work:
save changes when the field loses focus (onBlur);

after edit and click out of input still see the old value

danon321 (approved):

PR #1028, opened 2024-01-04
Develop

adrian200002:
[DEMO LINK](https://adrian200002.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoInfo.tsx
+++ b/src/components/TodoInfo.tsx
@@ -26,26 +26,29 @@ export const TodoInfo = ({ todo }: Props) => {
   } = useTodoContext();
 
   const handleComplete = async (id: number, isCompleted: boolean) => {
-    setIsUpdating((prevIds) => [...prevIds, id]);
-
     if (allTodos) {
-      const updatedTodos = allTodos.map(td => {
-        if (td.id === id) {
-          return {
-            ...todo,
-            completed: isCompleted,
-          };
-        }
-
-        return td;
-      });
-
+      setIsUpdating((prevIds) => [...prevIds, id]);
       try {
         await updateTodo(id, {
           completed: isCompleted,
         });
 
-        setAllTodos(updatedTodos);
+        setAllTodos((prevTodos) => {
+          if (prevTodos) {
+            return prevTodos.map(td => {
+              if (td.id === id) {
+                return {
+                  ...todo,
+                  completed: isCompleted,
+                };
+              }
+
+              return td;
+            });
+          }
+
+          return prevTodos;
+        });
       } catch (error) {
         errorHandler(Errors.updateError);
       } finally {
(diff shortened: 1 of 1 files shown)


jacekkosiba (approved):

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <a
+        href="#/completed"
+        className={classNames('filter__link', {
+          selected: activeFilter === 'Completed',
+        })}
+        data-cy="FilterLinkCompleted"
+        onClick={() => handleTodosFilter('Completed')}
+      >
+        Completed
+      </a>
```
danon321, line 43: you are repeating the code there. Maybe try to do this in loop?

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,60 @@
...
+    visibleTodos,
+    allTodos,
+    setAllTodos,
+    errorHandler,
+  } = useTodoContext();
+
+  const activeTodos = useMemo(() => {
+    let counter = 0;
+
+    allTodos?.forEach(todo => {
```
danon321, line 18: Try .filter() instead and return length

src/components/TodoInfo.tsx
```diff
@@ -0,0 +1,221 @@
...
+    allTodos,
+    setAllTodos,
+    errorHandler,
+    tempTodo,
+    isUpdating,
+    setIsUpdating,
+    inputRef,
+  } = useTodoContext();
+
+  const handleComplete = async (id: number, data: boolean) => {
```
danon321, line 28: Find better name than 'data'

src/components/TodoInfo.tsx
```diff
@@ -0,0 +1,221 @@
...
+import { deleteTodo, updateTodo } from '../api/todos';
+import { useTodoContext } from '../context';
+import { Errors } from '../types/Errors';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo = ({ todo }: Props) => {
+  const [isEditing, setIsEditing] = useState<number | null>(null);
```
danon321, line 13: when i see ;isEditing i feel like it should be boolean value. You can do that or change name to editTodoId

danon321 (changes requested):

mbruhler (changes requested):
Hi,
Unfortunately I can't approve this yet, due to bad state handling.

To reproduce, go to developer tools, click network and then use throttling to slow donw the internet bandwidth, say to 3G SLOW

Then check your app, it messes app what's completed, and also while deleting - bad result is present after deletion of multiple tasks.

Check this out, fix this and request re review

danon321 (approved):

PR #1031, opened 2024-01-05
implement todo

Vlad4567:
- [DEMO LINK](https://Vlad4567.github.io/react_todo-app-with-api/)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,45 @@
...
+        return !todo.completed;
+
+      case Filter.Completed:
+        return todo.completed;
+
+      default:
+        return true;
+    }
+  });
+
```
IvanFesenko, line 24: No need to call filter in case Filter.All it may cause performace issue

IvanFesenko (approved):
Well done. Check my comment

PR #1033, opened 2024-01-07
Develop

baraban2003:


src/App.tsx
```diff
@@ -1,24 +1,200 @@
...
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {/* this buttons is active only if there are some active todos */}
+          {todos.length > 0 && (
```
lerastarynets, line 146:
```suggestion
          {!!todos.length && (
```

src/App.tsx
```diff
@@ -1,24 +1,200 @@
...
+            <ToggleAll
+              onToggleAll={() => toggleAll()}
```
lerastarynets, line 148:
```suggestion
              onToggleAll={toggleAll}
```
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

src/App.tsx
```diff
@@ -1,24 +1,200 @@
...
+              todos={todos}
+            />
+          )}
+
+          {/* Add a todo on form submit */}
+          <TodoInput
+            onSubmit={(title) => addTodo(title)}
```
lerastarynets, line 155:
```suggestion
            onSubmit={addTodo}
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+                data-cy="TodoTitleField"
+                type="text"
+                className="todo__title-field"
+                placeholder="Empty todo will be deleted"
+                value={newTitle}
+                onChange={event => setNewTitle(event.target.value)}
+                onDoubleClick={() => {
+                  setNewTitle(title);
+                  setEditing(false);
+                }}
```
lerastarynets, line 101: Pls move the function inside `onDoubleClick` handler to a separate const and use as a callback here

lerastarynets (changes requested):
Unfortunately, your project is not reachable via the demo link(

baraban2003:
> Unfortunately, your project is not reachable via the demo link(

sorry for that, it was my fault

src/App.tsx
baraban2003: Yes, I know about this and this was my first approach. But the crazy settings of ESLint doesn't allow to do this. 
  116:22  error  JSX props should not use functions                   react/jsx-no-bind
  126:15  error  JSX props should not use functions                   react/jsx-no-bind
  147:17  error  JSX props should not use functions

src/App.tsx
baraban2003: the same as privious

src/App.tsx
lerastarynets: @baraban2003 ok, good that u understand how to do that right

lerastarynets (changes requested):
When I add an item adding is pending, the dropdown with items disappears for a moment, pls fix this by adding a loader on input

baraban2003:
> When I add an item adding is pending, the dropdown with items disappears for a moment, pls fix this by adding a loader on input

Yes, you are right. Thank you!

src/App.tsx
```diff
@@ -1,24 +1,195 @@
...
+      .then(() => setTempTodo(null))
+      .catch(() => {
+        setErrorMessage(Errors.UnableToAddATodo);
+        throw new Error(Errors.UnableToAddATodo);
+      })
+      .finally(
+        () => setLoadingTodosPause(prevIds => prevIds
+          .filter(prevId => prevId !== 0)),
+      );
+  }, [errorMessage]);
```
IvanFesenko, line 67: 2024-01-08_16h36_06
```suggestion
  }, []);
```

src/App.tsx
```diff
@@ -1,24 +1,195 @@
...
+        ))
+      )))
+      .catch(() => setErrorMessage(Errors.UnableToUpdateATodo))
+      .finally(() => {
+        setLoadingTodosPause(currentTodos => currentTodos
+          .filter(todoId => updatedTodo.id !== todoId));
+      });
+  }
+
+  async function toggleAll() {
```
IvanFesenko, line 119:
```suggestion
  const toggleAll = async () => {
```

src/App.tsx
```diff
@@ -1,24 +1,195 @@
...
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {/* this buttons is active only if there are some active todos */}
+          {!!todos.length && (
+            <ToggleAll
+              onToggleAll={() => toggleAll()}
```
IvanFesenko, line 143: No need in additional callback cuz of lint) - fix other places
```suggestion
              onToggleAll={toggleAll}
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,66 @@
...
+      case Filter.completed:
+        return todo.completed;
+
+      case Filter.all:
+        return todo;
+
+      default:
+        return true;
+    }
+  });
```
IvanFesenko, line 38: No need to call `filter` in case `Filter.all` - it may cause performance issues

IvanFesenko (changes requested):
Good job, Check my comments.
1 items -> 1 item
2024-01-08_16h41_34

IvanFesenko (approved):
Well done

PR #1044, opened 2024-01-15
solution

aureliagorzna:
- [DEMO LINK](https://aureliagorzna.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -1,0 +1,77 @@
+/* eslint-disable max-len */
+/* eslint-disable no-console */
+import React, {
+  Dispatch, SetStateAction, useEffect, useRef,
+} from 'react';
+import { postTodo } from '../api/todos';
+import { Todo, ErrorType } from '../types';
+
+const USER_ID = 12139;
+
+type Props = {
+  setTempTodo: Dispatch<SetStateAction<Todo | null>>
+  addTodo: (todo: Todo) => void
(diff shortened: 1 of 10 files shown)


src/components/Header.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  return (
+    <header className="todoapp__header">
+      {/* this buttons is active only if there are some active todos */}
+      <button
+        type="button"
+        className="todoapp__toggle-all active"
+        data-cy="ToggleAllButton"
+        aria-label="text"
+      />
```
WiktoriaBielec, line 63: This button doesn't work

WiktoriaBielec (approved):
One button doesn't work:(

src/components/Header.tsx
mpiskadlo: Yup, missing functionality

mpiskadlo (changes requested):

danon321 (changes requested):
Input doesnt focus at start or whenever
After edit todo to empty string it should be deleted:
if the new title is empty delete the todo the same way the x button does it;

aureliagorzna:
Input is supposed to focus at the end of the string, based on the reference example

mpiskadlo (changes requested):
I think Daniel meant that the input should be initially focused (on the page load)

mbruhler (approved):
Good job! Approved

PR #1051, opened 2024-01-19
add task

Vitalii-Fediaiev:
[DEMO LINK](https://vitalii-dediaiev.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,301 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { TodoList } from './components/TodoList';
+import { TodoFilter } from './components/TodoFilter';
+import * as clientService from './api/Todos';
+import { Todo } from './types/Todo';
+import { Errors } from './types/Errors';
+import { FilterOptions } from './types/FilterOptions';
 
-const USER_ID = 0;
+const USER_ID = 8;
+
+const getFilteredTodos = (todos: Todo[], filterBy: FilterOptions) => {
+  switch (filterBy) {
+    case FilterOptions.Active:
+      return todos.filter(todo => !todo.completed);
+    case FilterOptions.Completed:
+      return todos.filter(todo => todo.completed);
+    default:
+      return todos;
+  }
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState(FilterOptions.All);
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,24 +1,301 @@
...
 import { UserWarning } from './UserWarning';
+import { TodoList } from './components/TodoList';
+import { TodoFilter } from './components/TodoFilter';
+import * as clientService from './api/Todos';
+import { Todo } from './types/Todo';
+import { Errors } from './types/Errors';
+import { FilterOptions } from './types/FilterOptions';
 
-const USER_ID = 0;
+const USER_ID = 8;
```
TarasHoliuk, line 11: Is that the real user id you got in the first task (react-todo-app-load-todos)?

src/App.tsx
```diff
@@ -1,24 +1,301 @@
...
+  }
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState(FilterOptions.All);
+  const filteredTodos = getFilteredTodos(todos, filterBy);
+  const uncompletedTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
+  const [loadingTodosId, setLoadingTodosId] = useState<number[]>([]);
```
TarasHoliuk, line 30:
```suggestion
  const [loadingTodosIds, setLoadingTodosIds] = useState<number[]>([]);
```
:+1: but let's rename it to show there are a lot of ids (array) not only 1

src/App.tsx
```diff
@@ -1,24 +1,301 @@
...
+    clientService.getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => setErrorMasage(Errors.cantGetArray));
+  }
+
+  const addProcessing = (id: number) => {
+    setLoadingTodosId(current => [...current, id]);
+  };
+
+  const removeProcessing = (idToRemove: number) => {
```
TarasHoliuk, line 55: I guess you can simplify it - after any action, you should make this array empty. No need to filter out ids one by one

src/App.tsx
```diff
@@ -1,24 +1,301 @@
...
+  };
+
+  const updateTodo = (updatedTodo: Todo) => {
+    addProcessing(updatedTodo.id);
+
+    return clientService.updateTodo(updatedTodo)
+      .catch((e) => {
+        setErrorMasage(Errors.cantUpdate);
+        removeProcessing(updatedTodo.id);
+        throw e;
```
TarasHoliuk, line 104: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

TarasHoliuk (changes requested):
GJ :+1: 

Let;s improve your code

Also, demo link doesn't work:
image

PR #1053, opened 2024-01-20
Develop

yuran007:
[DEMO LINK](https://yuran007.github.io/react_todo-app-with-api/)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,75 @@
...
+      });
+    }, [complitedTodoIDs, dispatch],
+  );
+
+  const isEmptyTodos = useMemo(() => {
+    return todos.length === 0;
+  }, [todos]);
+
+  return isEmptyTodos
+    ? <></>
```
TarasHoliuk, line 55: just FYI - `null` is valid element as well

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,155 @@
...
+    () => todos.every(todo => todo.completed),
+    [todos],
+  );
+
+  const isDisabled = useMemo(
+    () => !(loadingIDs === null),
+    [loadingIDs],
+  );
+
+  const isGreaterZero = useMemo(
```
TarasHoliuk, line 35: memoization also needs some resources so it's not effective to `useMemo` everywhere - [article](https://www.freecodecamp.org/news/memoization-in-javascript-and-react/#when-to-use-the-usememo-hook), in the article is the link to more info about  "When to useMemo and useCallback"

just FYI

src/store/store.tsx
```diff
@@ -0,0 +1,170 @@
...
+      return state;
+  }
+}
+
+const initialState:State = {
+  todos: [],
+  filter: Filters.All,
+  error: null,
+  tempTodo: null,
+  loadingIDs: null,
```
TarasHoliuk, line 128: since it's an array it's better to set an empty array as a default value, as we usually do with arrays

src/utils/showError.ts
```diff
@@ -0,0 +1,19 @@
+import { ShowError } from '../types/ShowErrors';
+
+export const showErrors = (error: ShowError) => {
```
TarasHoliuk, line 3: seems redundant
Do you use it anywhere? I guess you may delete it when adding to the portfolio (you will create a clear repo, not forked and w/o the mentor's comments)

TarasHoliuk (approved):
You did an amazing job - added a few comments just FYI :+1: 

Approved :heavy_check_mark:

src/store/store.tsx
yuran007: Will get in mind in next project

src/utils/showError.ts
yuran007: You are right

PR #1054, opened 2024-01-20
add_react-app-with-api

Dizelkos:
[DEMO LINK](https://Dizelkos.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/providers/TodosContext.tsx
+++ b/src/providers/TodosContext.tsx
@@ -1,0 +1,135 @@
+import {
+  FC, createContext, useContext, useEffect, useReducer,
+} from 'react';
+import { Todo } from '../types/Todo';
+import { getTodos } from '../api/todos';
+import { USER_ID } from '../utils/userId';
+import { ActionType } from '../types/ActionType';
+import { Errors } from '../types';
+
+type Action =
+  {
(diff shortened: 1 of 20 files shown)


src/services/getFilteredItems.tsx
```diff
@@ -0,0 +1,23 @@
+import { Filters, Todo } from '../types';
+
+export const getFilteredItems = (
```
TarasHoliuk, line 3: Seems you don't use it at all - delete it

src/providers/TodosContext.tsx
```diff
@@ -0,0 +1,135 @@
...
+          payload: {
+            isError: true,
+            errorMessage: Errors.UnableLoad,
+          },
+        });
+      });
+  };
+
+  const updateTodos = () => {
+    fetchTodos();
```
TarasHoliuk, line 104: I guess it's ok here - but with the current API, you always get enough info in response to not make additional requests. For example, when you create a new todo in response you'll get a created todo with `id` generated on the server. You can use this todo from response to update local state

TarasHoliuk (approved):
Well done :heavy_check_mark: 
Added 2 comments just FYI :slightly_smiling_face:

PR #1055, opened 2024-01-25
complete task

boytrend0108:
[DEMO LINK](https://boytrend0108.github.io/react_todo-app-with-api/


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+            type: 'setFilter',
+            payload: Filter.completed,
+          })}
+        >
+          Completed
+        </a>
+      </nav>
+
+      {/* don't show this button if there are no completed todos */}
+      {isCompletedTodo(todos) && (
```
TarasHoliuk, line 64: The filter should always be centered:
/home/vestr/Pictures/Screenshots/Screenshot from 2024-01-25 19-36-09.png

src/services/todosServices.ts
```diff
@@ -0,0 +1,80 @@
...
+    default:
+      return todos;
+  }
+}
+
+export function isCompletedTodo(todos: Todo[]): boolean {
+  return todos.some(todo => todo.completed);
+}
+
+export function isActiveTodo(todos: Todo[]): boolean {
```
TarasHoliuk, line 45: it might be just contants

src/State/State.tsx
```diff
@@ -0,0 +1,146 @@
...
+export const initialState: State = {
+  updatedAt: new Date(),
+  errorMessage: null,
+  todos: [],
+  activeTodos: 0,
+  clearAll: false,
+  filterBy: Filter.all,
+  isSubmitting: false,
+  isEscapeKeyup: false,
+  setAllCompleted: false,
```
TarasHoliuk, line 45: it sounds like function (because of `set` verb). But why do you need it?

src/State/State.tsx
```diff
@@ -0,0 +1,146 @@
+import React, { useReducer } from 'react';
+import { Todo } from '../types/Todo';
+import { Filter } from '../types/Filter';
+
+type State = {
+  updatedAt: Date;
```
TarasHoliuk, line 6: why do you need it?

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,141 @@
...
+  currentTitle,
+  setCurrentTitle,
+}) => {
+  const { title, completed, id } = todo;
+
+  const { clearAll, isEscapeKeyup } = useContext(StateContext);
+  const dispatch = useContext(DispatchContext);
+
+  useEffect(() => {
+    if (clearAll && completed && setIsLoading) {
```
TarasHoliuk, line 34: what about `toggleAll` function - there also should be loading during `toggleAll`

I'd recommend you to save the array of IDs you want to show loading for
For example, if you delete 1 todo set only one `id` into this array, when all - set all todos ids into this array. Use it for all actions

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,141 @@
...
+      setIsLoading(true);
+    }
+  }, [clearAll, completed, dispatch, setIsLoading]);
+
+  useEffect(() => {
+    if (isEscapeKeyup && setCurrentTitle && setIsEditing) {
```
TarasHoliuk, line 40: didn't get why you need `isEscapeKeyup` variable.
Keep `isEditing` state here - it is an individual state of each TodoItem

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,37 @@
...
+  useEffect(() => {
+    setTimeout(() => {
+      dispatch({ type: 'setError', payload: null });
+    }, 3000);
+  }, [errorMessage, dispatch]);
+
+  return (
+    // {/* Notification is shown in case of any error */}
+    // {/* Add the 'hidden' class to hide the message smoothly */}
+    <>
```
TarasHoliuk, line 18: redundant fragment

src/components/TodoField/TodoField.tsx
```diff
@@ -0,0 +1,117 @@
...
+    if (edit.current) {
+      edit.current.focus();
+    }
+  }, [isEditing]);
+
+  function editTodo(event: React.FormEvent) {
+    event.preventDefault();
+    dispatch({ type: 'setIsSubmitting', payload: true });
+
+    const promise: Promise<void> = new Promise((resolve) => {
```
TarasHoliuk, line 38: Why did you create a separate new promise here? just call `updateTodo` inside if

src/components/TodoItem/TodoItem.tsx
TarasHoliuk: Oh, I see, it's part of `TodoField`

Logic in these components is too complicated. You may try to make it simpler and write it's all in one component

src/State/State.tsx
```diff
@@ -0,0 +1,146 @@
...
+  | { type: 'setTempTodo', payload: Todo | null }
+  | { type: 'addActiveTodo', payload: Todo }
+  | { type: 'deleteTodo', payload: number }
+  | { type: 'updateTodo', payload: Todo };
+
+export const initialState: State = {
+  updatedAt: new Date(),
+  errorMessage: null,
+  todos: [],
+  activeTodos: 0,
```
TarasHoliuk, line 40: no need to save it here - just count the actual number using `todos` (filter it and get length, etc)

src/State/State.tsx
```diff
@@ -0,0 +1,146 @@
...
+  clearAll: false,
+  filterBy: Filter.all,
+  isSubmitting: false,
+  isEscapeKeyup: false,
+  setAllCompleted: false,
+  tempTodo: null,
+};
+
+function reducer(state: State, action: Action): State {
```
TarasHoliuk, line 49: it contains too much info
Do not store here individual info like `isEscapeKayup`

TarasHoliuk (changes requested):
Hi! Your app works, but it has too complicated code. Try to simplify it a bit (I left some comments with suggestions)

If you need more info or some help - ask in the chat

l4st1m0za (changes requested):
Great job, one thing to fix:

When I create a todo, and try to change it's title to the space, it's accepting the change and then I have empty todo. Fix this!

image

anastasiiavorobiova (approved):
Great work!

PR #1060, opened 2024-01-29
Develop

stunnerboyone:
[DEMO LINK](https://stunnerboyone.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/Components/ContextProvider/ContextProvider.tsx
+++ b/src/Components/ContextProvider/ContextProvider.tsx
@@ -1,0 +1,125 @@
+import {
+  FC,
+  ReactNode,
+  useState,
+  useEffect,
+  useMemo,
+} from 'react';
+
+import {
+  getTodos,
+  deleteTodo,
+  addTodo,
+  updateTodo,
+} from '../../api/todos';
+
+import { Context } from '../../Context';
+
+import { Todo } from '../../types/Todo';
+import { ErrorMessage } from '../../types/ErrorMessage';
+import { Filter } from '../../types/Filter';
+
+interface Props {
+  children: ReactNode;
+}
+
+export const ContextProvider: FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<Filter>(Filter.ALL);
+  const [
+    errorMessage,
+    setErrorMessage,
+  ] = useState('');
+
+  const USER_ID = 12176;
+
+  const handleErrorChange = (value: string) => {
+    setErrorMessage(value);
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => handleErrorChange(ErrorMessage.UNABLE_TO_LOAD));
+  }, [todos]);
+
+  const filteredTodos = useMemo(() => {
+    if (filter === Filter.ACTIVE) {
+      return todos.filter((item) => !item.completed);
+    }
+
+    if (filter === Filter.COMPLETED) {
+      return todos.filter((item) => item.completed);
+    }
+
+    return todos;
+  }, [filter, todos]);
+
+  const handleActiveTodos = useMemo(() => {
+    return todos.reduce((sum, item) => {
+      if (!item.completed) {
+        return sum + 1;
+      }
+
+      return sum;
+    }, 0);
+  }, [todos]);
+
+  const clearCompleted = () => {
+    const updatedTodos = todos.filter((item) => !item.completed);
+
+    setTodos(updatedTodos);
+  };
+
+  const handleRemoveTodo = (todoId: number) => {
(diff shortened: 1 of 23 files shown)


src/Components/ContextProvider/ContextProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+    setErrorMessage(value);
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => handleErrorChange(ErrorMessage.UNABLE_TO_LOAD));
+  }, [todos]);
+
+  const handleActiveTodos = useMemo(() => {
```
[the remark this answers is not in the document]
TarasHoliuk: Also, if you get rid of it in the Header the footer will be the only component that uses it. So it'll be better to calculate it there and remove it from the context

src/Components/Header/Header.tsx
```diff
@@ -0,0 +1,71 @@
...
+      title: todoTitle,
+      userId: USER_ID,
+      completed: false,
+    };
+
+    addTodo(newTodo)
+      .catch(() => handleErrorChange(ErrorMessage.UNABLE_TO_ADD));
+  };
+
+  const addingTodo = (event:React.FormEvent<HTMLFormElement>) => {
```
TarasHoliuk, line 33: No need to split the logic into 2 functions - write one. You could move the validation into a separate function, but it's very small, so there is no necessary to do it

src/Components/ContextProvider/ContextProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const [
+    errorMessage,
+    setErrorMessage,
+  ] = useState('');
+
+  const USER_ID = 12176;
+
+  const handleErrorChange = (value: string) => {
+    setErrorMessage(value);
+  };
```
TarasHoliuk, line 36: redundant function - just use `setErrorMessage` everywhere instead of `handleErrorChange`

src/Components/ContextProvider/ContextProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+      })
+      .catch(() => setErrorMessage(ErrorMessage.UNABLE_TO_UPDATE));
+  };
+
+  return (
+    <Context.Provider value={{
+      USER_ID,
+      todos,
+      errorMessage,
+      handleErrorChange,
```
TarasHoliuk, line 73:
```suggestion
      handleErrorChange: setErrorMessage,
```

src/Components/ContextProvider/ContextProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+    return todos.reduce((sum, item) => {
+      if (!item.completed) {
+        return sum + 1;
+      }
+
+      return sum;
+    }, 0);
+  }, [todos]);
+
+  const completeAll = () => {
```
TarasHoliuk, line 54: It shouldn't be here - it's related only to `Header` (used only there), so define this function in the Header component

src/Components/Main/Main.tsx
```diff
@@ -0,0 +1,31 @@
+import { useContext, useMemo } from 'react';
+
+import { TodoItem } from '../TodoItem';
+
+import { Context } from '../../Context';
+import { Filter } from '../../types/Filter';
+
+export const Main = () => {
```
TarasHoliuk, line 8:
```suggestion
export const TodoList = () => {
```
it'll better describe what exactly it is

src/Components/Notification/Notification.tsx
```diff
@@ -0,0 +1,31 @@
+import { useContext } from 'react';
+import classNames from 'classnames';
+
+import { Context } from '../../Context';
+
+export const Notification = () => {
+  const { errorMessage, handleErrorChange } = useContext(Context);
+
+  return (
+    <>
```
TarasHoliuk, line 10: redundant [fragment](https://react.dev/reference/react/Fragment)

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const handleEdit = () => {
+    setEditing(true);
+
+    if (inputRef.current) {
+      inputRef.current.value = title;
+    }
+  };
+
+  const handleUpdateTodo = (updatedTodo: Todo) => {
+    setIsLoading(true);
```
TarasHoliuk, line 41: Would be nice to show the loader when you delete/add/delete all completed/toggle all todos, as in the working example.
It's not required in the task (except adding) so optionally - just if you are able to do it.

About adding: in the previous task description was the instruction with `tempTodo` about that:
https://github.com/mate-academy/react_todo-app-add-and-delete#adding-a-todo

l4st1m0za (approved):
Good job!

PR #1061, opened 2024-01-30
add solution

HokageHEL:
[DEMO LINK](https://HokageHEL.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,166 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { Header } from './Components/Header';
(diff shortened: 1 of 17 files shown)


src/App.tsx
```diff
@@ -1,24 +1,166 @@
...
+      title,
+      completed: false,
+    })
+      .then(newTodo => {
+        setTodos((currentTodos) => [...currentTodos, newTodo]);
+        removeLoadingTodoId(0);
+      })
+      .catch((error) => {
+        setErrorMessege(ErrorMessage.UNABLE_TO_ADD_A_TODO);
+        throw error;
```
TarasHoliuk, line 59: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

TarasHoliuk (approved):
Amazing :fire: 

Added one comment, just FYI :slightly_smiling_face:

PR #1063, opened 2024-01-30
done part 3

kasianeno:
[DEMO LINK](https://kasianeno.github.io/react_todo-app-with-api/)


loralevitska (changes requested):
Good job! There are things that should be fixed:

- after double click todo text removes automatically, fix it
- fix adding empty todo 
image

etojeDenys (approved):
good job

PR #1064, opened 2024-01-31
Develop

MarkiMark3:
(https://MarkiMark3.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,363 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import * as todoService from './api/todos';
+import { TodoContext } from './TodoContext';
+import { TodoList } from './TodoList';
+import { TodoItem } from './TodoItem';
 
-const USER_ID = 0;
+const USER_ID = 22;
+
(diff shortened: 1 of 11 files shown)


Esceype (changes requested):
GJ!
Need to fix:
- If edit todo and add spaces at the start in the list it will show without spaces but if we go again to edit mode, then spaces will be there again.

src/App.tsx
```diff
@@ -1,24 +1,340 @@
...
+    errorTimerId = setTimeout(() => {
+      setErrorMessage('');
+    }, 3000);
+  };
+
+  useEffect(() => {
+    todoService.getTodos(USER_ID)
+      .then(data => {
+        setTodos(data);
+      })
```
etojeDenys, line 67:
```suggestion
      .then(setTodos)
```
we can simplfy it

src/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,223 @@
...
+      setTitleEdit(activeTodo.title);
+    }
+
+    return setTodos(todos);
+  };
+
+  const handleBlur = () => {
+    setSelectedTodo(activeTodo.id);
+    setUpdating(true);
+    if (!titleEdit.length) {
```
etojeDenys, line 141:
```suggestion
    if (!titleEdit.trim().length) {
```
better to delete todo in this case

https://github.com/mate-academy/react_todo-app-with-api/assets/70220667/20a6e2c1-2717-4ade-be5b-b73f604b9ba9

etojeDenys (approved):
good job

PR #1081, opened 2024-02-06
add task solution

yaneznayou:
[DEMO LINK](https://yaneznayou.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Filter/Filter.tsx
+++ b/src/components/Filter/Filter.tsx
@@ -1,0 +1,68 @@
+import React, { useCallback } from 'react';
+import classNames from 'classnames';
+import { Status } from '../../types/Status';
+
+interface Props {
+  currentStatus: Status;
+  setStatus: (newStatus: Status) => void;
+}
+
+export const Filter: React.FC<Props> = React.memo(({
+  currentStatus,
+  setStatus,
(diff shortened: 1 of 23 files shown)


src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,68 @@
...
+    if (currentStatus !== Status.active) {
+      setStatus(Status.active);
+    }
+  }, [currentStatus, setStatus]);
+
+  const handleFilterCompleted = useCallback(() => {
+    if (currentStatus !== Status.completed) {
+      setStatus(Status.completed);
+    }
+  }, [currentStatus, setStatus]);
```
vpashko, line 30: in order not to create many handlers, you can create a state filterBy together with possible fields in the enum type and move it to separate file.

```js
export function filterTodos(todosList: Todo[], filterBy: Status) {
  return todosList.filter(todo => {
    switch (filterBy) {
      case Status.Active:
        return !todo.completed;

      case Status.Completed:
        return todo.completed;

      default:
        return todo;
    }
  });
}
```
but your solution is also working

YegorKorenkov (approved):
Awesome work! Good job!

PR #1086, opened 2024-02-07
all tests passed

rtkachuk-moon:
[DEMO LINK](https://rtkachuk-moon.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Main/Main.tsx
+++ b/src/components/Main/Main.tsx
@@ -1,0 +1,49 @@
+import React, { useContext } from 'react';
+import { TodoItem } from '../TodoItem/TodoItem';
+import { TempTodoItem } from '../TempTodoItem/TempTodoItem';
+import { TodosContext } from '../../TodosContext/TodosContext';
+import { filterTodos } from '../../services/filterTodos';
+
+export const Main: React.FC = () => {
+  const {
+    editedTodoTitle,
+    setEditedTodoTitle,
+    handleSpanDoubleClick,
+    handleDeleteTodo,
+    editedTodoId,
+    handleEditTodo,
+    handleCompleteChange,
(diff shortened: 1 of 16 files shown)


src/components/Main/Main.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <TodoItem
+          todo={todo}
+          key={todo.id}
+          editedTodoTitle={editedTodoTitle}
+          setEditedTodoTitle={setEditedTodoTitle}
+          handleSpanDoubleClick={handleSpanDoubleClick}
+          handleDeleteTodo={handleDeleteTodo}
+          editedTodoId={editedTodoId}
+          handleEditTodo={handleEditTodo}
+          handleCompleteChange={handleCompleteChange}
```
OleksandrRezanov, line 36: You have context, why do you need to use props in this case?

DanilWeda, line 36: agree. You can use context inside without props.

DanilWeda (changes requested):
Good job. 
Fix please, the context issues that are described. 
Many thanks!

src/components/Main/Main.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <section className="todoapp__main" data-cy="TodoList">
+
+      {filteredTodos.map(todo => (
+        <TodoItem
+          todo={todo}
+          key={todo.id}
+        />
+
+      ))}
+      {temptTodo && (
```
DanilWeda, line 26:
```suggestion
      {!!temptTodo && (
```
in rendering with &&  need to convert the value to boolean.

DanilWeda (approved):
Good.
Approved.
Thanks!

PR #1103, opened 2024-02-09
Develop

mariuszgit:
https://mariuszgit.github.io/react_todo-app-with-api/


yevhenii-pyl (approved):
Solid!
General advices are the same: 

1. Revise your useMemo + useCallback usage;
2. Context is a solution only for very limited time, but still a solution here.

PR #1120, opened 2024-02-24
solution

katya-sn:
https://katya-sn.github.io/react_todo-app-with-api/

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,54 @@
+import React, { useContext, useMemo } from 'react';
+import { TodoFilter } from './TodoFilter';
+import { TodoContext } from '../context/TodoContext';
+import { Error } from '../types/ErrorMessage';
+import * as TodoService from '../api/todos';
+
+export const Footer: React.FC = () => {
+  const { todos, deleteTodo, handleSetErrorMessage, handleSetUpdatingIds } =
+    useContext(TodoContext);
+
+  const completedIds: number[] = useMemo(
+    () => todos.filter(({ completed }) => completed).map(({ id }) => id),
+    [todos],
+  );
+  const hasActive = useMemo(
+    () =>
+      todos.filter(({ completed }) => {
+        return !completed;
+      }).length,
+    [todos],
+  );
+  const handleDeleteComplTodos = () => {
+    handleSetErrorMessage(Error.none);
+
+    completedIds.forEach(id => {
+      handleSetUpdatingIds(id);
+
+      TodoService.deleteTodo(id)
+        .then(() => deleteTodo(id))
+        .catch(() => handleSetErrorMessage(Error.deleteTodo))
+        .finally(() => handleSetUpdatingIds(null));
+    });
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${hasActive} items left`}
+      </span>
+
+      <TodoFilter />
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        disabled={!completedIds.length}
+        onClick={handleDeleteComplTodos}
+      >
+        Clear completed
+      </button>
+    </footer>
+  );
+};
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -1,0 +1,104 @@
+import React, { useContext, useEffect, useRef, useState } from 'react';
+import classNames from 'classnames';
+import { TodoContext } from '../context/TodoContext';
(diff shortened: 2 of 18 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export const Footer: React.FC = () => {
+  const { todos, deleteTodo, handleSetErrorMessage, handleSetUpdatingIds } =
+    useContext(TodoContext);
+
+  const completedIds: number[] = useMemo(
+    () => todos.filter(({ completed }) => completed).map(({ id }) => id),
+    [todos],
+  );
+  const hasActive = useMemo(
```
TarasHoliuk, line 15:
```suggestion
  const activeTodosCount = useMemo(
```
hasActive is a bit misleading name (we use `has`, `is` prefixes with boolean variables)

src/components/Header.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const [disableInput, setDisableInput] = useState(false);
+
+  const {
+    todos,
+    tempTodo,
+    addTodo,
+    updateTodo,
+    handleSetTempTodo,
+    handleSetErrorMessage,
+    handleSetUpdatingIds,
```
TarasHoliuk, line 19:
```suggestion
    setTempTodo,
    setErrorMessage,
    setUpdatingIds,
```

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+        return wait(3000).then(() => setErrorMessage(null));
+      });
+  }, []);
+
+  useEffect(() => {
+    wait(3000).then(() => setErrorMessage(Error.none));
+  }, [errorMessage]);
+
+  const filteredTodos = getFilteredTodos(todos, filter);
```
TarasHoliuk, line 67: since `filter` is available from context it's better to calculate `filteredTodos` in the component you use it (you use it only in one component - `TodoList`). Or at least wrap it into `useMemo`

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,139 @@
...
+    tempTodo,
+    errorMessage,
+    updatingIds,
+    addTodo,
+    deleteTodo,
+    updateTodo,
+    handleSetErrorMessage,
+    handleSetTempTodo,
+    handleSetUpdatingIds,
+    filterChange,
```
TarasHoliuk, line 133:
```suggestion
    filterChange: setFilter,
```
`filterChange` wrapper is redundant

or just
```suggestion
    setFilter,
```

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,139 @@
...
+    } else {
+      setUpdatingIds([]);
+    }
+  };
+
+  const filterChange = (newFilter: Status) => {
+    setFilter(newFilter);
+  };
+
+  const todosValue = {
```
TarasHoliuk, line 120: it's better to wrap context value into `useMemo`

src/components/Header.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { useContext, useEffect, useRef, useState } from 'react';
+import classNames from 'classnames';
+import { TodoContext } from '../context/TodoContext';
+import { Error } from '../types/ErrorMessage';
+import * as TodoService from '../api/todos';
```
TarasHoliuk, line 5:
```suggestion
import * as todoService from '../api/todos';
```
it's not a class, type or react component

TarasHoliuk (changes requested):
Almost perfect :fire: 

Added a few minor comments only

src/context/TodoContext.tsx
katya-sn: I' ve already  wrapped todosValue, but mentor reject that in the previous task(add and delete) saying that "there is no benefits in memorizing the value of the context because it will be another value every time you change something in context itself". soooo, do I still have to useMemo here?

src/context/TodoContext.tsx
katya-sn: absolutely. I'll move it to the TodoList

src/components/Header.tsx
katya-sn: Ok, now I use just setTempTodo and setErrorMessage (setWHATEVER: React.Dispatch<React.SetStateAction<type>>). but I have to keep SetUpdatingTodos via 

const SetUpdatingIds = (id: number | null) => {
    if (id) {
      setUpdatingIds(currentIds => [...currentIds, id]);
    } else {
      setUpdatingIds([]);
    }
  };    
  
  
  for implementation two paths:

1)passing a number, which gets added to the array.
2)setting the array to empty.

please let me know if there is a better way - I'll do it.

PR #1134, opened 2024-03-11
solution

Yehorf21:
[DEMO LINK](https://Yehorf21.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,214 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { USER_ID, getTodos, patchTodos, postTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { Tabs } from './types/Tabs';
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [input, setInput] = useState('');
+  const [formKey, setFormKey] = useState(0);
+  const [selectedTab, setSelectedTab] = useState(Tabs.All);
+  const [error, setError] = useState('');
+  const [isErrorShown, setIsErrorShown] = useState(false);
+  const [isLoading, setIsLoading] = useState<number[]>([]); // ids of the changed todos
+
+  const areAllCompleted = todos.every(todo => todo.completed);
+  const todoFocus: React.RefObject<HTMLInputElement> = useRef(null);
+
+  useEffect(() => {
+    async function getFetchedTodos() {
+      try {
+        const fetchedTodos = await getTodos();
+
+        setTodos(fetchedTodos);
+      } catch {
+        setError('Unable to load todos');
+        setIsErrorShown(true);
+      }
+    }
+
+    getFetchedTodos();
+  }, []);
+
+  useEffect(() => setFilteredTodos(todos), [todos]);
+
+  useEffect(() => todoFocus.current?.focus(), [todos, isLoading]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">
-          React Todo App - Add and Delete
-        </a>
-      </p>
+  if (error) {
+    setTimeout(() => setIsErrorShown(false), 3000);
+  }
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+  const handleTodoInputFocus = () => todoFocus.current?.focus();
+
+  const handleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInput(event.target.value);
+  };
+
+  const handleOnSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    async function getFetchedTodos() {
+      const newTodo = {
+        userId: USER_ID,
+        title: input,
+        completed: false,
+      };
+      let fetchedTodos;
+      const id = new Date().getTime();
+
+      try {
+        if (input.trim()) {
+          setIsLoading([id]);
+          setTodos([
+            ...todos,
+            {
+              id,
+              ...newTodo,
+            },
+          ]);
+          fetchedTodos = await postTodos(newTodo);
+          setTodos([...todos, fetchedTodos]);
+          setInput('');
+        } else {
+          setError('Title should not be empty');
+          setIsErrorShown(true);
+        }
+      } catch {
+        setError('Unable to add a todo');
+        setIsErrorShown(true);
+        setTodos(todos);
+      } finally {
+        setIsLoading([]);
+      }
+    }
+
+    getFetchedTodos();
+    setFormKey(n => n + 1);
+  };
+
+  const handleCrossThemAll = () => {
+    const todosToChange = todos.filter(t => !t.completed);
+
+    async function crossThemAll() {
+      if (areAllCompleted) {
+        setIsLoading(todos.map(t => t.id));
+      } else {
+        setIsLoading(todosToChange.map(t => t.id));
+      }
+
+      let todosPromise;
+
+      if (areAllCompleted) {
+        todosPromise = todos.map(t => {
+          return patchTodos(t.id, { ...t, completed: false });
+        });
+      } else {
+        todosPromise = todosToChange.map(t => {
+          return patchTodos(t.id, { ...t, completed: true });
+        });
+      }
+
+      try {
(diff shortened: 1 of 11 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+    removeCompleted();
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {todosLeft === 1
+          ? `${todosLeft} item left `
+          : `${todosLeft} items left `}
```
TarasHoliuk, line 72:
```suggestion
          ? `${todosLeft} item left`
          : `${todosLeft} items left`}
```
image

The reason test fails

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+      } catch {
+        setError('Unable to load todos');
+        setIsErrorShown(true);
+      }
+    }
+
+    getFetchedTodos();
+  }, []);
+
+  useEffect(() => setFilteredTodos(todos), [todos]);
```
TarasHoliuk, line 39: no need to create a separate state for `filteredTodos` (if it depends on other state or props values).
Do something like this instead:
```suggestion
  const filteredTodos = useMemo(() => todos.filter(//some filter logic based on `selectedTab`), [todos]);
```

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [input, setInput] = useState('');
+  const [formKey, setFormKey] = useState(0);
+  const [selectedTab, setSelectedTab] = useState(Tabs.All);
+  const [error, setError] = useState('');
+  const [isErrorShown, setIsErrorShown] = useState(false);
```
TarasHoliuk, line 18: this state is redundant as well - it's always the same as `!!error`. if you can define variable based on another state or props value then no need to make it `state`

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+          <button
+            type="button"
+            className={cn('todoapp__toggle-all', {
+              active: areAllCompleted,
+            })}
+            data-cy="ToggleAllButton"
+            onClick={handleCrossThemAll}
+          />
+
+          <form onSubmit={handleOnSubmit} key={formKey}>
```
TarasHoliuk, line 157: why do you need key here?

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,123 @@
...
+      </span>
+
+      <nav className="filter" data-cy="Filter">
```
TarasHoliuk, line 75: It always should have the same position no matter if there is `Clear completed` displayed:
image

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [input, setInput] = useState('');
+  const [formKey, setFormKey] = useState(0);
+  const [selectedTab, setSelectedTab] = useState(Tabs.All);
+  const [error, setError] = useState('');
+  const [isErrorShown, setIsErrorShown] = useState(false);
+  const [isLoading, setIsLoading] = useState<number[]>([]); // ids of the changed todos
```
TarasHoliuk, line 19: a good approach to handle loading status for different todos :+1: 
But now you store array, not boolean value here, so let's rename it to smth like `loadingTodoIds` - id describes better the stored value

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+
+  const areAllCompleted = todos.every(todo => todo.completed);
+  const todoFocus: React.RefObject<HTMLInputElement> = useRef(null);
+
+  useEffect(() => {
+    async function getFetchedTodos() {
```
TarasHoliuk, line 25:
```suggestion
    async function fetchTodos() {
```

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+  const handleTodoInputFocus = () => todoFocus.current?.focus();
+
+  const handleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInput(event.target.value);
+  };
+
+  const handleOnSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    async function getFetchedTodos() {
```
TarasHoliuk, line 60: misleading function name

TarasHoliuk: but you don't need to create a new function - just write the logic right inside `handleOnSubmit`

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+      const newTodo = {
+        userId: USER_ID,
+        title: input,
+        completed: false,
+      };
+      let fetchedTodos;
```
TarasHoliuk, line 66: you don't need this `let` variable here

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+        if (input.trim()) {
+          setIsLoading([id]);
+          setTodos([
+            ...todos,
+            {
+              id,
+              ...newTodo,
+            },
+          ]);
+          fetchedTodos = await postTodos(newTodo);
```
TarasHoliuk, line 79:
```suggestion
          const todo = await postTodos(newTodo);
```
it's one todo in response

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+      const newTodo = {
+        userId: USER_ID,
+        title: input,
+        completed: false,
+      };
+      let fetchedTodos;
+      const id = new Date().getTime();
+
+      try {
+        if (input.trim()) {
```
TarasHoliuk, line 70: you can add this check before `try` (at the very beginning of the function) like

```
if (!input.trim()) {
  // do something
  return;
}
```
such structure will be easier to read

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+          setIsLoading([id]);
+          setTodos([
+            ...todos,
+            {
+              id,
+              ...newTodo,
+            },
+          ]);
```
TarasHoliuk, line 78: what about `tempTodo` approach (not critical - up to you). It would look a bit clearer

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+        await Promise.all(todosPromise);
+        if (areAllCompleted) {
+          setTodos(todos.map(t => ({ ...t, completed: false })));
+        } else {
+          setTodos(todos.map(t => ({ ...t, completed: true })));
+        }
+      } catch {
+        setError('Unable to update todos');
+        setIsErrorShown(true);
+        setTodos(todos);
```
TarasHoliuk, line 131: redundant

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+    }
+
+    getFetchedTodos();
+    setFormKey(n => n + 1);
+  };
+
+  const handleCrossThemAll = () => {
+    const todosToChange = todos.filter(t => !t.completed);
+
+    async function crossThemAll() {
```
TarasHoliuk, line 102: again - you don't need internal function

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+        setIsLoading([]);
+      }
+    }
+
+    getFetchedTodos();
+    setFormKey(n => n + 1);
+  };
+
+  const handleCrossThemAll = () => {
+    const todosToChange = todos.filter(t => !t.completed);
```
TarasHoliuk, line 100: you can define correct  `todosToChange` based on `areAllCompleted` and always use `todosToChange` instead of this if: `if (areAllCompleted)`

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,123 @@
...
+  setIsLoading,
+  setError,
+  setIsErrorShown,
+}) => {
+  const todosLeft = todos.filter(todo => !todo.completed).length;
+  const thereAreCompleted = todos.some(todo => todo.completed);
+
+  const handleTabs = (tab: Tabs) => {
+    switch (tab) {
+      case Tabs.Active:
```
TarasHoliuk, line 32: It's not tabs actually, it's filter
So Tabs name looks a bit misleading

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,123 @@
...
+        setFilteredTodos(todos.filter(todo => !todo.completed));
```
TarasHoliuk, line 33: you shouldn't do it here

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,218 @@
...
+          checked={todo.completed}
+          onClick={handleToggle}
+        />
+      </label>
+
+      {isEdited ? (
+        // eslint-disable-next-line
+        <form
+          onSubmit={handleEditSubmission}
+          onKeyDown={handleCloseEditField}
```
TarasHoliuk, line 177: add it to input instead and remove `// eslint-disable-next-line`

TarasHoliuk (changes requested):

TarasHoliuk (changes requested):
App works as expected, but let's improve your code
Also, try o pass tests

Yehorf21:
I fixed my code according to your guidance. (I didn't change my approach with creating new todos though).

PR #1138, opened 2024-03-12
add edit and toggle for todos

AnastasiaChaikovskaya:
[DEMO LINK](https://AnastasiaChaikovskaya.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,74 @@
+import React from 'react';
+import cn from 'classnames';
+import { Filter } from '../../types/Filter';
+
+type Props = {
+  currentFilter: Filter;
+  setFilter: React.Dispatch<React.SetStateAction<Filter>>;
+  itemsLeft: number;
+  handleClearCompletedTodo: () => void;
+  hasCompleted: boolean;
+};
+
+export const Footer: React.FC<Props> = ({
+  setFilter,
+  itemsLeft,
+  currentFilter,
+  handleClearCompletedTodo,
+  hasCompleted,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${itemsLeft} items left`}
+      </span>
+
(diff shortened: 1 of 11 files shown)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+        return wait(3000).then(() => setErrorMessage(''));
+      })
+      .finally(() => {
+        setLoading(false);
+        setAddTodoId(null);
+      });
+  };
+
+  const onChangeTittle = (event: React.FormEvent<HTMLFormElement>) => {
```
DarkMistyRoom, line 71:
```suggestion
  const onChangeTitle = (event: React.FormEvent<HTMLFormElement>) => {
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,74 @@
...
+  handleClearCompletedTodo,
+  hasCompleted,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${itemsLeft} items left`}
+      </span>
+
+      {/* Active link should have the 'selected' class */}
```
DarkMistyRoom, line 26: remove all unnecessary comments

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,43 @@
...
+  handleDeleteTodo,
+  setLoading,
+  setAddTodoId,
+  changeTitleByid,
+  setErrorMessage,
+  changeCopletedById,
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos?.map(todo => (
```
DarkMistyRoom, line 28: better to don't render this component at all if the is no todos

DarkMistyRoom (changes requested):
Good job! But let's improve it a bit:

- would be better to use Context instead of passing prop through several components

sTorba24 (approved):
Good work! I approve it with small comments, pay attention to them

PR #1150, opened 2024-03-24
React todo app with api

nk8bnj:
[DEMO LINK](https://nk8bnj.github.io/react_todo-app-with-api/)


polosanya (changes requested):
Good job! Let's try to improve your solution a bit)
Completed todos are also triggered on toggleAll button, would be great to fix it.
Same on updating a separate todo.
Also, there is no need to autofocus the input after every action
toggleAll

BudnikOleksii (approved):
Good work 👍

PR #1162, opened 2024-04-01
Develop

Diliradon:
https://diliradon.github.io/react_todo-app-with-api/


src/components/Header.tsx
```diff
@@ -0,0 +1,153 @@
...
+          });
+
+          clearQueryAndEnableInput();
+          setTempTodo(null);
+        })
+        .catch(() => {
+          dispatch({
+            type: ActionTypes.SetValuesByKeys,
+            payload: {
+              errorMessage: 'Unable to add a todo',
```
SanyaBratashchuk, line 66: create variables for strings like that

SanyaBratashchuk (approved):
image

PR #1165, opened 2024-04-02
Develop

ShvetsOlha:
[DEMO LINK](https://ShvetsOlha.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,12 +4,22 @@ import { TodoList } from './components/TodoList';
 import { UserWarning } from './UserWarning';
 import * as todoService from './api/todos';
 import classNames from 'classnames';
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 import { Todo } from './types/Todo';
 import { Filter } from './types/Filter';
 import { Error } from './types/Error';
 import { wait } from './utils/fetchClient';
 import { Footer } from './components/Footer/Footer';
+
+const filterTodos = (currentTodos: Todo[], currentFilter: Filter) => {
+  if (currentFilter === Filter.Active) {
+    return currentTodos.filter((todo: Todo) => !todo.completed);
+  } else if (currentFilter === Filter.Completed) {
+    return currentTodos.filter((todo: Todo) => todo.completed);
+  } else {
+    return currentTodos;
+  }
+};
 
 export const App: React.FC = () => {
   const [input, setInput] = useState('');
@@ -68,23 +78,9 @@ export const App: React.FC = () => {
   }
 
   // eslint-disable-next-line react-hooks/rules-of-hooks
-  const filterTodos = useCallback(
-    (currentTodos: Todo[], currentFilter: Filter) => {
-      if (currentFilter === Filter.Active) {
-        return currentTodos.filter((todo: Todo) => !todo.completed);
-      } else if (currentFilter === Filter.Completed) {
-        return currentTodos.filter((todo: Todo) => todo.completed);
-      } else {
-        return currentTodos;
-      }
-    },
-    [],
-  );
-
-  // eslint-disable-next-line react-hooks/rules-of-hooks
   const filteredTodos = useMemo(() => {
     return filterTodos(todos, filter);
-  }, [filterTodos, filter, todos]);
+  }, [filter, todos]);
 
   const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
     setInput(event.target.value);


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Filter } from '../../types/Filter';
+
+type Props = {
+  filter: Filter;
+  onFilterChange: (newFilter: Filter) => void;
+};
+
+export const TodosFilter: React.FC<Props> = ({
+  filter,
+  onFilterChange = () => {},
```
TarasHoliuk, line 11:
```suggestion
  onFilterChange,
```
default value is redundant - prop is required

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,175 @@
...
+import { useEffect, useRef, useState } from 'react';
+
+type Props = {
+  todo: Todo;
+  onDelete?: (todoId: number) => void;
+  onloadingTodoIds?: number[];
+  onEdit?: (currentTodo: Todo) => void;
+  isEditingTodo?: boolean;
+  setIsEditingTodo?: (isEditing: boolean) => void;
+  onChecked?: (currentTodo: Todo) => void;
```
TarasHoliuk, line 13: why all props are optional? make them required
Same for `TodoList`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+      if (event.key === 'Escape') {
+        if (editedTodo) {
+          setTitle(todo.title);
+        }
+      }
+    } catch (error) {
+      setEditedTodo(todo);
+      setIsEditingTodo(true);
+      throw error;
```
TarasHoliuk, line 68: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,175 @@
...
+  errorFromServer?: boolean;
+};
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  onloadingTodoIds = [],
+  onDelete = () => {},
+  onEdit = () => {},
+  isEditingTodo,
+  setIsEditingTodo = () => {},
```
TarasHoliuk, line 23: it's internal values related to the `TodoItem` only - remove this state the App.tsx and add here

src/App.tsx
```diff
@@ -1,26 +1,395 @@
...
+            todo.id === updatedTodo.id ? updatedTodo : todo,
+          ),
+        );
+        setServerError(false);
+      })
+      .catch(error => {
+        setTodos(todos);
+        setErrorMessage(Error.UnableToUpdate);
+        setServerError(true);
+        throw error;
```
TarasHoliuk, line 230: In case you throw error here no need to `setServerError ` and pass it as a prop, just catch it in the `TodoItem` (or wherever you call this function) and handle additionally

TarasHoliuk: same for all similar cases

src/App.tsx
```diff
@@ -1,26 +1,395 @@
...
+  if (!todoService.USER_ID) {
     return <UserWarning />;
   }
 
+  const handleFilterChange = (newFilter: Filter) => {
+    setFilter(newFilter);
+  };
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const filterTodos = useCallback(
```
TarasHoliuk, line 79: since you do not use values from the scope, but get everything you need as arguments move this function out of component and remove useCallback

src/App.tsx
```diff
@@ -1,26 +1,395 @@
...
+    }
+
+    return () => clearTimeout(timeoutId);
+  }, [errorMessage]);
+
+  if (!todoService.USER_ID) {
     return <UserWarning />;
   }
 
+  const handleFilterChange = (newFilter: Filter) => {
```
TarasHoliuk, line 74: redundant function wrapper - just use `setFilter` everywhere instead of `handleFilterChange`

src/App.tsx
```diff
@@ -1,26 +1,395 @@
...
+    event.preventDefault();
+
+    if (!input.trim()) {
+      setErrorMessage(Error.EmptyTitle);
+
+      return;
+    }
+
+    setIsSubmitting(true);
+    setErrorMessage('');
```
TarasHoliuk, line 112: is it necessary? you do it in handleTitleChange

src/App.tsx
```diff
@@ -1,26 +1,395 @@
...
+    setTempTodo({
+      id: 0,
+      title: input.trim(),
+      userId: todoService.USER_ID,
+      completed: false,
+    });
+
+    if (tempTodo) {
+      setTodos(prevTodos => [...prevTodos, tempTodo]);
+    }
```
TarasHoliuk, line 123: no need to do it, anyway you display tempTodo separately

src/App.tsx
```diff
@@ -1,26 +1,395 @@
...
+            />
+            {tempTodo !== null && (
+              <CSSTransition key={0} timeout={300} classNames="temp-item">
+                <TodoItem todo={tempTodo} />
+              </CSSTransition>
+            )}
+          </>
+        )}
+        {todos.length > 0 && (
+          <footer className="todoapp__footer" data-cy="Footer">
```
TarasHoliuk, line 359: this component is quite big
let's move footer to a separate component

src/App.tsx
```diff
@@ -1,26 +1,395 @@
...
+              isEditing={isEditing}
+              setIsEditing={setIsEditing}
+              onCheckedTodo={handleCheckedTodo}
+              serverError={serverError}
+            />
+            {tempTodo !== null && (
+              <CSSTransition key={0} timeout={300} classNames="temp-item">
+                <TodoItem todo={tempTodo} />
+              </CSSTransition>
+            )}
```
TarasHoliuk, line 355: move it to TodoList - it's also part of the list

TarasHoliuk (changes requested):
GJ, you handle a lot of cases which are usually not handled 👍 

Added comments to improve your code a bit

src/App.tsx
ShvetsOlha: serverError is used in TodoItem not to close Editing Form if there is an error

src/components/TodoItem/TodoItem.tsx
ShvetsOlha: because I am using also TodoItem as a tempTodo and there I do not need all this props

src/App.tsx
ShvetsOlha: yes, it is required in tests

src/components/TodoItem/TodoItem.tsx
ShvetsOlha: tried to remove it, but the tests fail without this line

vadiimvooo (approved):
Well done!)

PR #1176, opened 2024-04-05
add solution

illya-onyshchuk:
[DEMO LINK](https://illya-onyshchuk.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,168 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 
-const USER_ID = 0;
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+import { wait } from './utils/fetchClient';
+import { UserWarning } from './UserWarning';
+import { Footer } from './components/Footer';
+import { creatTodo, deleteTodo, getTodos, USER_ID } from './api/todos';
+import { TodoList } from './components/TodoList';
+import { getFilterTodos } from './utils/getFilterTodos';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Header } from './components/Header';
+import { Errors } from './types/Errors';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [newTitle, setNewTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [errorMessage, setErrorMessage] = useState<Errors>(Errors.Default);
+  const [status, setStatus] = useState<Status>(Status.All);
+  const [isLoading, setIsLoading] = useState(false);
+  const [deletedTodoIds, setDeletedTodoIds] = useState<number[]>([]);
+
+  const focusInput = useRef<HTMLInputElement>(null);
+
+  const isErrorAddTodo = errorMessage === Errors.Add;
+
+  const handleClearError = () => setErrorMessage(Errors.Default);
+
+  const handleError = (error: Errors) => {
+    setErrorMessage(error);
+
+    wait(3000).then(() => handleClearError());
+  };
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        handleError(Errors.Load);
+      });
+  }, []);
+
+  useEffect(() => {
+    focusInput.current?.focus();
+  }, [isErrorAddTodo, todos.length]);
+
+  const filteredTodos = useMemo(() => {
+    return getFilterTodos(todos, status);
+  }, [todos, status]);
+
+  const addTodo = async (creatNewTodo: Omit<Todo, 'id'>) => {
+    setTempTodo({ ...creatNewTodo, id: 0 });
+
+    try {
+      setIsLoading(true);
+      const newTodo = await creatTodo(creatNewTodo);
+
+      setTodos(currentTodos => [...currentTodos, newTodo] as Todo[]);
+      setNewTitle('');
+    } catch {
+      handleError(Errors.Add);
+    } finally {
+      setTempTodo(null);
+      setIsLoading(false);
+    }
+  };
+
+  const removeTodo = (todoId: number) => {
+    setDeletedTodoIds(prevIds => [...prevIds, todoId]);
+
+    return deleteTodo(todoId)
+      .then(() => {
+        setTodos(currentTodos =>
+          currentTodos.filter(todo => todo.id !== todoId),
+        );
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,26 +1,168 @@
...
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        handleError(Errors.Load);
+      });
+  }, []);
+
+  useEffect(() => {
```
ChrisHryts, line 44: why does it depend on todos.length?

src/App.tsx
```diff
@@ -1,26 +1,168 @@
...
+      })
+      .catch(() => {
+        handleError(Errors.Delete);
+      })
+      .finally(() => {
+        setDeletedTodoIds(prevIds => prevIds.filter(id => id !== todoId));
+      });
+  };
+
+  const handleSetStatus = (field: Status) => {
```
ChrisHryts, line 86: there is no point in creating a separate function if it is only one line

ChrisHryts (approved):

src/App.tsx
illya-onyshchuk: Because when you update the todo, you don't need to focus on the input.

src/App.tsx
```diff
@@ -1,26 +1,168 @@
...
+        handleError(Errors.Load);
+      });
+  }, []);
+
+  useEffect(() => {
+    focusInput.current?.focus();
+  }, [isErrorAddTodo, todos.length]);
+
+  const filteredTodos = useMemo(() => {
+    return getFilterTodos(todos, status);
```
Moroz-Dmytro, line 49: You can avoid using `return`

src/App.tsx
Moroz-Dmytro: Good point, this function is not really needed

src/App.tsx
```diff
@@ -1,26 +1,168 @@
...
+      userId: USER_ID,
+      completed: false,
+    };
+
+    addTodo(newTodo);
+  };
+
+  const handelSetNewTitle = (value: string) => {
+    setNewTitle(value);
+  };
```
Moroz-Dmytro, line 112: Same here, you don't really need separate function here

src/components/Header.tsx
```diff
@@ -0,0 +1,80 @@
...
+  newTitle,
+  isLoading,
+  todos,
+  onError,
+}) => {
+  let isEveryTodoctive = todos.every(todo => todo.completed);
+
+  if (!todos.length) {
+    isEveryTodoctive = false;
+  }
```
Moroz-Dmytro, line 32:
```suggestion
  const isEveryTodoctive = todos.every(todo => todo.completed) && !!todos.length;
```

src/components/Header.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const handletoggleAll = async () => {
```
Moroz-Dmytro, line 34:
```suggestion
  const handleToggleAll = async () => {
```

src/components/Header.tsx
```diff
@@ -0,0 +1,80 @@
...
+    try {
+      let unfinishedTasks = todos.filter(todo => !todo.completed);
```
Moroz-Dmytro, line 36: Try to reafactor this part, to use `const` instead

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const isIncludesId = deletedTodoIds?.includes(todo.id);
+
+  useEffect(() => {
+    if (isEditing) {
+      inputField.current?.focus();
+    }
+  }, [isEditing]);
+
+  const handleToggleComplet = async (value: boolean) => {
```
Moroz-Dmytro, line 36:
```suggestion
  const handleToggleStatus = async (value: boolean) => {
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+import cn from 'classnames';
+import { Todo } from '../types/Todo';
+import { updateTodo } from '../api/todos';
+import { Errors } from '../types/Errors';
+
+type Props = {
+  todo: Todo;
+  onDeleteTodo?: (todoId: number) => void;
+  onUpdateTodos?: (newTodo: Todo) => void;
+  deletedTodoIds?: number[];
```
Moroz-Dmytro, line 11: Why all these fields are optional?

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const handleToggleComplet = async (value: boolean) => {
+    try {
+      setIsLoading(true);
+
+      const updatedTodo = await updateTodo(todo.id, {
+        ...todo,
+        completed: value,
+      });
+
+      if (onUpdateTodos) {
```
Moroz-Dmytro, line 45: Try to refactor your solution, not to use such `if`s

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+    } catch {
+      onError(Errors.Update);
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  const handleEditTodo = () => {
+    setIsEditing(true);
+  };
```
Moroz-Dmytro, line 92: Not really need function for this

Moroz-Dmytro (changes requested):

src/components/TodoItem.tsx
illya-onyshchuk: Because when I show `tempTodo`  I don't need these fields.

src/App.tsx
BudnikOleksii: > Good point, this function is not really needed

Not fixed

src/App.tsx
```diff
@@ -1,26 +1,163 @@
...
+import { Errors } from './types/Errors';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [newTitle, setNewTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [errorMessage, setErrorMessage] = useState<Errors>(Errors.Default);
+  const [status, setStatus] = useState<Status>(Status.All);
+  const [isLoading, setIsLoading] = useState(false);
+  const [deletedTodoIds, setDeletedTodoIds] = useState<number[]>([]);
```
BudnikOleksii, line 22:
```suggestion
  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
```
You can store not only todos that you plan to delete.
As of now I don't see any loaders when I click the button to toggle all todos statuses

BudnikOleksii (changes requested):
Almost done, when you toggle all there are no loaders on todos that are loading at that moment

BudnikOleksii (approved):
👍

PR #1182, opened 2024-04-06
add task solution

creonluv:
DEMO LINK](https://creonluv.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/EditForm.tsx
+++ b/src/components/EditForm.tsx
@@ -1,0 +1,125 @@
+import { useEffect, useRef, useState } from 'react';
+import { Todo } from '../types/Todo';
+import { Loader } from './Loader';
+
+type Props = {
+  editTodo: Todo | null;
+  setEdit: (edit: boolean) => void;
+  updtTodo: (todo: Todo, data: Partial<Todo>) => Promise<Todo>;
+  deleteTodo: (id: number) => Promise<void>;
+  setLoadingTodosIds: (todos: number[]) => void;
+  loadingTodosIds: number[];
+  todo: Todo;
+};
+
+export const EditForm: React.FC<Props> = ({
+  editTodo,
+  setEdit,
+  updtTodo,
+  deleteTodo,
+  setLoadingTodosIds,
+  loadingTodosIds,
+  todo,
+}) => {
+  const [newValue, setNewValue] = useState(editTodo?.title || '');
+  const editField = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    editField.current?.focus();
+  }, []);
+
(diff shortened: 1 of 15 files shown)


src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,164 @@
...
+    };
+
+    const tempTodo = {
+      id: 0,
+      userId: 11946,
+      title: newTodoTitle.trim(),
+      completed: false,
+    };
+
+    setTempTodo(tempTodo);
```
TarasHoliuk, line 80: Couldn't you handle `tempTodo` in `addTodo` in the TodoApp.tsx, instead of passing it as prop?

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,164 @@
...
+        setNewTodoTitle('');
+      })
+      .finally(() => {
+        setTempTodo(null);
+        setLoadingTodosIds(loadingTodosIds.filter(id => id !== tempTodo.id));
+        setFocusInput(true);
+      });
+  };
+
+  const handleToggleAll = () => {
```
TarasHoliuk, line 95: you defined all other functions in the TodoApp.tsx except for this one. For this particular function, you passed `setTodos` function. Let's stick to the one approach - pass "prepared" functions instead of raw `setTodos`

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,101 @@
...
+          id="todoCheckbox"
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={completed}
+          onChange={handleChangeCheckbox}
+        />
+      </label>
+
+      {edit ? (
```
TarasHoliuk, line 69: you may get rid of `edit` state and use only `editTodo`. Make `editTodo` prop required in the `EditForm`

src/components/EditForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+  };
+
+  const onKeyEscape = () => {
+    setEdit(false);
+  };
+
+  const onKeyEnter = (event: React.KeyboardEvent) => {
+    event.preventDefault();
+
+    if (newValue?.trim().length === 0) {
```
TarasHoliuk, line 80:
```suggestion
    if (!newValue?.trim().length) {
```

TarasHoliuk (changes requested):
GJ 👍 
Let's refactor your code a bit to reduce props drilling.

P.S. Have you tried to fix this test?:
image

I see it works, but please investigate it if you didn't.

volodymyr-soltys97 (approved):
Good job 👍

PR #1184, opened 2024-04-07
add task solution

Math8585:
[DEMO LINK](https://Math8585.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,184 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useEffect, useState } from 'react';
+import * as todosService from './api/todos';
+import { Todo } from './types/Todo';
+import { NewTodoForm } from './components/FormTodo';
+import { TodoList } from './components/TodoList';
+import classNames from 'classnames';
+import { getfilteredTodos } from './utils/getFilterTodos';
+import { FILTERS } from './types/Filters';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<null | Todo>(null);
+  const [filter, setFilter] = useState(FILTERS.all);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [processingIds, setProcessingIds] = useState<number[]>([]);
+
+  function showErrorMessage(error: string) {
+    setErrorMessage(error);
+    setTimeout(() => {
+      setErrorMessage('');
+    }, 3000);
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">
-          React Todo App - Add and Delete
-        </a>
-      </p>
+  useEffect(() => {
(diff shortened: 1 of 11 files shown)


src/api/todos.ts
```diff
@@ -0,0 +1,20 @@
...
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
+};
+
+export const deleteTodos = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const createTodos = ({ title, userId, completed }: Todo) => {
+  return client.post<Todo>(`/todos`, { title, userId, completed });
+};
```
vladarskyi, line 16: You can simplify it: 
```suggestion
export const createTodos = (data: Todo) => {
  return client.post<Todo>(`/todos`, data);
};
```

src/components/FormTodo.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+    setIsSubmiting(true);
+
+    const tempTodo = { userId, completed: false, title: normalisedTitle };
+
+    const newTodo = {
+      userId,
+      completed: false,
+      title: normalisedTitle,
+      id: +new Date(),
```
vladarskyi, line 51: The server adds the ID automatically, so it's redundant.

vladarskyi (approved):
LGTM 👍

src/App.tsx
```diff
@@ -1,26 +1,184 @@
...
+      .upDateTodos(updatedTodo)
+      .then(todo => {
+        setTodos(currentTodos => {
+          const newTodos = [...currentTodos];
+          const index = newTodos.findIndex(tod => tod.id === updatedTodo.id);
+
+          newTodos.splice(index, 1, todo);
+
+          return newTodos;
+        });
```
BudnikOleksii, line 121: Use `map` instead of `findIndex` and `splice`

src/api/todos.ts
BudnikOleksii: > You can simplify it:

Good point

src/components/FormTodo.tsx
BudnikOleksii: > The server adds the ID automatically, so it's redundant.

Good point

src/utils/getFilterTodos.ts
```diff
@@ -0,0 +1,16 @@
...
+  switch (filter) {
+    case FILTERS.active:
+      preperedTodos = preperedTodos.filter(todo => !todo.completed);
+      break;
+    case FILTERS.completed:
+      preperedTodos = preperedTodos.filter(todo => todo.completed);
+  }
+
+  return preperedTodos;
+}
```
BudnikOleksii, line 16:
```suggestion
export function getfilteredTodos(todos: Todo[], filter: string) {
  switch (filter) {
    case FILTERS.active:
      return todos.filter(todo => !todo.completed);
    case FILTERS.completed:
      return todos.filter(todo => todo.completed);
    default:
      return todos;
  }
}
```
- You don't need to create a copy as `filter` doesn't mutate an array.
- You should always have the default case in the switch

BudnikOleksii (changes requested):
Looks great 🔥 
Just a few minor fixes are required

src/api/todos.ts
Math8585: I use id to control the focus of the input in the FormTodo component. To avoid creating a new state, I use this state:  
const [processingIds, setProcessingIds] = useState<number[]>([]);

src/components/FormTodo.tsx
Math8585: I use id to control the focus of the input in the FormTodo component. To avoid creating a new state, I use this state:  
const [processingIds, setProcessingIds] = useState<number[]>([]);

src/api/todos.ts
Math8585: I can't use this option (because the object contains an id)

BudnikOleksii (approved):
👍

PR #1186, opened 2024-04-07
add task solution

vladarskyi:
[DEMO LINK](https://vladarskyi.github.io/react_todo-app-with-api/)


src/utils/fetchClient.ts
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
hattoriKant0, line 1: It will be better if you remove all comments, especially this comment that disable ESlinter.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,45 @@
...
+      }
+
+      return response.json();
+    });
+}
+
+export const client = {
+  get: <T>(url: string) => request<T>(url),
+  post: <T>(url: string, data: any) => request<T>(url, 'POST', data),
+  patch: <T>(url: string, data: any) => request<T>(url, 'PATCH', data),
```
hattoriKant0, line 43: Don't use `any` type. You need to specify what type of data you want send to the server.

hattoriKant0 (approved):
I approve, but I don't think you need `TodoListExamles.tsx` in final version of the project and I think you need to put `App.tsx` somewhere else because it is your main file and not a component (I am not sure about this).

Nice work!

BudnikOleksii (approved):
Nicely done 🚀

PR #1192, opened 2024-04-08
add solution 

Vlobdam:
[DEMO LINK](https://vlobdam.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,7 +19,7 @@ import { getFilteredTodos } from './utils/getFilteredTodos';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [processingTodos, setProcessingTodos] = useState<Todo[]>([]);
+  const [processingTodos, setProcessingTodos] = useState<number[]>([]);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [errorText, setErrorText] = useState<ErrorText>(ErrorText.NoError);
   const [title, setTitle] = useState('');
@@ -29,19 +29,13 @@ export const App: React.FC = () => {
   );
 
   const preparedTodos = getFilteredTodos(todos, statusFilter);
-  const addToProcessingListById = (todoId: number) => {
-    const todoToAdd = todos.find(todo => todo.id === todoId);
-
-    if (todoToAdd === undefined) {
-      return;
-    }
-
-    setProcessingTodos(prevTodos => [...prevTodos, todoToAdd]);
+  const addToProcessingList = (todoId: number) => {
+    setProcessingTodos(prevTodoIds => [...prevTodoIds, todoId]);
   };
 
-  const removeFromProcessingListById = (todoId: number) => {
+  const removeFromProcessingList = (todoId: number) => {
     setProcessingTodos(prevTodos =>
-      prevTodos.filter(prevTodo => prevTodo.id !== todoId),
+      prevTodos.filter(prevTodoId => prevTodoId !== todoId),
     );
   };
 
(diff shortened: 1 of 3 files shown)


Dmitrel (approved):
on enter before recieved a response

          1) should send a request
          
          what about this test?

src/App.tsx
```diff
@@ -1,26 +1,202 @@
...
+import { ErrorNotification } from './components/ErrorNotification';
+import { Todo } from './types/Todo';
+import { ErrorText } from './types/ErrorText';
+import { StatusFilter } from './types/StatusFilter';
+import { wait } from './utils/fetchClient';
+import { getFilteredTodos } from './utils/getFilteredTodos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [processingTodos, setProcessingTodos] = useState<Todo[]>([]);
```
BudnikOleksii, line 22: You can store only `processingTodos` ids, so it will be much easier

src/App.tsx
```diff
@@ -1,26 +1,202 @@
...
+  };
+
+  const handleError = useCallback((message: ErrorText) => {
+    setErrorText(message);
+    wait(3000).then(() => handleHideError());
+  }, []);
+
+  const handleTodoDelete = (todoId: number) => {
+    addToProcessingListById(todoId);
+    setErrorText(ErrorText.NoError);
```
BudnikOleksii, line 63: You have `handleHideError` that do it

src/App.tsx
```diff
@@ -1,26 +1,202 @@
...
+  const handleTodoAdd = (newTitle: string) => {
+    const newTodo: Todo = {
+      id: 0,
+      userId: USER_ID,
+      title: newTitle,
+      completed: false,
+    };
+
+    setTempTodo(newTodo);
+    setErrorText(ErrorText.NoError);
```
BudnikOleksii, line 96: The same, fix it in all places

src/App.tsx
```diff
@@ -1,26 +1,202 @@
...
+      .then(res => {
+        setTitle('');
+
+        setTodos(prevTodos => {
+          return [
+            ...prevTodos,
+            {
+              ...newTodo,
+              id: res.id,
+            },
```
BudnikOleksii, line 108: You get in the response a new todo, so set and use it

src/App.tsx
```diff
@@ -1,26 +1,202 @@
...
+      completed: !checkedTodo.completed,
+    });
+  };
+
+  useEffect(() => {
+    const fetchTodos = () => {
+      getTodos()
+        .then(data => {
+          setTodos(data);
+        })
```
BudnikOleksii, line 151:
```suggestion
        .then(setTodos)
```

src/components/TodoAppHeader/TodoAppHeader.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+type Props = {
+  todos: Todo[];
+  onAddNewTodo: (_: string) => void;
+  onError: (_: ErrorText) => void;
+  isPosting: boolean;
+  title: string;
+  setTitle: (_: string) => void;
+  inputRef: React.RefObject<HTMLInputElement>;
+  onTodoCheck: (_: Todo) => void;
```
BudnikOleksii, line 14: Why do you call arguments `_`? You should use it only if you don't need the first one and want to use others, but it's not the case here

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

StasSokolov1 (approved):
🔥

PR #1197, opened 2024-04-09
add 3d part task solution

khromakeiqt:
[DEMO LINK](https://khromakeiqt.github.io/react_todo-app-with-api/)

I can't understand why this one test is not passing. Trying to debug for a long time.

 1)
       Renaming
         Edit Form
           on enter before recieved a response
             should send a request:
     AssertionError: Timed out retrying after 4000ms: expected renameCallback to have been called exactly "once", but it was never called
      at Context.eval (webpack://react_todo-app-with-api/./cypress/integration/page.spec.js:1486:36)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,200 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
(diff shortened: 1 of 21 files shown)


Vlobdam (approved):
Can't test, since demo link does not work, but code looks fine

PR #1215, opened 2024-04-20
initial commit

Patryk-Buczkowski:
[DEMO LINK](https://patryk-buczkowski.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/Footer.tsx
+++ b/src/Footer.tsx
@@ -31,17 +31,9 @@ export const Footer = ({
     try {
       const completedTodos = todos.filter(todo => todo.completed);
 
-      completedTodos.map(todo => {
-        removeTodo(todo);
-
-        // setIsLoading(state => [...state, todo.id]);
-
-        // deleteTodo(todo.id).then(() =>
-        //   setTodos(prevTodos =>
-        //     prevTodos.filter(prevTodo => !prevTodo.completed),
-        //   ),
-        // );
-      });
+      for (const todo of completedTodos) {
+        await removeTodo(todo);
+      }
     } catch {
       setErrMessage('An error occurred while deleting completed todos');
     } finally {
--- a/src/Footer.tsx
+++ b/src/Footer.tsx
@@ -1,0 +1,83 @@
+import { Todo } from './types/Todo';
+import { Status } from './enums/status';
+import cn from 'classnames';
+
+type Props = {
+  setErrMessage: (string: string) => void;
+  todos: Todo[];
+  isAnyCompleted: boolean;
+  stat: Status;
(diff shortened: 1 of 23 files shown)


Patryk-Buczkowski:
why this test do not pass

    Todo Toggling

      before receving response

        √ should send an update request

        √ should show loader after click

        √ should not show error message after click

        √ should not hide a todo if filtered

      on success

        1) should toggle a todo

src/Footer.tsx
```diff
@@ -0,0 +1,63 @@
...
+import cn from 'classnames';
+import { useContext } from 'react';
+import { ContextTodos } from './TodoContext';
+
+type Props = {
+  isAnyCompleted: boolean;
+};
+
+export const Footer = ({ isAnyCompleted }: Props) => {
+  const { todos, removeTodo, setErrMessage, setIsLoading, stat, setStat } =
```
BogdanMaliuta, line 11: split it into lines

src/Header.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const isAllActive = todos.every(todo => todo.completed);
+
+  const selectInputTitle = useRef<HTMLInputElement>(null);
+
+  //focus on input for creating new title
+  useEffect(() => {
+    if (selectInputTitle.current) {
+      selectInputTitle.current.focus();
+    }
+  }, [isLoading]);
```
BogdanMaliuta, line 26: you have focus in your app.tsx remove 1 of them

src/Header.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const handleChangeCompleted = async () => {
+    const checkAll = todos.every(todo => todo.completed);
+
+    try {
+      for (const todo of todos) {
+        if (todo.completed === !checkAll) {
```
BogdanMaliuta, line 33: ?
```suggestion
        if (!todo.completed) {
```

BogdanMaliuta (changes requested):
This solution much better then previos solutions.
But have some thing to fix, please check my comments and fix it

BogdanMaliuta (approved):

src/Footer.tsx
Patryk-Buczkowski: when i do it eslint trow err then

src/Header.tsx
Patryk-Buczkowski: is it have the same result?

src/Header.tsx
Patryk-Buczkowski: this is for useRef and both are for diffrent inputs, i found better place for this in header ;)

PR #1223, opened 2024-04-23
Develop

yurovych:
https://yurovych.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,206 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useContext, useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { USER_ID, getTodos, addTodo } from './api/todos';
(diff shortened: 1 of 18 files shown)


src/App.tsx
```diff
@@ -1,26 +1,212 @@
...
+
+  const onSubmit = () => {
+    const newTodo = {
+      id: +new Date(),
+      title: inputValue.trim(),
+      completed: false,
+      userId: USER_ID,
+    };
+
+    if (inputValue.trim() === '') {
```
loralevitska, line 71: Fix all cases
```suggestion
    if (!inputValue.trim()) {
```

src/components/todosList.tsx
```diff
@@ -0,0 +1,22 @@
...
+interface ListProps {
+  items: Todo[];
+}
+
+export const TodosList: React.FC<ListProps> = ({ items }) => {
+  const { todos } = useContext(TodosContext);
+
+  return (
+    !!todos.length && (
+      <>
```
loralevitska, line 15: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots and `ul` paddings

loralevitska (approved):
Well done)

PR #1230, opened 2024-05-01
VB solution #3

vbuldenko:
[DEMO LINK](https://vbuldenko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoItem.tsx
+++ b/src/components/TodoItem.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import cn from 'classnames';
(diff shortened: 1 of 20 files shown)


src/components/TodoItem.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+  useEffect(() => {
+    if (editMode) {
+      inputTitleRef.current?.focus();
+    }
+  }, [editMode]);
+
+  useEffect(() => {
+    if (editMode) {
+      document.addEventListener('keyup', handleKeyUp);
```
TarasHoliuk, line 83: you could add it to the `input` with same effect

TarasHoliuk (approved):
Great job!

But note that the styles folder is called `styles copy`. 🙂

PR #1231, opened 2024-05-02
Develop

ArtemZakhar:


src/App.tsx
```diff
@@ -1,26 +1,54 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
```
undeletable, line 1: [Not critical]
Disabling eslint rules is not a big deal for a homework. But for a real work, I would not recommend to disable eslint rules if errors are produced on your code. These rules were most likely added for some reason, so it's better to fix your code to align it with rules. If this is not possible or you feel that a rule doesn't give enough profit, you can discuss changing its configuration or even removal with the team.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,83 @@
...
+  onCoverShow: React.Dispatch<React.SetStateAction<number[]>>;
+};
+
+export const Footer: FC<Props> = ({ onError, onCoverShow }) => {
+  const { todos, filter } = useContext(StateContext);
+  const dispatch = useContext(DispatchContext);
+
+  const isActiveTodo = todos.some(todo => todo.completed);
+
+  const status = todos.filter(todo => !todo.completed);
```
undeletable, line 18: `pendingTodos` or something like that would be a better choice for the name

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,83 @@
...
+type Props = {
+  onError: (message: string) => void;
+  onCoverShow: React.Dispatch<React.SetStateAction<number[]>>;
+};
+
+export const Footer: FC<Props> = ({ onError, onCoverShow }) => {
+  const { todos, filter } = useContext(StateContext);
+  const dispatch = useContext(DispatchContext);
+
+  const isActiveTodo = todos.some(todo => todo.completed);
```
undeletable, line 16: `hasActiveTodo` would be a better choice for this constant's name. `isActiveTodo` implies to indicate if a certain todo is active - which is not the case here.

src/App.tsx
```diff
@@ -1,26 +1,54 @@
...
 
 export const App: React.FC = () => {
+  const [errorMessage, setErrorMessage] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [coverShow, setCoverShow] = useState<number[]>([]);
+
+  const dispatch = useContext(DispatchContext);
+
+  useEffect(() => {
+    getTodos()
```
undeletable, line 22: [Not critical]
The perfect way to handle this would be defining the whole `getTodos().then(...).catch(...)` logic as a separate function, maybe hook. `() => setErrorMessage('Unable to load todos'))` might be passed as an argument to it.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const status = todos.filter(todo => !todo.completed);
+
+  const handleCleareCompleted = () => {
+    const completedTodo = todos.filter(todo => todo.completed);
+    const arrOfId = completedTodo.map(todo => todo.id);
+
+    onCoverShow(arrOfId);
+
+    completedTodo.forEach(todo => {
+      deleteTodo(todo.id)
```
undeletable, line 27: [Not critical]
The perfect way to handle this would be defining the whole promise chain as a separate function, maybe hook. It might accept list of ids as argument, along with callbacks for catch and finally parts.

src/App.tsx
ArtemZakhar: thank you)
just for your information: this goes with the start code)

src/components/Footer/Footer.tsx
ArtemZakhar: this is a tough word for me) 
will do my best next time)

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+export const Footer: FC<Props> = ({ onError, onCoverShow }) => {
+  const { todos, filter } = useContext(StateContext);
+  const dispatch = useContext(DispatchContext);
+
+  const isActiveTodo = todos.some(todo => todo.completed);
+
+  const status = todos.filter(todo => !todo.completed);
+
+  const handleCleareCompleted = () => {
```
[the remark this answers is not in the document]
ArtemZakhar: I would use Grammarly or Google Translate for such cases next time)

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,83 @@
...
+            {filterAct.title}
+          </a>
+        ))}
+      </nav>
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        {...(!isActiveTodo && {
```
undeletable, line 73: [Not critical]
This part of JSX can be simplified:
```
{...(!isActiveTodo && {
  disabled: true,
  style: { opacity: 0 },
})}
onClick={handleCleareCompleted}
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }, [todos, addingTodo]);
+
+  const isAllTodoComplete = todos.every(todo => todo.completed);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    setAddingTodo(true);
+
+    const trimedTitle = todoTitle.trim();
```
undeletable, line 31: Typo: trim**m**edTitle

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const newTodo: Todo = {
+      title: todoTitle.trim(),
+      completed: false,
+      userId: USER_ID,
+    };
+
+    setTempTodo(newTodo);
+
+    createTodo(newTodo)
+      .then(response => {
```
undeletable, line 49: Same note here as for getting and deleting todos

src/App.tsx
undeletable: Correct, I've seen that - and that's why this is not the reason for rejecting. This is more like advice for your future work.

src/App.tsx
undeletable: Posting it here as well.
It can be a hook (preferrable):
```
const useGetInitialData = ({onSuccess, onError, onFinally}: {onSuccess?: (response) => void; onError?: (error) => void}, onFinally?: () => void) => {
    useEffect(() => {
        getTodos().then(onSuccess).catch(onError).finally(onFinally);
    }, []);
};

const App = () => {
    useGetInitialData({
    onSuccess: response => {
        dispatch({ type: Action.initialTodo, payload: response });
    },
    onError: () => setErrorMessage('Unable to load todos')
});
    ...
};
```
...or just a function:
```
const getInitialData = ({onSuccess, onError, onFinally}: {onSuccess?: (response) => void; onError?: (error) => void}, onFinally?: () => void) => {
    return getTodos().then(onSuccess).catch(onError).finally(onFinally);
};

const App = () => {
    useEffect(() => {
        useGetInitialData({
            onSuccess: response => {
                dispatch({ type: Action.initialTodo, payload: response });
            },
            onError: () => setErrorMessage('Unable to load todos')
        });
    }, []);
    ...
};
```

src/components/Footer/Footer.tsx
undeletable: Something like that:
```
const deleteTodos = ({
  ids,
  onSuccess,
  onError,
  onFnally,
}: {
  ids: string[];
  onSuccess?: (id: string) => void;
  onError?: (id: string) => void;
  onFnally?: (id: string) => void;
}) => {
  ids.forEach(id => {
    deleteTodo(id).then(onSuccess).catch(onError).finally(onFinally);
  });
};
```

PR #1235, opened 2024-05-09
eact_todo-app-with-api (finally part )

georgiyevv:
[DEMO LINK](https://svyatoslavPy.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,11 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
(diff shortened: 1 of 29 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix this failed test.
You need to disable input until we get a response from the server
image

anastasiiavorobiova (changes requested):
Great work! Consider keeping the entered text in the input if the request fails and removing a temporary todo. 
Screenshot 2024-05-09 at 14 22 55

anastasiiavorobiova (changes requested):
Consider removing the temporary todo if the request fails 
Screenshot 2024-05-09 at 15 35 18

ghost (approved):
Good job!

PR #1244, opened 2024-05-13
Develop

marinatea:
[DEMO LINK](https://marinatea.github.io/react_todo-app-with-api/)


nataliaklonowska (changes requested):
- todo on fail when updating should stay in focus and shouldn't remember changes in editing state next time we try to update
- toggle state don't save changes after refresh

marinatea:
It is this test: 1)
       Renaming
         Edit Form
           on fail
             should stay open on fail:
     AssertionError: Timed out retrying after 4000ms: **Expected to find 
element: `[data-cy="TodoTitleField"]`**, but never found it.

I tryied to check with original code, but can't understand where is it :(  @nataliaklonowska

choeqq (approved):

PR #1245, opened 2024-05-13
solution

etuzimek:
[DEMO LINK](https://etuzimek.github.io/react_todo-app-with-api/)


nataliaklonowska (changes requested):
loader are visible only on one todo even on group deletions and toggle all

PR #1249, opened 2024-05-14
todo-app-with-api

MariiaSuper:
https://MariiaSuper.github.io/react_todo-app-with-api/

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,18 +1,18 @@
 import { useContext, useEffect, useRef, useState } from 'react';
 import { DispatchContext, StateContext } from '../../utils/Store';
 import { addTodo, updateTodo } from '../../api/todos';
-// import { Todo } from '../../types/Todo';
-// import { Todo } from '../../types/Todo';
+import classNames from 'classnames';
 
 export const Header = () => {
   const dispatch = useContext(DispatchContext);
   const state = useContext(StateContext);
+  const { loading } = useContext(StateContext);
   const [title, setTitle] = useState('');
   const inputReference = useRef<HTMLInputElement>(null);
 
   useEffect(() => {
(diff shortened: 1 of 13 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  let list = tempTodo ? [...todos, tempTodo] : todos;
+
+  if (filterTodos === FilterOfTodos.Active) {
+    list = list.filter(todo => todo.completed === false);
+  }
+
+  if (filterTodos === FilterOfTodos.Completed) {
+    list = list.filter(todo => todo.completed === true);
+  }
```
etojeDenys, line 17: better to use switch case here instead of if/else

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,250 @@
...
+        payload: { id: todo.id },
+      });
+      dispatch({
+        type: 'setError',
+        payload: 'Unable to delete a todo',
+      });
+    };
+
+    deleteTodo(todo.id).then(successDeleteTodo).catch(unsuccessDeleteTodo);
+  };
```
etojeDenys, line 50: do we need it here?

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,180 @@
...
+        dispatch({
+          type: 'removeFromLoading',
+          payload: { id: todo.id },
+        });
+      });
+  };
+
+  const notCompleted = state.todos.filter(todo => todo.completed === false);
+
+  const AllCompleted = () => {
```
etojeDenys, line 87:
```suggestion
  const allCompleted = () => {
```

etojeDenys (approved):

PR #1251, opened 2024-05-16
Develop

yanamomot:
[DEMO LINK](https://yanamomot.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,158 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { USER_ID, getTodos, getDelete, getAdd } from './api/todos';
+import { Todo } from './types/Todo';
+import { SortType } from './types/SortType';
 
-const USER_ID = 0;
+import { Todos } from './components/Todos/Todos';
+import { Footer } from './components/Footer/Footer';
+import { getFilter } from './components/FilterFunc/FilterFunc';
+import { Form } from './components/Header-Form/Form';
+import { ErrorType } from './types/ErrorType';
+import { ErrorMessage } from './components/ErrorMessage/ErrorMessage';
+import { TodoItem } from './components/TodoItem/TodoItem';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+
+  const [titleNew, setTitleNew] = useState('');
+  const [isSubmitingNewTodo, setIsSubmitingNewTodo] = useState(false);
+
+  const [errorMessage, setErrorMessage] = useState<ErrorType | null>(null);
+  const [sortField, setSortField] = useState(SortType.All);
+
+  const [tempTodo, setTempTodo] = useState<null | Todo>(null);
+  const [isDeleting, setIsDeleting] = useState(false);
+
+  const activeInput = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    setErrorMessage(null);
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage(ErrorType.UnableLoad);
+
+        setTimeout(() => {
+          setErrorMessage(null);
+        }, 3000);
+      });
+  }, []);
+
+  useEffect(() => {
+    activeInput.current?.focus();
+  }, [todos, errorMessage]);
+
+  const sortedTodos = getFilter(todos, sortField);
+
+  const onDelete = (todoId: number) => {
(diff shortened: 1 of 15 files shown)


metrakonon (approved):
Good job! :+1:

src/api/todos.ts
```diff
@@ -0,0 +1,24 @@
...
+  return client.post<Todo>('/todos', { title, completed, userId });
+};
+
+export const getUpdate = (
+  id: Todo['id'],
+  title: Todo['title'],
+  completed: Todo['completed'],
+) => {
+  return client.patch<Todo>(`/todos/${id}`, { title, completed });
+};
```
alexander-ignatow, line 24: Please, fix function names and make them consistent
For instance:
getTodos, deleteTodo, addtodo, updateTodo

alexander-ignatow: Also, add and update methods are inconsistent in terms of parameters:
- add has one parameter of the type `Omit<Todo, 'id'>)`
- update has three primitive parameters

Please either convert update to accept single parameter, or add to accept set of the primitives

src/App.tsx
```diff
@@ -1,26 +1,280 @@
...
+  const [isDeleting, setIsDeleting] = useState(false);
+
+  const [beingEdited, setBeingEdited] = useState<number[]>([]);
+  const [beingUpdated, setBeingUpdated] = useState<number | null>(null);
+
+  const activeInput = useRef<HTMLInputElement>(null);
+  const sortedTodos = getFilter(todos, sortField);
+
+  // others region
+  const allTodosCompleted = todos.every(todo => todo.completed);
```
alexander-ignatow, line 31: I don't quite understand what you optimise with this

src/components/FilterFunc/FilterFunc.tsx
alexander-ignatow, line 1: This looks like a component, but it is not
please rename tsx -> ts
move this file to /utils

src/components/FilterFunc/FilterFunc.tsx
```diff
@@ -0,0 +1,13 @@
+import { SortType } from '../../types/SortType';
+import { Todo } from '../../types/Todo';
+
+export function getFilter(todos: Todo[], sortField: SortType) {
```
alexander-ignatow, line 4: naming id bad

Probably you should rename it to `filter` of something like this

src/components/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,27 @@
+import { ErrorType } from '../../types/ErrorType';
+
+type Props = {
+  errorMessage: ErrorType | null;
+  setErrorMessage: (error: ErrorType | null) => void;
```
alexander-ignatow, line 5: You never set actual ErrorMessage from inside this component, so I suggest rename it into close or clear, etc

```suggestion
  clear: () => void;
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  sortField: SortType;
+  setSortField: (field: SortType) => void;
+  todos: Todo[];
+  clearCompleted: () => void;
+};
+
+export const Footer: React.FC<Props> = ({
```
alexander-ignatow, line 11: React.FC is not needed

https://github.com/typescript-cheatsheets/react?tab=readme-ov-file#section-2-getting-started

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={`filter__link ${sortField === SortType.All && 'selected'}`}
+          data-cy="FilterLinkAll"
+          onClick={() => setSortField(SortType.All)}
+        >
+          All
+        </a>
```
alexander-ignatow, line 38: As you see, <a> elements are almost the same, so you can 

1. Create separate component , say `FooterLink` (optional)
2. Iterate through `SortType` keys and render items for each key (please do it)

alexander-ignatow (changes requested):
Good start, but some cleanup is needed

In addition to the comments please use `classnames` to generate `className` for the component in a cleaner way (multiple places)

src/components/Footer/Footer.tsx
yanamomot: thanks for that! I will fix it

yanamomot: image
image
image
image
image
Hello! I tried implement that, but on working demo on All button should be just '#/' address. Therefore, I left buttons unchanged

alexander-ignatow (approved):
Looks much better now!

Lets move on 👍

alexander-ignatow (approved):
👍

PR #1253, opened 2024-05-16
add_solution

metrakonon:
have a nice day! :tada: 
[DEMO](https://metrakonon.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/Components/Footer/Footer.tsx
+++ b/src/Components/Footer/Footer.tsx
@@ -1,0 +1,86 @@
+import { useContext } from 'react';
+import cn from 'classnames';
+import { deleteTodo } from '../../api/todos';
+import { TodoContext } from '../../Context/TodoContext';
+
+import { Error } from '../../types/Error';
+
+export const Footer: React.FC = () => {
+  const { todos, status, setStatus, deleteTodoLocal, setError, focusInput } =
+    useContext(TodoContext);
+  const activeTodoCount = todos.filter(todo => !todo.completed).length;
+  const completedTodoCount = todos.filter(todo => todo.completed).length;
+
+  const handleStatusChange = (newStatus: string) => () => {
+    setStatus(newStatus);
+  };
+
+  const handleClearAllCompleted = () => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    completedTodos.forEach(todo => {
+      deleteTodo(todo.id)
+        .then(() => {
+          deleteTodoLocal(todo.id);
+        })
+        .catch(() => {
+          setError(Error.DeleteTodo);
+        });
+    });
+
+    focusInput();
+  };
+
+  return (
+    !!todos.length && (
+      <>
+        <footer className="todoapp__footer" data-cy="Footer">
+          <span className="todo-count" data-cy="TodosCounter">
+            {`${activeTodoCount} items left`}
+          </span>
+
+          <nav className="filter" data-cy="Filter">
+            <a
+              href="#/"
+              className={cn('filter__link', { selected: status === 'all' })}
+              data-cy="FilterLinkAll"
+              onClick={handleStatusChange('all')}
+            >
+              All
+            </a>
+
+            <a
+              href="#/active"
+              className={cn('filter__link', { selected: status === 'active' })}
+              data-cy="FilterLinkActive"
+              onClick={handleStatusChange('active')}
+            >
+              Active
+            </a>
+
+            <a
+              href="#/completed"
+              className={cn('filter__link', {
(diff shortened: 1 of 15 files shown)


src/utils/createTodo.ts
```diff
@@ -0,0 +1,14 @@
+function generateId(): number {
+  return Math.floor(Math.random() * 900000) + 100000;
+}
```
kinqbert, line 3: I don't think you need to generate ID. After you send request to add a new todo, server returns this newly added todo with a generated ID.

src/Context/TodoContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setError(Error.LoadTodos);
+      });
+  }, []);
+
+  const filterTodosByStatus = (_todos: Todo[], filterBy: string) => {
```
kinqbert, line 75: I believe this underscore shouldn't be here

kinqbert (approved):
Good job! Mostly code looks fine to me, but there are some major issues that have to be fixed. Check bellow.

There is a major issue with toggle all button. After you press it, it disables main input field (check video).

https://github.com/mate-academy/react_todo-app-with-api/assets/100777995/e0b5a000-afa9-423f-a4c6-9ff6bc12025b

src/Context/TodoContext.tsx
metrakonon: It is to avoid name duplication while still having declarative names for arguments.

src/utils/createTodo.ts
metrakonon: Thanks for catching the bug and mentioning the issue with the ID.

src/utils/clickOutside.tsx
DanilWeda, line 1: Good job!

src/Context/TodoContext.tsx
DanilWeda, line 75: You can move it function outside and use it as helper func to avoid shadow naming

src/Components/Header/Header.tsx
```diff
@@ -0,0 +1,114 @@
...
+          .then(() => {
+            updateTodoLocal(updatedTodo);
+          })
+          .catch(() => {
+            setError(Error.UpdateTodo);
+          })
+          .finally(() => {
+            setIsLoaded(false);
+          });
+      }
```
DanilWeda, line 79: For this, you can try to use the static promise method.

src/Components/Header/Header.tsx
```diff
@@ -0,0 +1,114 @@
...
+    focusInput,
+    setTempTodo,
+    todos,
+    updateTodoLocal,
+  } = useContext(TodoContext);
+  const [title, setTitle] = useState('');
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  const allTodoCompleted = todos.every(todo => todo.completed);
+  const hasTodos = todos.length > 0;
```
DanilWeda, line 26:
```suggestion
  const hasTodos = !!todos.length;
```
as alternative

src/Components/Footer/Footer.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+    completedTodos.forEach(todo => {
+      deleteTodo(todo.id)
+        .then(() => {
+          deleteTodoLocal(todo.id);
+        })
+        .catch(() => {
+          setError(Error.DeleteTodo);
+        });
+    });
```
DanilWeda, line 29: for this, you can try to use promise static methods

src/Components/Footer/Footer.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { deleteTodo } from '../../api/todos';
+import { TodoContext } from '../../Context/TodoContext';
+
+import { Error } from '../../types/Error';
+
+export const Footer: React.FC = () => {
+  const { todos, status, setStatus, deleteTodoLocal, setError, focusInput } =
+    useContext(TodoContext);
+  const activeTodoCount = todos.filter(todo => !todo.completed).length;
+  const completedTodoCount = todos.filter(todo => todo.completed).length;
```
DanilWeda, line 12: You can use one array iteration to get these two values.

DanilWeda (changes requested):
Great! 
Check some small comments for fix ( I left their in other requests, also) 
Many thanks!

alexander-ignatow (approved):
Let's move on with this for now, but at the future please mind my comment

PR #1257, opened 2024-05-17
add solving

ordila:


cypress/integration/page.spec.js
```diff
@@ -1540,7 +1579,7 @@ describe('', () => {
           todos.assertTitle(0, 'Something');
         });
 
-        it('should show trim the new title', () => {
+        it.skip('should show trim the new title', () => {
```
undeletable, line 1582: According to the task conditions:
> If you want to enable tests
> ...
> All tests should pass, even if some behaviour in not well explained in the task

So you should not skip the test

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  const errorMessage = errorType ? errors[errorType].message : '';
+
+  useEffect(() => {
+    if (errorType) {
+      setTimeout(() => {
+        setErrorType(null);
+      }, 3000);
+    }
+  });
```
undeletable, line 18: It makes sense to add `errorType` as the dependency of `useEffect` in this case: there's no need to call this effect on every render, only when `errorType` changes.
Also it makes sense to clear the timeout before `errorType` changes:
```
useEffect(() => {
  // set timeout
  return () => {
    // clear timeout
  };
}, [errorType]);
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,107 @@
...
+    todos,
+    setTodos,
+    setStatus,
+    status,
+    setErrorType,
+    setTodoDeleteId,
+    inputRef,
+  } = useContext(AppContext);
+
+  const activeTodo = todos.reduce(
```
undeletable, line 18: It's better to name the constant `activeTodosCount` or like that

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,107 @@
...
+    (acc, current) => (current.completed ? acc : acc + 1),
+    0,
+  );
+
+  const someCompletedTodos = todos.some(todo => todo.completed);
```
undeletable, line 23: It's better to name the constant `hasCompletedTodos` or like that

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+      setTempTodo({
+        isLoading: true,
+        todo: {
+          title: trimmedQuery,
+          id: 0,
+          completed: false,
+        },
+      });
+      setIsLoading(true);
```
undeletable, line 42: No need to use duplicate `setIsLoading(true);` call

src/components/TodoInfo/TodoInfo.1.tsx
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
undeletable, line 1: Looks like this file can be removed, being almost a duplicate of `TodoInfo.tsx`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,196 @@
...
+      setErrorType('update');
+    } finally {
+      setIsDeleting(false);
+    }
+  };
+
+  const handleFormSubmit = async (event?: React.FormEvent<HTMLFormElement>) => {
+    event?.preventDefault();
+
+    const trimmedTitle = newTitle.trimStart().trimEnd();
```
undeletable, line 84: We can obtain the same result using just `newTitle.trim()`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  const { id, title, completed } = todo;
+
+  const [isDeleting, setIsDeleting] = useState(false);
+
+  const [showForm, setShowForm] = useState(false);
+
+  const [newTitle, setNewTitle] = useState(title);
+
+  const inputRefd = useRef<HTMLInputElement>(null);
```
undeletable, line 24: Looks like there's a typo: inputRef, not inputRef**d**

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,196 @@
...
+    } catch (err) {
+      setErrorType('update');
+      inputRefd.current?.focus();
+    } finally {
+      setIsDeleting(false);
+    }
+  };
+
+  const handleBlur = () => {
+    const trimmedTitle = newTitle.trimStart().trimEnd();
```
undeletable, line 127: We can obtain the same result using just `newTitle.trim()`

undeletable (changes requested):
The solution looks good, just please take care of the comments to improve it

mrkovaliv (approved):
👍 👍 👍

PR #1259, opened 2024-05-17
third part of todo_app

iamgranovskaya:
[DEMO](https://iamgranovskaya.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoContext.tsx
+++ b/src/components/TodoContext.tsx
@@ -1,0 +1,219 @@
+import React, { useEffect, useMemo, useReducer, useState } from 'react';
+import { Todo } from '../types/Todo';
+import { deleteTodo, getTodos } from '../api/todos';
+
+export enum FilterBy {
+  All = 'ALL',
+  Active = 'ACTIVE',
+  Completed = 'COMPLETED',
+}
+
+export type FilterAction = FilterBy.All | FilterBy.Active | FilterBy.Completed;
+
+type TodoContextType = {
+  todos: Todo[];
+  originalTodos: Todo[];
+  dispatch: (action: Action) => void;
+  filteredBy: FilterBy;
+  setFilteredBy: (type: FilterBy) => void;
+  error: string | null;
+  handleError: (message: string) => void;
+  deleteCandidates: number[];
+  handleClearCompleted: (ids: number[]) => void;
+  tmpTodo: Todo | null;
+  handleTmpTodo: (todo: Todo | null) => void;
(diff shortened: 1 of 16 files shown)


src/components/TodoContext.tsx
```diff
@@ -0,0 +1,219 @@
...
+      return todos;
+  }
+}
+
+export const TodoProvider: React.FC<Props> = ({ children }) => {
+  const [todos, dispatch] = useReducer(reducer, []);
+  const [tmpTodo, setTmpTodo] = useState<Todo | null>(null);
+  const [filteredBy, setFilteredBy] = useState<FilterBy>(FilterBy.All);
+  const [error, setError] = useState<string | null>(null);
+  const [deleteCandidates, setDeleteCandidates] = useState<number[]>([]);
```
SanyaBratashchuk, line 148: you don't need `deleteCandidates` state. You can just remove todos from `todos` state based on `todo.completed`

src/components/TodoContext.tsx
```diff
@@ -0,0 +1,219 @@
...
+    if (deleteCandidates.length) {
+      deleteCandidates.forEach(id => {
+        deleteTodo(id)
+          .then(() => {
+            dispatch({ type: ActionNames.Delete, payload: id });
+          })
+          .catch(() => handleError(errors.DeleteTodo));
+      });
+    }
+  }, [deleteCandidates]);
```
SanyaBratashchuk, line 206:
```suggestion
```

SanyaBratashchuk (changes requested):
handle plural it should be `1 item` but `2 items` 
image

src/components/TodoContext.tsx
iamgranovskaya: Dind't get your point, I have this state not only for deleting from server I also use it for loading when I delete all completed. Could you be more specific.

SanyaBratashchuk: To show loader you can have just another loadingIds state and store there an array of todoIds that are deleting

SanyaBratashchuk (changes requested):
Fix comments left on previous review. 
If you have any questions feel free to ask in fe_chat

src/components/TodoContext.tsx
iamgranovskaya: @SanyaBratashchuk why would i need another one if I alredy have it?

SanyaBratashchuk (approved):

PR #1260, opened 2024-05-18
add task solution

JohnyJDX:
[DEMO LINK](https://JohnyJDX.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,82 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { USER_ID, getTodos } from './api/todos';
+
+import ErrorNotification from './components/ErrorNotification';
+import Footer from './components/Footer';
+import Header from './components/Headre';
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
+import TodoList from './components/TodoList';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
-
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [selectedStatus, setSelectedStatus] = useState(Status.All);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletingId, setDeletingId] = useState(0);
+  const [updatingTodoId, setUpdatingTodoId] = useState(0);
+
+  useEffect(() => {
+    const fetchTodos = async () => {
+      try {
+        const todosFromServer = await getTodos();
+
+        setTodos(todosFromServer);
+      } catch {
+        setErrorMessage('Unable to load todos');
+        setTimeout(() => setErrorMessage(''), 3000);
+      }
+    };
+
+    fetchTodos();
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">
-          React Todo App - Add and Delete
-        </a>
-      </p>
+  const onErrorMessage = (message: string) => {
+    setErrorMessage(message);
+    setTimeout(() => setErrorMessage(''), 3000);
+  };
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+  return (
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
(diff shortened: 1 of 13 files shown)


src/components/Headre.tsx
alexander-ignatow, line 1: Filename problem :)

alexander-ignatow (changes requested):
Overall looks good, but you need to refactor props passed to the components

For instance:
`TodoItem` has `setTodos` property, that allows to change the entire set of Todos, which is bad, as all you need is to update THE SINGLE todo to the list.
So the fix here will be replacing this prop with `onTodoChange` that will accept the single todo, while the actual list will be changed where it is created, without exposing the property to change the entire set of todos.

Please review other components as well

Actually, way better solution would be moving todo's collection to the context, but it is up to you

src/components/Header.tsx
```diff
@@ -0,0 +1,123 @@
+import { Dispatch, FC, FormEvent, useEffect, useState } from 'react';
+import { USER_ID, addTodo, updateTodos } from '../api/todos';
+
+import useFocusInput from '../hooks/useFocusInput';
+import { Todo } from '../types/Todo';
+
+interface Props {
+  onErrorMessage: (message: string) => void;
+  setTodos: Dispatch<React.SetStateAction<Todo[]>>;
```
alexander-ignatow, line 9: `setTodos` Should not be here. please remove/replace it.

As long as Header is responsible only for adding single `todo` it's props should reflect this.
`setTodos` allows for instance removing all todos - that is way beyond the Header responsibility.

src/components/Header.tsx
```diff
@@ -0,0 +1,123 @@
...
+  setTempTodo: Dispatch<React.SetStateAction<Todo | null>>;
+  deletingId: number;
+  todos: Todo[];
```
alexander-ignatow, line 12: Why you pass all todo's to the Header?
You don't really need this info there, so please remove it.

src/components/Header.tsx
```diff
@@ -0,0 +1,123 @@
+import { Dispatch, FC, FormEvent, useEffect, useState } from 'react';
+import { USER_ID, addTodo, updateTodos } from '../api/todos';
+
+import useFocusInput from '../hooks/useFocusInput';
+import { Todo } from '../types/Todo';
+
+interface Props {
+  onErrorMessage: (message: string) => void;
+  setTodos: Dispatch<React.SetStateAction<Todo[]>>;
+  setTempTodo: Dispatch<React.SetStateAction<Todo | null>>;
```
alexander-ignatow, line 10: Don't expose more than you need 
```suggestion
  setTempTodo: (todo: Todo | null) => void;
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,89 @@
+import { Dispatch, FC } from 'react';
+
+import { deleteTodo } from '../api/todos';
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
+interface Props {
+  todos: Todo[];
```
alexander-ignatow, line 8: You don't need this property - passing full list is an overhead

src/components/Footer.tsx
```diff
@@ -0,0 +1,89 @@
...
+  selectedStatus: Status;
+  setSelectedStatus: (status: Status) => void;
+  setTodos: Dispatch<React.SetStateAction<Todo[]>>;
```
alexander-ignatow, line 11: this property gives too much opportunities to the footer - please remove and replace

src/components/Footer.tsx
```diff
@@ -0,0 +1,89 @@
...
+  setIsDeleting: Dispatch<React.SetStateAction<number>>;
```
alexander-ignatow, line 12: This property is not clear/redundant
Please remove or rename and change the type (without react.Dispatch)

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,19 @@
+import { FC } from 'react';
+
+interface Props {
+  errorMessage: string;
+}
+
+const ErrorNotification: FC<Props> = ({ errorMessage }) => {
```
alexander-ignatow, line 7: Optional - React.FC is not required

```suggestion
const ErrorNotification = ({ errorMessage }) => {
```

src/App.tsx
```diff
@@ -1,26 +1,154 @@
...
+  const [deletingId, setDeletingId] = useState(0);
+  const [updatingTodoId, setUpdatingTodoId] = useState(0);
+
+  const allCompletedTodos = todos.every(todo => todo.completed);
+  const activeTodosCount = todos.filter(todo => !todo.completed).length;
+  const completedTodosCount = todos.filter(todo => todo.completed).length;
+
+  const todosLength = todos.length;
+
+  useEffect(() => {
```
mrkovaliv, line 32: move on this logic to custom hook

mrkovaliv (approved):
👍 👍 👍 👍 

Added some comments)))

PR #1264, opened 2024-05-22
Solution

odunchyk:
[DEMO LINK](https://odunchyk.github.io/react_todo-app-with-api/)


GrigoryyKach (approved):
Good job!
I just don't understand why you need two identical Form and Header components

odunchyk:
> Good job! I just don't understand why you need two identical Form and Header components

ooops, seems like renaming of Form somehow gone wrong and now i have redundant Form component. should remove Form

fixed, thank you

DanilWeda (approved):
Great! Approved!
Thanks)

PR #1266, opened 2024-05-23
final solition

sevenlays:
[DEMO LINK](https://sevenlays.github.io/react_todo-app-with-api/)

Help me make it perfect!:)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,84 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import { useState } from 'react';
+import cn from 'classnames';
 
-const USER_ID = 0;
+import './styles/trasitiongroup.scss';
+import { TransitionGroup, CSSTransition } from 'react-transition-group';
 
-export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+import { getFilteredTodos } from './utils/getFilteredTodos';
+
+import { useTodosContext } from './context/TodosContext';
+
+import { ErrorsContainer } from './components/ErrorsContainer/ErrorsContainer';
+import { TodoItem } from './components/TodoItem/TodoItem';
+import { Header } from './components/Header/Header';
+import { Footer } from './components/Footer/Footer';
+
+import { Status } from './types/Status';
+import { deleteTodo } from './api/todos';
+
+export const App = () => {
+  const {
+    todos,
+    errorMessage,
+    tempTodo,
+    setLoadingIds,
+    setTodos,
+    setErrorMessage,
+  } = useTodosContext();
+  const [status, setStatus] = useState<Status>(Status.All);
+
+  const filteredTodos = getFilteredTodos(todos, status);
+
+  function deletingTodo(todoId: number) {
+    setLoadingIds(prev => [...prev, todoId]);
+    deleteTodo(todoId)
+      .then(() => {
+        setTodos(currentTodos =>
+          currentTodos.filter(todoForDeleting => todoForDeleting.id !== todoId),
+        );
+      })
+      .catch(() => {
+        setErrorMessage('Unable to delete a todo');
+      })
+      .finally(() => {
+        setLoadingIds([]);
+      });
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">
-          React Todo App - Add and Delete
-        </a>
-      </p>
+    <div className={cn('todoapp', { 'has-error': errorMessage })}>
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <Header />
+
+        <section className="todoapp__main" data-cy="TodoList">
+          <TransitionGroup>
+            {filteredTodos.map(todo => (
+              <CSSTransition timeout={300} key={todo.id} classNames="item">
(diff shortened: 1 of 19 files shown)


AYupko (approved):
Good job!

src/utils/tempTodo.ts
alexander-ignatow, line 1: Not critical - but that probably can be placed with TodosApi

src/hooks/useTodos.ts
```diff
@@ -0,0 +1,57 @@
...
+export type UseTodos = {
+  todos: Todo[];
+  setTodos: Dispatch<SetStateAction<Todo[]>>;
+  errorMessage: string;
+  setErrorMessage: (error: string) => void;
+  tempTodo: Todo | null;
+  setTempTodo: (todo: Todo | null) => void;
+  loadingIds: number[];
+  setLoadingIds: Dispatch<SetStateAction<number[]>>;
+};
```
alexander-ignatow, line 14: This looks like an overhead

You can simply remove this type and still benefit of all Typescrit things

src/context/TodosContext.tsx
```diff
@@ -0,0 +1,40 @@
...
+  loadingIds: [],
+  setLoadingIds: () => {},
+};
+
+const TodosContext = createContext<UseTodos>(initialContextValue);
+
+export const TodosProvider: React.FunctionComponent<TodosProviderProps> = ({
+  children,
+}) => {
+  const todosData = useTodos();
```
alexander-ignatow, line 25: I don't see the point in `useTodos` hook
It just adds unnecessary complications without giving you anything.

As long as you use it __only__ inside this context, you can simply throw away this hook and move all code directly to the context.

src/components/TodoItem/TodoLoader.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const TodoLoader: React.FunctionComponent<TodoLoaderProps> = ({
+  todo,
+}) => {
+  const { loadingIds } = useTodosContext();
+
+  return (
+    <div
+      data-cy="TodoLoader"
+      className={cn('modal overlay', {
+        'is-active': loadingIds?.includes(todo.id),
```
alexander-ignatow, line 18: Looking at your types, you should not need `?` in `loadingIds?.includes`

If your code actually crashes without it, this may be a sign of some bug

Probably non-mandatory fix, but it makes sense to re-check

src/components/ErrorsContainer/ErrorsContainer.tsx
alexander-ignatow, line 1: Here and in other components:

It is a common practice to have an "entry point" to the component or module named `index.tsx`
It makes code more readable and slightly improves imports
Just rename `ErrorsContainer.tsx` to `index.tsx` and see what happens

src/App.tsx
```diff
@@ -1,26 +1,84 @@
...
+    tempTodo,
+    setLoadingIds,
+    setTodos,
+    setErrorMessage,
+  } = useTodosContext();
+  const [status, setStatus] = useState<Status>(Status.All);
+
+  const filteredTodos = getFilteredTodos(todos, status);
+
+  function deletingTodo(todoId: number) {
```
alexander-ignatow, line 34: Why not move it to your context?

But naming needs improvement

src/App.tsx
```diff
@@ -1,26 +1,84 @@
...
+    todos,
+    errorMessage,
+    tempTodo,
+    setLoadingIds,
+    setTodos,
+    setErrorMessage,
+  } = useTodosContext();
+  const [status, setStatus] = useState<Status>(Status.All);
+
+  const filteredTodos = getFilteredTodos(todos, status);
```
alexander-ignatow, line 32: This logic may also go to your context

The responsibility should be the following:
Context rules all data
Components decide how this data is displayed

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,121 @@
...
+  function createNewTodo(currentTitle: string) {
+    const newTodo: Todo = {
+      ...temporaryTodo,
+      title: currentTitle.trim(),
+    };
+
+    return newTodo;
+  }
+
+  function addNewTodo(newTodo: Todo) {
```
alexander-ignatow, line 31: Add new todo logic  may also go to the context

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,121 @@
...
+        setErrorMessage('Unable to add a todo');
+        setTempTodo(null);
+      })
+      .finally(() => {
+        setTempTodo(null);
+        setLoadingIds([]);
+      });
+  }
+
+  function updatingTodo(todoForUpdate: Todo, updatedTodos: Todo[]) {
```
alexander-ignatow, line 47: update todo logic can also be in the context (naming needs to be fixed)

```suggestion
  function updateTodo(todoForUpdate: Todo, updatedTodos: Todo[]) {
```
```suggestion
  function updatingTodo(todoForUpdate: Todo, updatedTodos: Todo[]) {
```

src/context/TodosContext.tsx
alexander-ignatow: If you decide keeping it, it makes sense at least move it to the context/TodosContext folder, as it relates only to the context

alexander-ignatow (changes requested):
Overal - very good

Please see my comments and slightly change the structure of the code

The point is the following: 

- Context is responsible for data and data manipulations
- Components are responsible for displaying the data
- If you wish you can keep `useTodos` hook - then it will be responsible for keeping the state (i.e. storing the data)

These changes will make your code more structured and easier to maintain in future

src/hooks/useTodos.ts
alexander-ignatow: > setTodos: (value: Todo[]) => void; - in this case i have problems when func setstate

I missed, that you are using both versions of the setter
then it is ok to keep your version

alexander-ignatow (approved):
Looks good 👍 !

Still, for the future I advice change the naming of the files like this:

ErrorsContainer/ErrorsContainer.tsx -> ErrorsContainer/index.tsx   ;)

PR #1270, opened 2024-05-24
add solution

D-Kondratiev:
[DEMO LINK](https://D-Kondratiev.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/ErrorNotification.tsx
+++ b/src/components/ErrorNotification.tsx
@@ -1,0 +1,41 @@
+import React, { useEffect } from 'react';
+import cn from 'classnames';
+
+import { Errors } from '../types/Errors';
+
+interface Props {
+  errorMessage: Errors | null;
+  onDeleteError: () => void;
+}
+
+export const ErrorNotification: React.FC<Props> = ({
+  errorMessage,
+  onDeleteError,
+}) => {
+  useEffect(() => {
+    if (!errorMessage) {
+      return;
+    }
+
+    const timerID = setTimeout(() => onDeleteError(), 3000);
+
+    return () => clearTimeout(timerID);
+  }, [errorMessage]);
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={cn('notification is-danger is-light has-text-weight-normal', {
+        hidden: !errorMessage,
+      })}
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={onDeleteError}
+      />
+      {errorMessage}
+    </div>
+  );
+};
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -1,0 +1,59 @@
+import React, { useEffect } from 'react';
+import cn from 'classnames';
+import { Todo } from '../types/Todo';
+
+interface Props {
+  inputRef: React.RefObject<HTMLInputElement>;
+  todos: Todo[];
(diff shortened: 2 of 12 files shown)


MarchAlbion (approved):

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const ErrorNotification: React.FC<Props> = ({
+  errorMessage,
+  onDeleteError,
+}) => {
+  useEffect(() => {
+    if (!errorMessage) {
+      return;
+    }
+
+    const timerID = setTimeout(() => onDeleteError(), 3000);
```
alexander-ignatow, line 20: please make `3000` a constant with some explaining name

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+  onEmptyTitleDelete?: () => void;
+}
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  changingIDs,
+  onDelete = () => Promise.resolve(),
+  onToggle = () => {},
+  onRename = () => Promise.resolve(),
+  onEmptyTitleDelete = () => {},
```
alexander-ignatow, line 20: Why do you need these defaults?

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,42 @@
...
+  const options: RequestInit = { method };
+
+  if (data) {
+    options.body = JSON.stringify(data);
+    options.headers = {
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  return wait(100)
```
alexander-ignatow, line 26: This should be a constant as well

alexander-ignatow (changes requested):
Please make some minor changes

the solution is good 👍 , however the code was better structured if you used context.

src/components/TodoItem.tsx
D-Kondratiev: for prevent errors

src/components/Header.tsx
```diff
@@ -0,0 +1,59 @@
...
+  onSubmit,
+  titleText,
+  onType,
+  toggleAllTodos,
+}) => {
+  const areAllTodosCompleted: boolean = todos.every(todo => todo.completed);
+
+  useEffect(() => {
+    inputRef.current?.focus();
+  }, [tempTodo]);
```
DanilWeda, line 30: You can disable esling rule for this case.

src/components/Header.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <header className="todoapp__header">
+      {todos.length !== 0 && (
```
DanilWeda, line 34:
```suggestion
      {!!todos.length  && (
```
as alternative

DanilWeda (changes requested):
Great! 
Check some minor comments.
Thanks)

src/components/TodoItem.tsx
alexander-ignatow: Ok, a few more notes

1. You have already made these props optional - so no point pass the defaults. Just add checks for undefined (via `?.`) and it will solve your problem much cleaner
2. `onEmptyTitleDelete` looks an overkill - how exactly it is different from `onDelete`?

alexander-ignatow (approved):
Looks good 👍 

Please, see my comment below to consider it for the future

PR #1272, opened 2024-05-24
Solution

dyakym:
[DEMO LINK](https://dyakym.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,216 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { FormEvent, useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import {
+  USER_ID,
+  getTodos,
+  createTodo,
+  deleteTodo,
+  updateTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { getFilteredTodos } from './utils/getFiltredTodos';
+import { FilterStatus } from './types/FilterStatus';
+import { Footer } from './components/Footer/Footer';
+import { Error } from './types/Error';
+import cn from 'classnames';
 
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,26 +1,216 @@
...
+        className={cn(
+          'notification is-danger is-light has-text-weight-normal',
+          { hidden: !error },
+        )}
+      >
+        <button
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={() => setError('')}
```
undeletable, line 210: [Not critical]
Clearing error function is used in multiple places, so it can be defined once as a separate function and called in all those places.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,115 @@
+import { Todo } from '../../types/Todo';
+import React, { useState, useEffect, FormEvent } from 'react';
+import cn from 'classnames';
+
+interface Props {
```
undeletable, line 5: [Not critical]
A better name would be `TodoItemProps` or something like that

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,38 @@
...
+  onRename,
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos.map(todo => (
+        <TodoItem
+          key={todo.id}
+          todo={todo}
+          onDelete={onDelete}
+          isLoading={loadingTodos.includes(todo.id)}
```
undeletable, line 28: [Not critical]
The logic of detecting if a todo is loading (`loadingTodos.includes(...)`) is used in multiple places. So you can define it as a separate function and call it from all these places.

undeletable (changes requested):
The solution looks and works good. There just are a couple of places which should be fixed:
- errors enum naming, mentioned in the comments;
- also please fix the code formatting. See eslint output:
```
  Line 32:11:  Insert `⏎······`  prettier/prettier

src/components/TodoItem/TodoItem.tsx
  Line 51:1:  Expected indentation of 6 spaces but found 7  @typescript-eslint/indent
  Line 51:7:  Delete `·`                                    prettier/prettier
```

Also I have some non-critical recommendations on code improvements, you can check them in the comments below.

mrkovaliv (approved):

PR #1276, opened 2024-05-24
add task solution

Follder:
[DEMO LINK](https://follder.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,311 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useRef, useState } from 'react';
+import {
+  USER_ID,
+  deleteTodo,
+  getTodos,
+  postTodo,
+  updateTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import cn from 'classnames';
+import { TodoItem } from './components/TodoItem';
 
-const USER_ID = 0;
+type Sort = 'All' | 'Active' | 'Completed';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortBy, setSortBy] = useState<Sort>('All');
+  const [title, setTitle] = useState('');
+  const [loadingIds, setLoadingIds] = useState<number[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [tempState, setTempState] = useState<Todo | null>(null);
+
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  useEffect(() => {
+    inputRef.current?.focus();
+
+    getTodos()
+      .then(todosFromServer => {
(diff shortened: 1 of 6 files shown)


SlawikH (approved):
Great job!

src/App.tsx
```diff
@@ -1,26 +1,195 @@
...
+  if (sortBy === 'Active') {
+    filteredTodos = activeTodo;
   }
 
+  if (sortBy === 'Completed') {
+    filteredTodos = completedTodo;
+  }
+  //   const completedTodo: Todo[] = todos.map(todo => {
+  //     if (todo.id === id) {
+  //       setLoadingIds([...loadingIds, todo.id]);
```
alexander-ignatow, line 38: Few issues here

1. Please move `type Sort` and sorting logic away from the component
2. Don't keep commented code
3. You are filtering in a bad way - your code will perform filtering (ln 24-25) in any case, regardless if it is needed or not. Please review your logic.

src/App.tsx
alexander-ignatow, line 1: [Optional fix] - this component looks a bit bloated. Consider breaking it into the smaller components

src/store/Errors.tsx
alexander-ignatow, line 1: Please consider renaming this file to make it look different from normal components
Say: ErrorsProvider, etc

src/store/Errors.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const ErrorsProvider: React.FC<Props> = ({ children }) => {
+  const [errorMessage, setErrorMessage] = useState('');
+
+  useEffect(() => {
+    if (!errorMessage) {
+      return;
+    }
+
+    const timeout = setTimeout(() => setErrorMessage(''), 3000);
```
alexander-ignatow, line 33: [Optional fix] minor, but can you please convert `3000` to some nicely named constant

src/store/Errors.tsx
```diff
@@ -0,0 +1,43 @@
...
+  ReactNode,
+  useEffect,
+  useState,
+  createContext,
+  useContext,
+} from 'react';
+
+type UErrorsContext = {
+  errorMessage: string;
+  setErrorMessage: React.Dispatch<React.SetStateAction<string>>;
```
alexander-ignatow, line 11: avoid exposing such specific types, unless you need it

```suggestion
  setErrorMessage: (value: string) => void;
```

alexander-ignatow (changes requested):
Good job! 

Lets make this code better, by slightly improving the structure

src/App.tsx
Follder: Hi! I think, we need filtering always apart from the situation when we don`t have any todos

src/App.tsx
Follder: Ok! I`ll make it shorter

src/store/Errors.tsx
Follder: Ok, thanks

src/store/Errors.tsx
Follder: Ok!

src/store/Errors.tsx
Follder: Don`t understand this moment.. the setState  function has React.Dispatch<React.SetStateAction<string>> type, but we can use (value: string) => void type for it?

mrkovaliv (approved):
👍 👍 👍 👍

PR #1278, opened 2024-05-25
task solved

ekateriina7:
https://ekateriina7.github.io/react_todo-app-with-api/

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,84 @@
+import React, { useCallback, useContext } from 'react';
+import { deleteTodo } from '../../api/todos';
+import { DispatchContext, StateContext } from '../../store/TodoContext';
+import { ActionTypes, FilterFields } from '../../store/types';
+
+export const Footer: React.FC = () => {
+  const dispatch = useContext(DispatchContext);
+  const { todos, filter } = useContext(StateContext);
+
+  const completedTodos = todos.filter(todo => todo.completed);
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const onChangeFilter = useCallback(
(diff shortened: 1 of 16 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,84 @@
...
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {activeTodos.length} items left
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          onClick={() => onChangeFilter(FilterFields.All)}
+          href="#/"
+          className={`filter__link ${filter === FilterFields.All ? 'selected' : ''}`}
```
volodymyr-soltys97, line 49: Use the classnames library for add class with condition, fix it everywhere

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
+      switch (filter) {
+        case FilterFields.Active:
+          return !todo.completed;
+        case FilterFields.Completed:
+          return todo.completed;
+        default:
+          return todo;
+      }
```
volodymyr-soltys97, line 23:
```suggestion
      switch (filter) {
        case FilterFields.Active:
          return !todo.completed;
          
        case FilterFields.Completed:
          return todo.completed;
          
        default:
          return todo;
      }
```

volodymyr-soltys97 (changes requested):
Great work 🔥 
Let's make your code better

vadiimvooo (approved):
Well done

PR #1279, opened 2024-05-25
solution

ABahatiy:
https://abahatiy.github.io/react_todo-app-with-api/

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -1,0 +1,80 @@
+import React, { Ref } from 'react';
+import cn from 'classnames';
+import { Todo } from '../types/Todo';
+import { updateTodo } from '../api/todos';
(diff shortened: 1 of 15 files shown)


src/components/Header.tsx
```diff
@@ -0,0 +1,80 @@
...
+  isLoading,
+  todos,
+  onError,
+  setLoadingTodoIds,
+}) => {
+  const isEveryTodoActive =
+    todos.every(todo => todo.completed) && !!todos.length;
+
+  const handleToggleAll = async () => {
+    const filteredTasks = todos.filter(todo => !todo.completed);
```
SanyaBratashchuk, line 34:
```suggestion
    const completedTodos = todos.filter(todo => !todo.completed);
```

SanyaBratashchuk (approved):
image

PR #1282, opened 2024-05-25
add solution

iAlfreX:
[DEMO LINK](https://iAlfreX.github.io/react_todo-app-with-api/)


Follder (approved):
Good job! =)

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,242 @@
...
+          )}
+
+          <form onSubmit={handleAddTodo}>
+            <input
+              data-cy="NewTodoField"
+              type="text"
+              className="todoapp__new-todo"
+              placeholder="What needs to be done?"
+              value={title}
+              onChange={event => setTitle(event.target.value.trimStart())}
```
undeletable, line 178: It's not a good idea to pass processed value directly into the input. For example: if user starts typing with a space and sees that nothing has changed, this would feel like the input doesn't work.
I'd recommend not to do this trim directly on change, and do the normalization directly before processing the value. In fact, you already do this in `handleAddTodo()`.

undeletable (changes requested):
Nice work! I just have a small notice about title text field behavior - please fix that.
I've also added a couple of non-critical recommendations. They're not required to be addressed for approval - these are rather advices for your further work. So please take a look.

iAlfreX:
Understand! Thanks for comments

PR #1285, opened 2024-05-27
Solution

lmmaarriiaa:
[DEMO LINK](https://lmmaarriiaa.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,221 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import { FC, useEffect, useRef, useState } from 'react';
+
+import { USER_ID, todoService } from './api/todos';
 import { UserWarning } from './UserWarning';
+import Footer from './components/Footer';
+import Header from './components/Header';
+import TodoList from './components/TodoList';
 
-const USER_ID = 0;
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
 
-export const App: React.FC = () => {
+// const USER_ID = 664;
+
+export const App: FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [selectedStatus, setSelectedStatus] = useState<Status>('all');
+  const [loadingTodosIds, setLoadingTodosIds] = useState<number[]>([]);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    inputRef.current?.focus();
+
+    todoService
+      .getTodos()
+      .then(setTodos)
+      .catch(() => setErrorMessage('Unable to load todos'));
+  }, []);
+
+  useEffect(() => {
+    if (!errorMessage) {
+      return;
+    }
+
+    const timeout = setTimeout(() => setErrorMessage(''), 3000);
+
+    return () => clearTimeout(timeout);
+  }, [errorMessage]);
+
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
+
+  let visibleTodos = todos;
+
+  if (selectedStatus === 'active') {
+    visibleTodos = activeTodos;
+  }
+
+  if (selectedStatus === 'completed') {
+    visibleTodos = completedTodos;
+  }
+
+  const addTodo = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const normalizedTitle = title.trim();
+
+    if (!normalizedTitle) {
+      setErrorMessage('Title should not be empty');
+
+      return;
+    }
+
+    if (inputRef.current) {
+      inputRef.current.disabled = true;
+    }
+
+    const newTodo = {
+      title: normalizedTitle,
+      userId: USER_ID,
+      completed: false,
+    };
+
+    setTempTodo({ ...newTodo, id: 0 });
+
+    todoService
+      .addTodo(newTodo)
+      .then((resTodo: Todo) => {
+        setTodos(currentTodos => [...currentTodos, resTodo]);
+        setTitle('');
+      })
+      .catch(() => {
+        setErrorMessage('Unable to add a todo');
+      })
+      .finally(() => {
+        if (inputRef.current) {
+          inputRef.current.disabled = false;
+        }
+
+        setTempTodo(null);
+        inputRef.current?.focus();
+      });
+  };
+
+  const deleteTodo = (todoId: number) => {
+    setLoadingTodosIds(current => [...current, todoId]);
+
+    todoService
+      .deleteTodo(todoId)
+      .then(() => {
+        setTodos(currentTodos =>
+          currentTodos.filter(todo => todo.id !== todoId),
+        );
+      })
+      .catch(() => {
+        setErrorMessage('Unable to delete a todo');
+      })
+      .finally(() => {
+        setLoadingTodosIds(current => current.filter(id => id !== todoId));
+        inputRef.current?.focus();
+      });
+  };
+
+  const deleteAllCompleted = () => {
+    completedTodos.forEach(todo => deleteTodo(todo.id));
+  };
+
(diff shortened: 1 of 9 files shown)


Olek-91 (approved):

src/App.tsx
```diff
@@ -1,26 +1,221 @@
...
+import { USER_ID, todoService } from './api/todos';
 import { UserWarning } from './UserWarning';
+import Footer from './components/Footer';
+import Header from './components/Header';
+import TodoList from './components/TodoList';
+
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
+
+// const USER_ID = 664;
```
DanilWeda, line 12: remove unused

src/App.tsx
```diff
@@ -1,26 +1,221 @@
...
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    inputRef.current?.focus();
+
+    todoService
+      .getTodos()
+      .then(setTodos)
+      .catch(() => setErrorMessage('Unable to load todos'));
```
DanilWeda, line 30: Please, move all errors in enum or constants

src/App.tsx
```diff
@@ -1,26 +1,221 @@
...
+        );
+      })
+      .catch(() => {
+        setErrorMessage('Unable to update a todo');
+        throw new Error();
+      });
+  };
+
+  const updateAllToggleStatus = () => {
+    if (activeTodos.length > 0) {
```
DanilWeda, line 158: just
```suggestion
    if (activeTodos.length ) {
```

src/App.tsx
```diff
@@ -1,26 +1,221 @@
...
+      activeTodos.forEach(updateToggleStatus);
+    } else {
+      completedTodos.forEach(updateToggleStatus);
+    }
+  };
+
+  const isAllToggleButtonVisible = activeTodos.length === 0;
```
DanilWeda, line 165:
```suggestion
  const isAllToggleButtonVisible = !activeTodos.length;
```
as alternative

src/App.tsx
```diff
@@ -1,26 +1,221 @@
...
+          selectedStatus={selectedStatus}
+          deleteTodo={deleteTodo}
+          onToggle={updateToggleStatus}
+          tempTodo={tempTodo}
+          loadingTodosIds={loadingTodosIds}
+          onRename={updateTodoName}
+          setLoadingTodosIds={setLoadingTodosIds}
+        />
+
+        {todos.length > 0 && (
```
DanilWeda, line 196:
```suggestion
        {!!todos.length  && (
```
as alternative

src/App.tsx
```diff
@@ -1,26 +1,221 @@
...
+          <Footer
+            activeTodosCount={activeTodos.length}
+            selectedStatus={selectedStatus}
+            setSelectedStatus={setSelectedStatus}
+            deleteAllCompleted={deleteAllCompleted}
+            clearAllVisible={completedTodos.length > 0}
```
DanilWeda, line 202:
```suggestion
            clearAllVisible={!!completedTodos.length}
```
as alternative

src/components/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeTodosCount} ${activeTodosCount === 1 ? 'item' : 'items'} left`}
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={cn('filter__link', { selected: selectedStatus === 'all' })}
+          data-cy="FilterLinkAll"
+          onClick={() => setSelectedStatus('all')}
```
DanilWeda, line 32: All status move to enums or constants and use in all places in your application

DanilWeda (changes requested):
Good! 
Check some small comments. 
Thanks)

src/components/Header.tsx
```diff
@@ -0,0 +1,46 @@
+import { forwardRef, Dispatch } from 'react';
+import cn from 'classnames';
+
+import { Todo } from '../types/Todo';
+
+interface Props {
+  todos: Todo[];
```
alexander-ignatow, line 7: Don't pass more information, than you need to the component
```suggestion
  hasTodos: boolean;
```

src/App.tsx
```diff
@@ -1,26 +1,220 @@
...
+import Header from './components/Header';
+import TodoList from './components/TodoList';
+
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
+import { Errors } from './types/Errors';
+
+export const App: FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
```
alexander-ignatow, line 15: [Optional fix] If you keep this state on the global level, then the whole app rerenders on each new character entered into the header

Please move it to the header

Anyways, you are using the title only inside `addTodo` func, so you can just make it an argument of the `addTodo`

src/components/Header.tsx
```diff
@@ -0,0 +1,46 @@
...
+  addTodo: (event: React.FormEvent<HTMLFormElement>) => void;
+  title: string;
+  setTitle: Dispatch<React.SetStateAction<string>>;
```
alexander-ignatow, line 10: title looks more like internal detail of the header - it makes sense to keep this state inside

src/components/Header.tsx
```diff
@@ -0,0 +1,46 @@
+import { forwardRef, Dispatch } from 'react';
+import cn from 'classnames';
+
+import { Todo } from '../types/Todo';
+
+interface Props {
+  todos: Todo[];
+  addTodo: (event: React.FormEvent<HTMLFormElement>) => void;
```
alexander-ignatow, line 8: You are exposing unnecessary inner details 

The much cleaner option is 
```suggestion
  addTodo: (title: string) => void;
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+import cn from 'classnames';
+
+import { Status } from '../types/Status';
+
+interface Props {
+  activeTodosCount: number;
+  selectedStatus: Status;
+  setSelectedStatus: (status: Status) => void;
+  deleteAllCompleted: () => void;
+  clearAllVisible: boolean;
```
alexander-ignatow, line 11: [optional fix]

Common convention is start names for `boolean` from "is", "are", "can", etc

```suggestion
  canClearAllVisible: boolean;
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+  onRename?: (todo: Todo) => Promise<void>;
+  setLoadingTodosIds?: Dispatch<React.SetStateAction<number[]>>;
+}
+const TodoItem: FC<Props> = ({
+  todo,
+  deleteTodo = () => {},
+  onToggle = () => {},
+  isLoading,
+  onRename = () => {},
+  setLoadingTodosIds = () => {},
```
alexander-ignatow, line 22: Please remove defaults - you don't really need them

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const trimmedNewTitle = newTitle.trim();
+
+    if (trimmedNewTitle === todo.title) {
+      setIsRenaming(false);
+
+      return;
+    }
+
+    if (!trimmedNewTitle) {
+      deleteTodo(todo.id);
```
alexander-ignatow, line 39: Remove defaults and just add null checks to the optional props
```suggestion
      deleteTodo?.(todo.id);
```

alexander-ignatow (changes requested):
Very good 👍 

Let's improve the structure of your code a bit with one more round

src/App.tsx
lmmaarriiaa: Hello! At first my function looked just like you say, but then I realized I need to accept the event to do event.preventDefault() in addTodo func, so changed everything. Can I do it without event.preventDefault?

lmmaarriiaa: Hello! I cannot move title into the header, because I need to setTitle to '' in my addTodo func. And I cannot move my addTodo func into the header because I need to setErrorMessage in it, and my errorMessage needs to be on top to be accessible everywhere> The better solution would be to make global context for the errorMessage, but here I didnt do it :( 
But I changed everything else that you said, thank you!

DanilWeda (approved):
Good job! 
Approved. 
Thanks)

PR #1290, opened 2024-05-27
Solution

OTsiolkovskiy:
[DEMO LINK](https://OTsiolkovskiy.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,218 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { FormEvent, useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  USER_ID,
+  creatTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+} from './api/todos';
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
+import { Error } from './types/Error';
 
-const USER_ID = 0;
+import { Header } from './conponents/Header/Header';
+import { TodoList } from './conponents/TodoList/TodoList';
+import { Footer } from './conponents/Footer/Footer';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState<Status>(Status.All);
+  const [error, setError] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodos, setLoadingTodos] = useState<number[]>([]);
+  const [newTitle, setNewTitle] = useState<string>('');
+
+  const titleField = useRef<HTMLInputElement>(null);
+
+  const completedTodos = todos.filter(todo => todo.completed);
+  const activeTodos = todos.filter(todo => !todo.completed);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => setError(Error.UnableLoad));
+  }, []);
+
+  useEffect(() => {
+    if (!error) {
+      return;
+    }
+
+    const timeout = setTimeout(() => setError(''), 3000);
+
+    return () => clearTimeout(timeout);
+  }, [error]);
+
+  useEffect(() => {
+    if (titleField.current) {
+      titleField.current?.focus();
+    }
+  }, [todos]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">
-          React Todo App - Add and Delete
-        </a>
-      </p>
+  const handleAddNewTitle = (event: FormEvent) => {
+    event.preventDefault();
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    const trimmedNewTitle = newTitle.trim();
+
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
+  const [filterBy, setFilterBy] = useState<Status>(Status.All);
+  const [error, setError] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodos, setLoadingTodos] = useState<number[]>([]);
+  const [newTitle, setNewTitle] = useState<string>('');
+
+  const titleField = useRef<HTMLInputElement>(null);
+
+  const completedTodos = todos.filter(todo => todo.completed);
+  const activeTodos = todos.filter(todo => !todo.completed);
```
alexander-ignatow, line 31: This is ineffective, as on each render you call the filter two times, regardless of whether you need it or not.

I suggest you write some function (say `filterTodos`) that will perform filtering only based on the currently selected filters - or not perform the filtering at all.

```suggestion
  const filteredTodos = filterTodos(todos, filterBy);
```

src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
+
+  useEffect(() => {
+    if (!error) {
+      return;
+    }
+
+    const timeout = setTimeout(() => setError(''), 3000);
+
+    return () => clearTimeout(timeout);
+  }, [error]);
```
alexander-ignatow, line 47: [no need to fix]
technically you have a bug with this - if two errors happen within 3000 ms, the second error will be displayed only for a fraction of the time.
To avoid this you should clear (or check and clear)  the timeout before setting the new error

But this is very unlikely in your app, so no need to fix it

src/conponents/Footer/Footer.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { Status } from '../../types/Status';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  selectedFilter: Status;
+  setSelectedFilter: (filterBy: Status) => void;
+  deleteAllCompleted: () => void;
+  activeTodos: Todo[];
+  completedTodos: Todo[];
```
alexander-ignatow, line 10: don't pass more information, than you need - this reduces the possibility of making an error

```suggestion
  hasActiveTodosCount: number;
  hasCompletedTodos: boolean;
```

src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState<Status>(Status.All);
+  const [error, setError] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodos, setLoadingTodos] = useState<number[]>([]);
+  const [newTitle, setNewTitle] = useState<string>('');
+
+  const titleField = useRef<HTMLInputElement>(null);
```
alexander-ignatow, line 28: These are the internals of the Header component - please move this state there
This would have several positive outcomes, the one is that as of now the whole app is re-rendered on each new character entered into the input

src/conponents/Header/Header.tsx
```diff
@@ -0,0 +1,48 @@
+import cn from 'classnames';
+import React, { FormEvent } from 'react';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todos: Todo[];
+  handleAddTodo: (event: FormEvent<HTMLFormElement>) => void;
```
alexander-ignatow, line 7: Why pass all event? This is the internal detail of the component.
Better variant is

```suggestion
  handleAddTodo: (todoText: string) => void;
```

src/conponents/Header/Header.tsx
```diff
@@ -0,0 +1,48 @@
...
+  titleField: React.RefObject<HTMLInputElement>;
+  title: string;
+  setTitle: React.Dispatch<React.SetStateAction<string>>;
```
alexander-ignatow, line 10: See the comment in the `App.tsx` - these should be the state of the Header component

src/conponents/Header/Header.tsx
```diff
@@ -0,0 +1,48 @@
+import cn from 'classnames';
+import React, { FormEvent } from 'react';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todos: Todo[];
```
alexander-ignatow, line 6: Please review your code, and pass only the information you need. For instance:

```suggestion
  hasTodos: boolean;
```

src/conponents/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Status } from '../../types/Status';
+import { Todo } from '../../types/Todo';
+import { filterTodos } from '../../utils/filteredTodos';
+import { TodoItem } from '../TodoItem/TodoItem';
+
+interface Props {
+  todos: Todo[];
+  tempTodo: Todo | null;
+  selectedFilter: Status;
+  handleDeleteTodo?: (id: number) => void;
```
alexander-ignatow, line 11: Do you really need to make this prop optional?

src/conponents/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,50 @@
...
+  loadingTodos: number[];
+  onToggle?: (todo: Todo) => void;
```
alexander-ignatow, line 13: Same here - why is this prop optional?

alexander-ignatow (changes requested):
Looks good!

Now let's improve the app structure, to make the code better and cleaner!

OTsiolkovskiy:
All fixed according to comments.

mrkovaliv (approved):
Nice work 👍

PR #1301, opened 2024-06-06
task done

Reaffith:
[DEMO LINK](https://Reaffith.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable react-hooks/exhaustive-deps */
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
 import { client } from './utils/fetchClient';
@@ -20,9 +20,10 @@ export const App: React.FC = () => {
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const timeoutRef = useRef<NodeJS.Timeout | null>(null);
   const inputRef = useRef<HTMLInputElement>(null);
-  const [todosToDelete, setTodosToDelete] = useState<number[]>([]);
+  const [todosIdsToDelete, setTodosToDelete] = useState<number[]>([]);
+  const [toggleAllLoaderIds, setToggleAllLoaderIds] = useState<number[]>([]);
 
-  const getFilteredTodos = () => {
+  const filteredTodos = useMemo(() => {
     switch (filter) {
       case Filter.Active:
         return todos.filter(todo => !todo.completed);
@@ -31,9 +32,7 @@ export const App: React.FC = () => {
       default:
         return todos;
     }
-  };
-
-  const todosToRender = getFilteredTodos();
+  }, [filter, todos]);
 
   useEffect(() => {
     const fetchTodos = async () => {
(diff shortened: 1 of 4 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix this failed test
You only need to send 1 request, but now 2 are sent
image
image

Reaffith:
some test failing, but when manually testing them, all seems to work properly
image
image

TarasHoliuk:
1. When I add Todo there are many requests in the network (you send a lot of requests on submit):
image

There is even the Error displayed below.

2. You should show loader for each todo that in processing when clicking `Toggle all` todo status

src/App.tsx
```diff
@@ -1,26 +1,118 @@
...
+import { Context } from './components/Context/Context';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState<Filter>(Filter.All);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null);
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [todosToDelete, setTodosToDelete] = useState<number[]>([]);
```
TarasHoliuk, line 23: not todos but todoIds in name - you store only ids here
Also, you should also show loader on status change

src/App.tsx
```diff
@@ -1,26 +1,118 @@
...
+
+  const getFilteredTodos = () => {
```
TarasHoliuk, line 25: it's better to use memo in such cases:

```suggestion
  const filteredTodos = useMemo(() => {
```
and add dependencies array below as second argument of useMemo

src/components/TodoInput/TodoInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+      return createTodo(newTodo)
+        .then(createdTodo => {
+          setTodos(prev => [...prev, createdTodo]);
+          setTempTodo(null);
+          setTitle('');
+        })
+        .catch(() => {
+          setErrorMessage('Unable to add a todo');
+          setTempTodo(null);
```
TarasHoliuk, line 32: do it in finally only

src/components/TodoInput/TodoInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+export const TodoInput: React.FC<Props> = ({ setTempTodo, inputRef }) => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [title, setTitle] = useState('');
+  const { todos, setTodos, setErrorMessage } = useContext(Context);
+
+  const addTodo = (newTodo: Omit<Todo, 'id'>) => {
+    setIsLoading(true);
+
+    if (newTodo.title !== '') {
```
TarasHoliuk, line 21: trim value
it's better to set isLoading true inside this if

src/components/TodoInput/TodoInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+            inputRef.current.focus();
+          }
+        });
+    } else {
+      setErrorMessage('Title should not be empty');
+    }
+
+    setIsLoading(false);
+
+    return null;
```
TarasHoliuk, line 46: redundant

src/components/TodoInput/TodoInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+    completed: false,
+  };
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    addTodo(todoToCreate);
+  };
+
+  const switchCompleted = () => {
+    const val = todos.filter(t => t.completed).length !== todos.length;
```
TarasHoliuk, line 61: what is val? Add meaningful name

src/components/TodoInput/TodoInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const copyOfTodos = val ? todos.filter(t => !t.completed) : [...todos];
+
+    copyOfTodos.map(todo => {
```
TarasHoliuk, line 64: you don't need to make copy of array - it's redundant

Also, map is wrong method here - use forEach

TarasHoliuk (changes requested):
Some functionality doesn't work as expected, so that is the reason tests fail. Ask for help in the chat if you can't fix them

SanyaBratashchuk (approved):
LGTM

PR #1302, opened 2024-06-07
Develop

SokolAlena:
[DEMO LINK](https://SokolAlena.github.io/react_todo-app-with-api/)


SanyaBratashchuk (changes requested):
handle plural it should be `1 item left` but `2 items left`
image

image

SokolAlena:
but it should be number of todos that are not completed, not all

SanyaBratashchuk (approved):

PR #1303, opened 2024-06-08
React Todo App With API

Denys209818:
[DEMO LINK](https://denys209818.github.io/react_todo-app-with-api)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,146 @@
 /* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback,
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,26 +1,146 @@
...
+          return todosItems.filter(todo => !todo.completed);
+        }
+
+        case FilterType.COMPLETED: {
+          return todosItems.filter(todo => todo.completed);
+        }
+      }
+    },
+    [filter],
+  );
```
SanyaBratashchuk, line 68: use use memo here for memorize a result of function not a function itself

SanyaBratashchuk (changes requested):
handle plural it should be `1 item` but `2 items`
image
2. empty todo should be deleted while editing

SanyaBratashchuk (approved):

PR #1309, opened 2024-06-15
third part of project - implementation

robsza1122:
https://robsza1122.github.io/react_todo-app-with-api/

### DIFF
--- a/.hintrc
+++ b/.hintrc
@@ -1,0 +1,19 @@
+{
+  "extends": [
+    "development"
+  ],
+  "hints": {
+    "axe/forms": [
+      "default",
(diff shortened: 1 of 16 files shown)


BogdanMaliuta (changes requested):
Cant see demo, also you didnt pass 1 test 
image

BogdanMaliuta (changes requested):
For now I know why test didn`t pass.
When you updating todo and have no success, for example no internet, your input should stay open, todo shouldn`t change and you need to show error message, but in your demo todo just changes and gets error message, you can test it with devtools using offline network mode

PR #1310, opened 2024-06-17
3/3 part of the todo-list task

ly-manka:
[DEMO LINK](https://ly-manka.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,471 @@
 /* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useRef, useState } from 'react';
+import cn from 'classnames';
 
-const USER_ID = 0;
+import * as todoService from './api/todos';
+import { Todo } from './types/Todo';
+import { LinkMode } from './types/LinkMode';
+import { EditingItem } from './types/EditingItem';
+import { CSSTransition, TransitionGroup } from 'react-transition-group';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,26 +1,471 @@
...
+  };
+
+  const [isSaving, setIsSaving] = useState(false);
+  const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
+  const [updatingTodoIds, setUpdatingTodoIds] = useState<number[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const timerId = useRef(0);
+
+  function focusInput() {
```
volodymyr-soltys97, line 87: You need to use 1 function creation type everywhere
```suggestion
  const focusInput = () => {
```

src/App.tsx
```diff
@@ -1,26 +1,471 @@
...
+                  todos.forEach(todo =>
+                    updateTodo({ ...todo, completed: !todo.completed }),
+                  );
+                } else {
+                  const activeTodos = todos.filter(todo => !todo.completed);
+
+                  activeTodos.forEach(todo =>
+                    updateTodo({ ...todo, completed: !todo.completed }),
+                  );
+                }
```
volodymyr-soltys97, line 272: Move this logic from jsx to the helper function and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better.
It's bad practice to write so much code in a single component.
You need to split the code into several components, for example TodoItem, TodoList, TodoFilter, Footer

volodymyr-soltys97 (approved):
Great work 🔥

PR #1314, opened 2024-06-19
3/3 todos

emberlast1:
[DEMO LINK](https://emberlast1.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,234 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import {
+  USER_ID,
+  addTodos,
+  deleteTodos,
+  getTodos,
+  updateTodos,
+} from './api/todos';
+import { Header } from './Components/Header/Header';
+import { Todo } from './types/Todo';
+import { Footer } from './Components/Footer/Footer';
+import { TodoList } from './Components/TodoList/TodoList';
+import { Errors } from './Components/Errors/Errors';
+import { Status } from './types/Status';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState(Status.All);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [inputDisabled, setInputDisabled] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState<number[]>([]);
+  const [inputTitle, setInputTitle] = useState('');
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputTitle(event.target.value);
+  };
+
+  const itemsLeft = todos.filter(todo => !todo.completed).length;
+
+  const clearAllButtonActive = todos.length - itemsLeft;
+
+  function onAdd({ userId, completed }: Todo) {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,234 @@
...
+  const [isLoading, setIsLoading] = useState<number[]>([]);
+  const [inputTitle, setInputTitle] = useState('');
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputTitle(event.target.value);
+  };
+
+  const itemsLeft = todos.filter(todo => !todo.completed).length;
+
+  const clearAllButtonActive = todos.length - itemsLeft;
```
TarasHoliuk, line 34:
```suggestion
  const isClearAllButtonActive = todos.length - itemsLeft;
```

src/Components/Footer/Footer.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
+import cn from 'classnames';
+import { Status } from '../../types/Status';
+
+type Props = {
+  status: Status;
+  onClick: (status: Status) => void;
+  itemsLeft: number;
+  clearAllButtonActive: number;
```
TarasHoliuk, line 9: it's a bit strange that it's a number
Let's pass a boolean value

src/App.tsx
```diff
@@ -1,26 +1,234 @@
...
+import { Errors } from './Components/Errors/Errors';
+import { Status } from './types/Status';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState(Status.All);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [inputDisabled, setInputDisabled] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState<number[]>([]);
```
TarasHoliuk, line 25:
```suggestion
  const [loadingTodoIds, setIsLoading] = useState<number[]>([]);
```
or `processingTodoIds`

isLoading sounds like boolean value which is not true in this case

src/App.tsx
```diff
@@ -1,26 +1,234 @@
...
+        setTodos(currentTodos => [...currentTodos, newTodo]);
+        setInputTitle('');
+      })
+      .catch(error => {
+        setErrorMessage('Unable to add a todo');
+        setInputDisabled(false);
+        setTimeout(() => {
+          setErrorMessage('');
+        }, 3000);
+        throw error;
```
TarasHoliuk, line 68: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

src/Components/Header/Header.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }, [inputDisabled]);
+
+  const handleOnSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    setError('');
+    setInputDisabled();
+
+    onSubmit({
+      id: 0,
+      userId: 762,
```
TarasHoliuk, line 42: import USER_ID and use here

src/Components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,48 @@
...
+            toggleTodo={toggleTodo}
+            updateTodo={updateTodo}
+          />
+        ))}
+
+      {tempTodo && (
+        <TodoItem
+          todo={tempTodo}
+          onDelete={() => {}}
+          isLoading={isLoading.includes(0)}
```
TarasHoliuk, line 41:
```suggestion
          isLoading
```
temp todo always should be displayed with loader
So we also can skip adding 0 to `isLoading` array in the App.tsx

TarasHoliuk (approved):
Amazing 🔥 

A few minor comments to make it perfect 🙂

PR #1316, opened 2024-06-25
create-project

Prokopovych16:


src/App.tsx
```diff
@@ -1,26 +1,249 @@
...
+import { deleteTodos, getTodos, patchTodos, postTodos } from './api/todos';
 
-const USER_ID = 0;
+const USER_ID = 700;
+
+enum Filter {
+  All = 'all',
+  Active = 'active',
+  Completed = 'completed',
+}
```
GUSILLUS, line 18: Consider creating ts file with types and move all types in this file

src/utils/Footer/Footer.tsx
```diff
@@ -0,0 +1,61 @@
...
+  notCompletedTodos,
+  filter,
+  setFilter,
+  handleClearCompleted,
+}) => {
+  const filterLinks = [
+    { key: Filter.All, label: 'All', href: '#/' },
+    { key: Filter.Active, label: 'Active', href: '#/active' },
+    { key: Filter.Completed, label: 'Completed', href: '#/completed' },
+  ];
```
GUSILLUS, line 28: Consider changing it with Enum or move it to out from component better to constant file

src/App.tsx
```diff
@@ -1,26 +1,249 @@
...
+        />
+
+        <TodoList
+          visibleTodos={visibleTodos}
+          handleDeleteTodo={handleDeleteTodo}
+          handleTodoStatusChange={handleTodoStatusChange}
+          loadingTodoId={loadingTodoId}
+          tempTodo={tempTodo}
+          updateTodo={updateTodo}
+        />
```
GUSILLUS, line 230: If we don't have any todos? We still show todo list?

GUSILLUS (changes requested):
GJ! Try to pass your last test
image

Prokopovych16:
[DEMO LINK](https://Prokopovych16.github.io/react_todo-app-with-api/)

src/App.tsx
```diff
@@ -1,26 +1,272 @@
...
+    todos?.length > 0 && todos?.every(todo => todo.completed);
+
+  // #region get
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => setError('Unable to load todos'));
+  }, []);
+
+  const getVisibleTodos = (generalTodos: Todo[], generalFilter: string) => {
```
TarasHoliuk, line 41: this function doesn't use any value from the scope, so you can move it out of component
Moreover, you can move it to a separate file like `helpers.ts`, etc.

src/App.tsx
```diff
@@ -1,26 +1,272 @@
...
+    let filteredTodos = generalTodos;
+
+    if (generalFilter === Filter.Active) {
+      filteredTodos = filteredTodos.filter(todo => !todo.completed);
+    } else if (generalFilter === Filter.Completed) {
+      filteredTodos = filteredTodos.filter(todo => todo.completed);
+    }
```
TarasHoliuk, line 48:
```suggestion
    if (generalFilter === Filter.Active) {
      return  generalTodos.filter(todo => !todo.completed);
    }

    if (generalFilter === Filter.Completed) {
      return generalTodos.filter(todo => todo.completed);
    }
```
you can immediately return and get rid of additional variable and `else if` (separated if blocks are easier to read)

src/App.tsx
```diff
@@ -1,26 +1,272 @@
...
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string>('');
+  const [filter, setFilter] = useState<string>(Filter.All);
+  const [newTodoTitle, setNewTodoTitle] = useState<string>('');
+  const [allCompleted, setAllCompleted] = useState<boolean>(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  // const [loadingTodoId, setLoadingTodoId] = useState<number | null>(null);
```
TarasHoliuk, line 27: remove comments

src/App.tsx
```diff
@@ -1,26 +1,272 @@
...
+
+  //#region patchStatus
+  const handleToggleAllCompleted = () => {
+    const arrayWithNotCompletedTodoIds = visibleTodos
+      .filter(todo => !todo.completed)
+      .map(todo => todo.id);
+
+    const arrayWithCompletedTodoIds = visibleTodos
+      .filter(todo => todo.completed)
+      .map(todo => todo.id);
```
TarasHoliuk, line 67: one of filtering is redundant - do it when it's necessary

src/App.tsx
```diff
@@ -1,26 +1,272 @@
...
+      completed: !allCompleted,
+    }));
+
+    if (allCompleted) {
+      setLoadingTodoIds(arrayWithCompletedTodoIds);
+    } else {
+      setLoadingTodoIds(arrayWithNotCompletedTodoIds);
+    }
+
+    Promise.all(
```
TarasHoliuk, line 80: Generally using Promise.all in such cases might be ok
But to pass tests just iterate through all todo ids you need to delete (forEach) and send requests separately. In this case it'll allow to handle each erroneous result separately

src/App.tsx
```diff
@@ -1,26 +1,272 @@
...
+        setTodos(currentTodos => currentTodos.filter(todo => todo.id !== id));
+      })
+      .catch(() => setError('Unable to delete a todo'))
+      .finally(() => {
+        setLoadingTodoIds([]);
+        setTempTodo(null);
+      });
+  };
+
+  // eslint-disable-next-line max-len, prettier/prettier
```
TarasHoliuk, line 160: try to get rid of error w/o disabling lint

src/App.tsx
```diff
@@ -1,26 +1,272 @@
...
+
+  const handleClearCompleted = () => {
+    const modifiedTodos = todos.filter(todo => todo.completed === true);
```
TarasHoliuk, line 163:
```suggestion
    const modifiedTodos = todos.filter(todo => todo.completed);
```

src/utils/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export const TodoItem: React.FC<ItemProps> = ({
+  todo,
+  handleDeleteTodo,
+  handleTodoStatusChange,
+  loadingTodoIds,
+  updateTodo,
+}) => {
+  const [editedTodo, setEditedTodo] = useState<string>('');
+  const [isDoubleClicked, setIsDoubleClicked] = useState<number | null>(null);
```
TarasHoliuk, line 13: it can be just a boolean value

src/utils/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,64 @@
...
+          todo={todo}
+          handleTodoStatusChange={handleTodoStatusChange}
+          handleDeleteTodo={handleDeleteTodo}
+          loadingTodoIds={loadingTodoIds}
+          updateTodo={updateTodo}
+        />
+      ))}
+
+      {tempTodo && (
+        <div
```
TarasHoliuk, line 28: you duplicate part of code here
You should just adjust your `TodoItem` component to be able to use it for tempTodo

Prokopovych16: it was the idea of my mentor to make it in this way.

etojeDenys: hm, sounds strange, can you explain the mentor's idea?

src/utils/TodoItem/TodoItem.tsx
etojeDenys: image
still not fixed

src/utils/Footer/Footer.tsx
etojeDenys: it's a constant, so better to move it out of the component
image

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+import { deleteTodos, getTodos, patchTodos, postTodos } from './api/todos';
+import { Filter, getVisibleTodos } from './helpers';
 
-const USER_ID = 0;
+const USER_ID = 700;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string>('');
+  const [filter, setFilter] = useState<string>(Filter.All);
```
etojeDenys, line 18:
```suggestion
  const [filter, setFilter] = useState<Filter>(Filter.All);
```

cypress/integration/page.spec.js
```diff
@@ -1249,7 +1249,7 @@ describe('', () => {
           // if before each pass
         });
 
-        it('should make all todos active', () => {
+        it.only('should make all todos active', () => {
```
etojeDenys, line 1252:
```suggestion
        it('should make all todos active', () => {
```
don't forget to remove it

etojeDenys (changes requested):

src/utils/TodoList/TodoList.tsx
Prokopovych16: I don't remember, what was the purpose of this feature, but it 100% not my idea.

etojeDenys (approved):

PR #1329, opened 2024-07-09
add task solution

MakhmudKerimov:
[DEMO LINK](https://MakhmudKerimov.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,141 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import * as todoApi from './api/todos';
+import { TodoList } from './components/TodoList';
+import { TodoFilter } from './components/TodoFilter';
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,141 @@
...
+        setTempTodo(null);
+      })
+      .catch(error => {
+        setTempTodo(null);
+        handleErrorMessage('Unable to add a todo');
+        throw error;
+      });
+  }
+
+  function deleteTodo(id: number) {
```
volodymyr-soltys97, line 66: It is important to follow the same way of creating functions everywhere. Fix it everywhere
```suggestion
  const deleteTodo = (id: number) => {
```

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: status === Status.completed,
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setStatus(Status.completed)}
+        >
+          Completed
+        </a>
```
volodymyr-soltys97, line 70: Use `Object.value(Status)` and `map()` method for rendering filters

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

anastasiiavorobiova (approved):
Great work!

PR #1331, opened 2024-07-10
[Feature] added edit flow

Arthur-Tolkachov:
[DEMO LINK](https://arthur-tolkachov.github.io/react_todo-app-with-api/)


TarasHoliuk (approved):
You did a great job! Logic is separated and structured

One suggestion (just FYI) about files/directories structure: 
1. You have DeleteTodo and CompleteTodo folders with hooks only. If folder name is in PasclCase it should contain component
2. You moved all logic to hooks and in components you have only JSX. It's an interesting approach, and can be at least discussed. But usually, we move some reusable logic in hooks, or at least separate parts of logic if component is too big

PR #1332, opened 2024-07-11
implement toggling and renaming todo

SrTrace:
[DEMO LINK](https://srtrace.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,86 @@
+import React from 'react';
+import { SelectedStatus, Todo } from '../../types/Todo';
+import cn from 'classnames';
+import { deleteTodo } from '../../api/todos';
+
+type FooterProps = {
+  todos: Todo[];
+  selectedStatus: string;
+  setStatus: (e: React.MouseEvent<HTMLElement>) => void;
+  onErrorMessage: (error: string) => void;
+  onDeleteCompletedTodo: (todos: Todo[]) => void;
+  numOfCompletedTodos: number;
+  numOfActiveTodos: number;
+};
+
+export const Footer: React.FC<FooterProps> = ({
+  todos,
+  selectedStatus,
+  setStatus,
+  onDeleteCompletedTodo,
+  onErrorMessage,
+  numOfActiveTodos,
+  numOfCompletedTodos,
+}) => {
+  function handleDeleteAllCompletedTodos() {
+    const completedTodos = todos.filter(todo => todo.completed);
+    const deletePromises = completedTodos.map(todo =>
+      deleteTodo(todo.id).then(() => todo),
+    );
+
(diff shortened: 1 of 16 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,86 @@
...
+          onDeleteCompletedTodo(successfullyDeletedTodos);
+        }
+      });
+    });
+  }
+
+  return (
+    <>
+      {/* Hide the footer if there are no todos */}
+      {todos.length !== 0 && (
```
volodymyr-soltys97, line 49:
```suggestion
      {!!todos.length && (
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  if (status === SelectedStatus.all) {
+    filteredTodos = todos;
+  } else if (status === SelectedStatus.active) {
+    filteredTodos = todos.filter(todo => !todo.completed);
+  } else if (status === SelectedStatus.completed) {
+    filteredTodos = todos.filter(todo => todo.completed);
+  }
+
+  if (filteredTodos.length === 0) {
```
volodymyr-soltys97, line 31:
```suggestion
  if (!filteredTodos.length) {
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,57 @@
...
+}) => {
+  let filteredTodos: Todo[] = [];
+
+  if (status === SelectedStatus.all) {
+    filteredTodos = todos;
+  } else if (status === SelectedStatus.active) {
+    filteredTodos = todos.filter(todo => !todo.completed);
+  } else if (status === SelectedStatus.completed) {
+    filteredTodos = todos.filter(todo => todo.completed);
+  }
```
volodymyr-soltys97, line 29: Use a switch case here instead of if / else if, it will be better

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,182 @@
...
+          setIsEditing(false);
+        }
+      })
+      .catch(() => {
+        onErrorMessage('Unable to delete a todo');
+      })
+      .finally(() => setIsLoading(false));
+  }
+
+  function handleChangeTodo(todoId: number, newTodoProp: string) {
```
volodymyr-soltys97, line 46: Consider will be better if you use function expression everywhere instead of function declaration, fix it everywhere
```suggestion
  const handleChangeTodo = (todoId: number, newTodoProp: string) => {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Great work 🔥

PR #1345, opened 2024-07-14
add solutin task react_todo-app-with-api

seriuksergii:
[DEMO LINK](https://seriuksergii.github.io/react_todo-app-with-api/)


src/types/errors.ts
artemforsoff, line 1: enam is used if we use it as both a type and a value
so in this case we can use just as a object like this:
```
export const Errors = {
  DEFAULT: '',
  LOAD: 'Unable to load todos',
  TITLE: 'Title should not be empty',
  ADD: 'Unable to add a todo',
  DELETE: 'Unable to delete a todo',
  UPDATE: 'Unable to update a todo',
}

File: src/components/Footer/Footer.tsx [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+  activeTodosCount,
+  filter: filter,
+  setFilter: setFilter,
+  onClear,
+  canClearAllVisible,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeTodosCount} ${pluralize('item', activeTodosCount)} left`}
```
artemforsoff, line 25: you can remove unnecessary quotes
```
<span className="todo-count" data-cy="TodosCounter">
  {activeTodosCount} {pluralize('item', activeTodosCount)} left
</span>
```

artemforsoff (approved):
overall great beautiful work 🔥
please pay attention to my comments and recommend using them in the future

PR #1348, opened 2024-07-15
Develop

alina-kabanets:
https://alina-kabanets.github.io/react_todo-app-with-api/


src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+        handleError(ErrorType.ADD_TODO);
+      })
+      .finally(() => {
+        dispatch({ type: Type.setIsSubmitting, payload: false });
+        dispatch({ type: Type.setTempTodo, payload: null });
+      });
+  };
+
+  useEffect(() => {
+    if (inputRef.current && !editingId && !updatingId) {
```
TarasHoliuk, line 40: only one input can be in focus at the time. editingId is defined only when editing input exists (and it exists only when it's in focus).  So there is no need to check it here. I.e. it could be just the local state inside the TodoItem component.

Not critical, just FYI 🙂

TarasHoliuk (approved):
Amazing! You did great job 🔥 

Added 1 comment just FYI)

src/components/TodoForm/TodoForm.tsx
alina-kabanets: Thank you very much for the review and comments! Just checked although - it doesn't work without editingId 🤷‍♀️

PR #1366, opened 2024-07-22
solution 3

akryto:
https://akryto.github.io/react_todo-app-with-api/
Implement the ability to edit a todo title on double click:

- show the edit form instead of the title and remove button;
- cancel editing on `Esс` key `keyup` event;
- if the title was changed show the loader while waiting for the API response;
- update the todo title on success;
- show `Unable to update a todo` in case of API error;
- or the deletion error message if we tried to delete the todo.


DSB-18 (approved):
Good job!
But usage of console.log on deployment is a bad practice as I know.
Plus some extra comments and hints (like disabling max-len) were found 😉

PR #1370, opened 2024-07-24
xxxx

Teslaanazuma:
[DEMO LINK](https://Teslaanazuma.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,161 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useEffect, useMemo, useRef, useState } from 'react';
+import { UserWarning } from './components/UserWarning';
+import { TodoList } from './components/TodoList';
+import { TodoFilter } from './components/TodoFilter';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Header } from './components/Header';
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,26 +1,161 @@
...
+      .then(todo => {
+        setTodos(prevTodos =>
+          prevTodos.map(prevTodo =>
+            prevTodo.id === todo.id ? updatedTodo : prevTodo,
+          ),
+        );
+      })
+      .catch(() => {
+        setErrorMessage(ErrorMessages.UPDATE_TODO);
+        throw new Error();
```
TarasHoliuk, line 93: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

src/components/Header.tsx
```diff
@@ -0,0 +1,88 @@
...
+          setTempTodo(null);
+        });
+    }
+  };
+
+  return (
+    <header className="todoapp__header">
+      {isTodo && (
+        <button
+          onClick={() => onChangeTodoStatus()}
```
TarasHoliuk, line 67:
```suggestion
          onClick={onChangeTodoStatus}
```
function wrapper is redundant in such cases (when you don't pass some argument from the scope)

TarasHoliuk (approved):
Looks and works nice 🔥 I see improvements in code comparing with the first task - GJ 👍 
Added a few small comments just FYI

PR #1373, opened 2024-07-25
My solution

Karanelus:
[DEMO LINK](https://karanelus.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Error/TodoError.tsx
+++ b/src/components/Error/TodoError.tsx
@@ -1,0 +1,33 @@
+import classNames from 'classnames';
+import { useAppContextContainer } from '../../context/AppContext';
+
+type Props = {
+  error: string | null;
+};
+
+const TodoError = ({ error }: Props) => {
+  const { handleClickCloseError } = useAppContextContainer();
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={classNames(
+        'notification',
+        'is-danger',
+        'is-light',
+        'has-text-weight-normal',
+        { hidden: error === null },
+      )}
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={handleClickCloseError}
+      />
+      {error}
+    </div>
+  );
+};
+
+export default TodoError;
--- a/src/components/Footer/TodoFooter.tsx
+++ b/src/components/Footer/TodoFooter.tsx
@@ -1,0 +1,45 @@
+/* eslint-disable no-param-reassign */
+import { useAppContextContainer } from '../../context/AppContext';
+import { footerOption } from './TodoFooter.data';
+import TodoFooterOption from './TodoFooterOption';
+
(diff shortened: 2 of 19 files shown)


Karanelus:
It's not completed yet, but it's take so much time, so i have a questions:
1) Why my form after dblClicking the input after type one symbol is rerendered?
2) How set load on every todo after click toggleAll?

Zibi95 (changes requested):
> 1. Why my form after dblClicking the input after type one symbol is rerendered?
> 2. How set load on every todo after click toggleAll?

1. In the change event, you are directly updating the list of all todos. You have an effect that depends on the todos, which triggers the action of focusing on the input in the TodoHeader. On blur, you submit the edited value... To implement the described functionality, you need to first store the edited todo title value in the local state and then update the list of todos upon submission.

2. Right now you have state which keeps one id of todos that will be in loading state. You can just extend this further to keep list (set would be great here) of ids that will be in loading state.

Karanelus:
Seems like it's work, but toggle is giving mistake all the time, even when error is catching =(

src/components/Error/TodoError.tsx
```diff
@@ -0,0 +1,33 @@
+import classNames from 'classnames';
+import { useAppContextContainer } from '../../context/AppContext';
+
+type Props = {
+  error: string | null;
```
BogdanMaliuta, line 5: better to use empty string when you have no error, not null

src/components/Footer/TodoFooter.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const completed = todos?.reduce((a, b) => (!b.completed ? (a += 1) : a), 0);
+  const isCompleted = todos?.some(el => el.completed);
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {completed} items left
+      </span>
+
+      {/* Active link should have the 'selected' class */}
```
BogdanMaliuta, line 19: remove comments from your code

src/components/Footer/TodoFooter.tsx
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable no-param-reassign */
+import { useAppContextContainer } from '../../context/AppContext';
+import { footerOption } from './TodoFooter.data';
+import TodoFooterOption from './TodoFooterOption';
+
+const TodoFooter = () => {
+  const { changeFilterType, filterType, todos, deleteCompleted } =
```
BogdanMaliuta, line 7: when here more than 3 values, split it into rows

BogdanMaliuta (approved):
nice work!
Just remove comments and make some minor 
fixes

src/components/Footer/TodoFooter.tsx
Karanelus: Pretier is not give me do that =)

PR #1390, opened 2024-08-08
Develop

jdifek:
[DEMO LINK](https://jdifek.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,259 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useEffect, useMemo, useRef, useState } from 'react';
+import { TodoList } from './components/TodoList';
+import { addTodo, deleteTodo, getTodos, updatedTodo } from './api/todos';
+import { Todo } from './types/Todo';
+import cn from 'classnames';
+import { Filter } from './types/types';
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [inputText, setInputText] = useState('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodos, setTempTodos] = useState<Todo[]>([]);
+  const [loadingError, setLoadingError] = useState('');
+  const [validationError, setValidationError] = useState('');
+  const [loadingTodos, setLoadingTodos] = useState<number[]>([]);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [filterSelected, setFilterSelected] = useState<Filter>(Filter.All);
+
+  const filterTodos = useMemo(() => {
+    switch (filterSelected) {
+      case Filter.Active:
+        return todos.filter(todo => !todo.completed);
+
+      case Filter.Completed:
+        return todos.filter(todo => todo.completed);
+
+      default:
+        return todos;
+    }
+  }, [todos, filterSelected]);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setLoadingError('Unable to load todos');
+        setTimeout(() => {
+          setLoadingError('');
+        }, 3000);
+      });
+  }, []);
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [todos, loadingTodos]);
+
+  const toggleTodo = (id: number) => {
+    setTodos(prevTodos =>
+      prevTodos.map(todo =>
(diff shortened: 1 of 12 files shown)


src/components/Todo.tsx
```diff
@@ -0,0 +1,122 @@
...
+            setFormTodo(false);
+          }}
+        >
+          <input
+            className="todo__title-field"
+            type="text"
+            value={inputText}
+            onChange={handleInputChange}
+            ref={refInputUpdate}
+            onBlur={() => {
```
TarasHoliuk, line 81: on blur, you should also send a request (behavior should be the same as on submit):
image

src/components/Todo.tsx
```diff
@@ -0,0 +1,122 @@
...
+            }}
+            placeholder="Empty todo will be deleted"
+          />
+        </form>
+      ) : (
+        <>
+          <span
+            data-cy="TodoTitle"
+            className="todo__title"
+            onDoubleClick={e => {
```
TarasHoliuk, line 92: Do you really need e.preventDefault() here?
Move to a separate handler function and just use it here

src/App.tsx
```diff
@@ -1,26 +1,259 @@
...
+      }
+
+      return;
+    }
+
+    const tempTodo: Todo = {
+      id: Date.now(),
+      title: inputText,
+      completed: false,
+      userId: 1129,
```
TarasHoliuk, line 85: import USER_ID and use here

src/App.tsx
```diff
@@ -1,26 +1,259 @@
...
+    };
+
+    setTempTodos(prevTempTodos => [...prevTempTodos, tempTodo]);
```
TarasHoliuk, line 88: There can be only one tempTodo at a time, do you really need an array here?

src/App.tsx
```diff
@@ -1,26 +1,259 @@
...
+        setLoadingError('Unable to add a todo');
+        setTempTodos(prevTempTodos =>
+          prevTempTodos.filter(todo => todo.id !== tempTodo.id),
+        );
+        setTimeout(() => {
+          setLoadingError('');
+        }, 3000);
+      })
+      .finally(() => {
+        setLoadingTodos(prev => {
```
TarasHoliuk, line 117: Since only one action can happen at a time, you can simplify it and just set empty array here

src/App.tsx
```diff
@@ -1,26 +1,259 @@
...
+  };
+
+  const leftItem = todos.filter(todo => !todo.completed).length;
+  const disabledButton = todos.every(todo => !todo.completed);
+
+  const filterValues = Object.values(Filter);
+
+  const allTodos = [...filterTodos, ...tempTodos];
+
+  const cnButton = todos.every(todo => todo.completed);
```
TarasHoliuk, line 192:
```suggestion
  const areAllTodosCompleted = todos.every(todo => todo.completed);
```

src/App.tsx
```diff
@@ -1,26 +1,259 @@
...
+  };
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setInputText(e.target.value);
+  };
+
+  const leftItem = todos.filter(todo => !todo.completed).length;
+  const disabledButton = todos.every(todo => !todo.completed);
+
+  const filterValues = Object.values(Filter);
```
TarasHoliuk, line 188: you can create this variable outside of the component (it's not dynamic)

src/App.tsx
```diff
@@ -1,26 +1,259 @@
...
+          todos={allTodos}
+          toggleTodo={toggleTodo}
+          deletePost={deletePost}
+          loadingTodos={loadingTodos}
+          updatedPost={updatedPost}
+        />
+        {todos.length > 0 && (
+          <Footer
+            leftItem={leftItem}
+            filterValues={filterValues}
```
TarasHoliuk, line 228: it's static value - no need to pass it as prop
just import Filter enum in the Footer componet

TarasHoliuk (changes requested):
GJ 👍 
let's improve your code a bit

Also, try to pass all tests

src/App.tsx
SanyaBratashchuk: image

src/App.tsx
SanyaBratashchuk: Create a const for userId and use it here

src/App.tsx
```diff
@@ -1,26 +1,299 @@
...
+        setTimeout(() => {
+          setLoadingError('');
+        }, 3000);
+      })
+      .finally(() => {
+        setLoadingTodos(prev => {
+          const updated = prev.filter(id => id !== tempTodo.id);
+
+          return updated;
+        });
```
SanyaBratashchuk, line 157:
```suggestion
        setLoadingTodos([]);
```

SanyaBratashchuk (changes requested):
image
comments left by previous reviewer is not fixed. If you have any questions feel free to ask in chat

SanyaBratashchuk (approved):

PR #1393, opened 2024-08-12
Solution

opalahecha:
[DEMO LINK](https://opalahecha.github.io/react_todo-app-with-api/)


etojeDenys (changes requested):
1. we don't need to make requests when title doesn't change

https://github.com/user-attachments/assets/bac94c4c-c2d3-4af8-9c48-62c7b5c34133

2. after click on clear all completed button, let's show loaders on todos that we need to update

VitaliyBondarenko1982 (approved):
Looks good.

PR #1394, opened 2024-08-13
add solution

DSB-18:
[DEMO LINK](https://dsb-18.github.io/react_todo-app-with-api/)


src/components/Header.tsx
```diff
@@ -0,0 +1,73 @@
...
+    } else {
+      setIsSubmitting(true);
+      onAddTodo(title)
+        .then(() => {
+          setNewTitle('');
+          setIsSubmitting(false);
+        })
+        .catch(() => {
+          setError(ErrorMessage.add);
+          setIsSubmitting(false);
```
artemforsoff, line 44: `setIsSubmitting(false` called in then and catch
will be better move to finally block of promise like this:
```
onAddTodo(title)
  .then(() => {
    setNewTitle('');
  })
  .catch(() => {
    setError(ErrorMessage.add);
  })
  .finally(() => {
    setIsSubmitting(false);
  });
```

artemforsoff (approved):
overall great beautiful work 🔥

PR #1396, opened 2024-08-14
Develop

smachylooo:
[DEMO LINK](https://smachylooo.github.io/react_todo-app-with-api/)


SanyaBratashchuk (approved):
handle plural it should be `2 items` but `1 item`
image

PR #1407, opened 2024-08-23
added Todo app with full functionality

OlehPhw:


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,239 @@
...
+            {editingTodo?.id !== id ? (
+              <>
+                <span
+                  data-cy="TodoTitle"
+                  className="todo__title"
+                  onDoubleClick={() =>
+                    handleOnDoubleClick({ id, userId, title, completed })
+                  }
+                >
+                  {title}
```
etojeDenys, line 196: I think you can reuse TodoItem component here too

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,239 @@
...
+      await deleteTodo(todoForDelate.id);
+      setTodos(prevTodos =>
+        prevTodos.filter(todo => todoForDelate.id !== todo.id),
+      );
+    } catch (error) {
+      setErrorMessage('Unable to delete a todo');
+    } finally {
+      setLoadingTodosId([]);
+    }
+  };
```
etojeDenys, line 67: why do you need to pass here setLoadingTodosId, setTodos, todos and etc. to this component, if you can create this handleOnClickDelete function in the App.tsx and just pass this function as a prop

etojeDenys: try to fix it everywhere

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #1433, opened 2024-09-05
add task solution

valentin19939:
[DEMO LINK](https://valentin19939.github.io/react_todo-app-with-api/)


SanyaBratashchuk (approved):
image

PR #1456, opened 2024-09-30
test 1.0

manch0ffline:
[DEMO LINK](https://manch0ffline.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -9,7 +9,9 @@ type Props = {
   errorFunction: (el: string) => void;
   deletingFunction: (el: boolean) => void;
   deletingListId: number[];
-  todosFunction: (el: Todo[]) => void;
+  todosFunction: (
+    todosOrCallback: Todo[] | ((todos: Todo[]) => Todo[]),
+  ) => void;
   todos: Todo[];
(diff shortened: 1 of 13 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,73 @@
...
+  focusInput,
+  activateTodosId,
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {sortedArray.map((todo: Todo) => (
+        <TodoItem
+          key={todo.id}
+          todo={todo}
+          todosFunction={todosFunction}
```
etojeDenys: how can some other developer understand what todosFunction will do, better to rename it

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,73 @@
...
+          errorFunction={errorFunction}
+          deletingFunction={deletingFunction}
```
etojeDenys:
```suggestion
          onDelete={deletingFunction}
```

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #1459, opened 2024-10-03
add task solution

siefimov:
[DEMO LINK](https://siefimov.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Todos.tsx
+++ b/src/components/Todos.tsx
@@ -1,0 +1,97 @@
+import React, { useState } from 'react';
+import { Todo } from '../types/Todo';
+import { TodoFilter } from '../enums/TodoFilters.enum';
+import { Header } from './Header';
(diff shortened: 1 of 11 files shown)


src/components/Todos.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const filteredTodos = todos.filter(todo => {
+    if (filter === TodoFilter.Active) {
+      return !todo.completed;
+    }
+
+    if (filter === TodoFilter.Completed) {
+      return todo.completed;
+    }
+
+    return true;
```
SanyaBratashchuk: use switch case here and move filter inside conditions to avoid extra iteration when no filters are selected

SanyaBratashchuk (changes requested):
image

anastasiiavorobiova (approved):
Great work!

PR #1463, opened 2024-10-07
solution

pasha1932:
It`s not finished solution, becouse test don`t pass and i don't know why, can explain you explain?
image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,227 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { FormEvent, useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  addTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+  USER_ID,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+// eslint-disable-next-line max-len
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
 
-const USER_ID = 0;
+export enum Status {
+  all = 'all',
+  active = 'active',
+  completed = 'completed',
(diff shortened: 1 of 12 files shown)


anastasiiavorobiova (changes requested):
Looks like the response is sent. Try to skip this test and pass other

pasha1932:
> Looks like the response is sent. Try to skip this test and pass other

image

src/App.tsx
```diff
@@ -1,26 +1,227 @@
...
+      userId: USER_ID,
+      title: textField,
+      completed: false,
+    };
+
+    setIsSubmiting(true);
+    setTempTodo({
+      id: 0,
+      userId: USER_ID,
+      title: textField,
```
anastasiiavorobiova:
```suggestion
      title: textField.trim(),
```

anastasiiavorobiova: To pass the next test

src/App.tsx
```diff
@@ -1,26 +1,227 @@
...
+    });
+
+    addTodo(newTodo)
+      .then(todo => {
+        setTodos(currentTodos => [...currentTodos, todo]);
+        setTextField('');
+      })
+      .catch(error => {
+        setErrorMessage('Unable to add a todo');
+        throw error;
```
anastasiiavorobiova:
```suggestion

```

src/App.tsx
```diff
@@ -1,26 +1,227 @@
...
+      .then(() => {
+        setTodos(currentTodos => {
+          return currentTodos.map(todo =>
+            todo.id === updatedTodo.id ? updatedTodo : todo,
+          );
+        });
+      })
+      .catch(error => {
+        setErrorMessage('Unable to update a todo');
+        throw error;
```
anastasiiavorobiova:
```suggestion

```

anastasiiavorobiova (changes requested):
Added a comment to fix the test 
Also, you can fix the test file to pass the create todo test 
Screenshot 2024-10-07 at 17 43 18

pasha1932:
some tests passed one time, the second time did not, and therefore I missed some

SanyaBratashchuk (approved):

PR #1486, opened 2024-10-21
solved task

PikalovAndrey:
[DEMO LINK](https://PikalovAndrey.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,26 +1,184 @@
...
+    loadingTodosCount,
+    inputRef,
+    inputValue,
+    handleSubmit,
+    handleUpdateTodo,
+    handleTodoDelete,
+    setLoadingTodosCount,
+    setTodos,
+    setInputValue,
+  } = useTodos(setErrorMessage);
```
Ann-Maksay, line 25: In my opinion, it's rather confusing. Part of the logic associated with todo is in the custom hook, and the other part is spread across components. It is better to either put everything in the hook or not use it at all.

src/App.tsx
```diff
@@ -1,26 +1,184 @@
...
+  const handleCompletedTodosDeleted = useCallback(() => {
+    const todosForDeleting = todos
+      .filter(todo => todo.completed)
+      .map(todo => todo.id);
+
+    setLoadingTodosCount(current => [...current, ...todosForDeleting]);
+
+    Promise.all(
+      todosForDeleting.map(id =>
+        deleteTodos(id)
```
Ann-Maksay: I'd rather think about how to avoid repeating the code. It uses the same logic as deleting a single element, so maybe you should think about how to reuse the already existing code?
This applies to deletion as well as any updates.

src/components/Header.tsx
```diff
@@ -0,0 +1,55 @@
...
+  value,
+  todos,
+  completedTodos,
+  inputRef,
+  loadingTodos,
+  onTodosToggle,
+  onSubmit,
+  onInputChange,
+}) => {
+  const isLoading = loadingTodos.length > 0;
```
Ann-Maksay, line 26:
```suggestion
  !!loadingTodos.length
```

src/components/TodoComponent.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+interface TodoComponentProps {
+  todo: Todo;
+  isLoading: boolean;
+  tempTodo: Todo | null;
+  inputRef: React.MutableRefObject<HTMLInputElement | null>;
+  onTodoDelete: (todoId: number) => void;
+  onTodoToggle: (todoId: number, currentCompletedStatus: boolean) => void;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  onUpdateTodo: any;
```
Ann-Maksay:
```suggestion
  onUpdateTodo: Promise<Todo>;
```

Wouldn't that be better? Or `Promise<unknown>` at least

Ann-Maksay (approved):
Nice work, but you should consider code consistency. For example, if some logic is in the hook, it should all be there. I won't reject it, but I highly recommend you to pay attention to it.
And a lot of code duplication.

src/App.tsx
igorkruz: Good point!

src/components/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+  filter,
+  setFilter,
+  onCompletedTodosDeleted,
+}) => {
+  const atLeastOneIsActive = todos.some(todo => todo.completed);
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {todos.length - completedTodos.length} items left
```
igorkruz: Create a variable for this calculation

src/components/TodoComponent.tsx
```diff
@@ -0,0 +1,165 @@
...
+    setIsEditing(false);
+  };
+
+  const isTempTodo = tempTodo && tempTodo.id === todo.id;
+
+  useEffect(() => {
+    if (isEditing && inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [isEditing, inputRef]); //?
```
igorkruz: Delete commented code

src/utils/updateTodo.ts
```diff
@@ -0,0 +1,49 @@
...
+): Promise<void> => {
+  setLoadingTodosCount(current => [...current, ...todoIds]);
+
+  const promises = todoIds.map(id => {
+    const action =
+      newStatus !== null
+        ? operation(id, { completed: newStatus })
+        : deleteOperation
+          ? deleteOperation(id)
+          : operation(id);
```
igorkruz: The nested ternary operator can be hard to read. Consider refactoring that section for clarity.

src/App.tsx
```diff
@@ -1,26 +1,97 @@
...
+
+  useEffect(() => {
+    if (errorMessage) {
+      const timer = setTimeout(() => {
+        setErrorMessage(ErrorMessages.NO_ERRORS);
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+  }, [errorMessage]);
```
igorkruz: Move to error notification file

igorkruz (changes requested):
Good Job 👍 
let's improve your solution

src/components/Errors.tsx
```diff
@@ -0,0 +1,38 @@
...
+export const Errors: FC<ErrorsProps> = ({ errorMessage, setErrorMessage }) => {
+  useEffect(() => {
+    if (errorMessage) {
+      const timer = setTimeout(() => {
+        setErrorMessage(ErrorMessages.NO_ERRORS);
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+  }, [errorMessage]);
```
KosolapovAnatolii: GitGab Actions gives you this recomendation:
'React Hook useEffect has a missing dependency: 'setErrorMessage'. Either include it or remove the dependency array. If 'setErrorMessage' changes too often, find the parent component that defines it and wrap that definition in useCallback'
Check it in your code

KosolapovAnatolii (changes requested):
https://github.com/user-attachments/assets/756a5772-0da7-4fde-9c41-2f1a9ddbc9f9

Good job! Left fix few bugs. Watch video, I've showed input doesn't recive focus after adding new todo, but it has to. 
Fix it and go forward to the next task!
P.s.: cool case to slove this task using custom hook! 🆒 😎

PikalovAndrey:
It's strange, that you have such behaviour, because from my side it works perfectly.🤨

KosolapovAnatolii (approved):
Completed! 🔥 🔥

PR #1487, opened 2024-10-21
feat: update title, toggle status

Ann-Maksay:
[DEMO LINK](https://Ann-Maksay.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,26 +1,219 @@
...
+    [todos],
+  );
+
+  const handleResetError = () => setError(Error.DEFAULT);
+
+  const handleError = (message: Error) => {
+    setError(message);
+
+    setTimeout(handleResetError, 3000);
+  };
```
YuriiMykhailenko: Add clear timer after 3 sec. Use logic from todays lesson or debounce

src/App.tsx
```diff
@@ -1,26 +1,219 @@
...
+
+  useEffect(() => {
+    getTodos()
+      .then(currentTodos => {
+        setTodos(currentTodos);
+      })
+      .catch(() => {
+        handleError(Error.LOADING_TODOS);
+      });
+  }, []);
```
YuriiMykhailenko: Please add here the dependencies.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,125 @@
...
+    if (isEditing) {
+      document.addEventListener('keyup', handleReverseUpdate);
+
+      editTitleInput.current?.focus();
+    }
+
+    return () => {
+      document.removeEventListener('keyup', handleReverseUpdate);
+    };
+  }, [isEditing]);
```
YuriiMykhailenko, line 60: here you missed dependency too

YuriiMykhailenko (approved):
You did well👍 But I think you should make a custom hook to reduce size of App.tsx, it's enormous.

  Check the working version, when you're editing a todo title and make double click on the other in case of saving-error it would not refocus on new todo(but your app works well in other cases so it's up to you).
  
   Also consider my comments😉

src/App.tsx
Ann-Maksay: A linter's warning is not necessarily a call to action. In this case, we expect this useEffect to be triggered just once regardless of anything, so an empty array of dependencies is more appropriate.

Ann-Maksay:
In my opinion, using a custom hook like useTodo here is more for practice than a real necessity. Since our application is only responsible for working with Todo, there is no real need to separate the logic. App.tsx size doesn't matter much as long as all the logic is in one place.

src/App.tsx
```diff
@@ -1,26 +1,213 @@
...
+  };
+
+  const handleDeleteCompleted = () => {
+    const completedIds = todos.reduce<number[]>((ids, todo) => {
+      if (todo.completed) {
+        ids.push(todo.id);
+      }
+
+      return ids;
+    }, []);
```
igorkruz: Use `filter `to create an array of completed todos, then `map `to extract their IDs. This improves readability and keeps the code straightforward.
```suggestion
const completedIds = todos
    .filter(todo => todo.completed)
    .map(todo => todo.id);
```

src/App.tsx
```diff
@@ -1,26 +1,213 @@
...
+  };
+
+  useEffect(() => {
+    getTodos()
+      .then(currentTodos => {
+        setTodos(currentTodos);
+      })
+      .catch(() => {
+        handleError(Error.LOADING_TODOS);
+      });
```
igorkruz: It's better to move the logic inside the `useEffect `hook to a separate function, as it improves code readability, reusability, and makes testing easier. It also helps keep `useEffect `focused on side effects.

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { getFilterLink } from '../../utils';
+
+type Props = {
+  filterOption: FilterOption;
+  onFilter: Dispatch<SetStateAction<FilterOption>>;
+};
+
+const filterOptionsArray = Object.values(FilterOption);
+
+export const Filter: React.FC<Props> = ({ filterOption, onFilter }) => {
```
igorkruz:
```suggestion
export const Filter: FC<Props> = ({ filterOption, onFilter }) => {
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Filter } from '../Filter/Filter';
+
+type Props = {
+  filterOption: FilterOption;
+  setFilterOption: Dispatch<SetStateAction<FilterOption>>;
+  todosAmount: number;
+  uncompletedTodosAmount: number;
+  onDeleteCompleted: () => void;
+};
+export const Footer: React.FC<Props> = ({
```
igorkruz:
```suggestion
export const Footer: FC<Props> = ({
```

igorkruz (changes requested):
Good Job 👍 
let's improve your solution

igorkruz (approved):
Well done 🔥

PR #1508, opened 2024-11-02
solution

Daryna-Kukharets:
[DEMO LINK](https://Daryna-Kukharets.github.io/react_todo-app-with-api/)


SanyaBratashchuk (approved):
image

PR #1517, opened 2024-11-15
finishing the chalenge

italomagno:
[DEMO LINK](https://italomagno.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,110 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { ErrorComponent } from './components/ErrorComponent';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [editingTodosId, setEditingTodosId] = useState<number[]>([]);
+  const [isActiveFilter, setIsActiveFilter] = useState(false);
+  const [filter, setFilter] = useState('all');
+
+  async function loadAllTodos() {
+    try {
+      const loadedTodos = await getTodos();
+
+      if (loadedTodos.length > 0) {
+        setIsActiveFilter(true);
+      }
+
+      setTodos(loadedTodos);
+      setFilteredTodos(loadedTodos);
+    } catch (e) {
+      setError('Unable to load todos');
+    }
+  }
+
+  useEffect(() => {
+    loadAllTodos();
+  }, []);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setError(null);
+    }, 3000);
+  }, [error]);
+
+  useEffect(() => {
+    if (todos.length === 0) {
+      setIsActiveFilter(false);
+
+      return;
+    }
+
+    setIsActiveFilter(true);
+  }, [todos.length]);
+
+  useEffect(() => {
+    if (filter === 'completed') {
+      setFilteredTodos(filteredTodos.filter(t => t.completed));
+    }
+
+    if (filter === 'active') {
+      setFilteredTodos(filteredTodos.filter(t => !t.completed));
+    }
+  }, [filter, filteredTodos.length, todos.length]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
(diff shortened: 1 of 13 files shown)


src/components/Header.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { addTodo, updateTodo } from '../api/todos';
+
+interface HeaderProps {
+  todos: Todo[];
+  setEditingTodosId: (ids: number[]) => void;
+  setError: (error: string) => void;
+  editingTodosId: number[];
+  setTodos: (todos: Todo[]) => void;
+  setFilteredTodos: (todos: Todo[]) => void;
+  setIsActiveFilter: (isActive: boolean) => void;
```
raulriato: You're not using the `setIsActiveFilter` in the header, so it shouldn't be in the declaration of the `props` type

src/App.tsx
```diff
@@ -1,26 +1,110 @@
...
+      <h1 className="todoapp__title">todos</h1>
+      <div className="todoapp__content">
+        <Header
+          todos={todos}
+          setEditingTodosId={setEditingTodosId}
+          setError={setError}
+          editingTodosId={editingTodosId}
+          setTodos={setTodos}
+          setFilteredTodos={setFilteredTodos}
+          setIsActiveFilter={setIsActiveFilter}
```
raulriato: It has no effect to sen `setIsActiveFilter` if you are not using it in the header

src/App.tsx
```diff
@@ -1,26 +1,110 @@
...
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { ErrorComponent } from './components/ErrorComponent';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [editingTodosId, setEditingTodosId] = useState<number[]>([]);
+  const [isActiveFilter, setIsActiveFilter] = useState(false);
```
raulriato: correct me if I'm wrong. This variable refers to the state of the footer, right? Whether if it should be rendered or not. If that's so, you could name it better, to something like `isFooterActive` and `setIsFooterActive`, you don't need the `Filter` because it is a boolean state, so it has only two possible states, and it's better to use the `Footer` in the name to make ir more specific

raulriato: The way it is, it can be confusing because you actually have a `Active` filter on the application

src/App.tsx
```diff
@@ -1,26 +1,110 @@
...
+
+  useEffect(() => {
+    if (filter === 'completed') {
+      setFilteredTodos(prev => prev.filter(t => t.completed));
+    }
+
+    if (filter === 'active') {
+      setFilteredTodos(prev => prev.filter(t => !t.completed));
+    }
+  }, [filter, filteredTodos, todos]);
```
raulriato: It's not a good idea to have the `filteredTodos` as a dependency here. It causes your list to not render all the todos when you change the filter from `Active`/`Completed` to `All`.

src/components/Header.tsx
```diff
@@ -0,0 +1,146 @@
...
+        id: newTodoIdAvailable,
+      };
+
+      setFilteredTodos([...todos, newTodo as Todo]);
+
+      const responseFromAddedTodo = await addTodo(newTodoTitle.trim());
+
+      setFilteredTodos([...todos].filter(t => t.id !== newTodoIdAvailable));
+      setTodos([...todos, responseFromAddedTodo]);
+      setFilteredTodos([...todos, responseFromAddedTodo]);
```
raulriato: This part of the code is a little confusing, you're manipolating the `filteredTodos` too much, making the rendered `todos` messy. Try imagining what you want to be rendered when you add a new `todo`, think of all scenarios, try adding a new `todo` when you have each filter on, and what you want to be rendered on the screen while adding the new `todo` in each case.

src/components/Header.tsx
```diff
@@ -0,0 +1,146 @@
...
+    } finally {
+      setEditingTodosId(editingTodosId.filter(id => id !== newTodoIdAvailable));
+
+      setIsAddingNewTodo(false);
+    }
+  }
+
+  function handleToggleAllTodos() {
+    const successfulUpdatedTodo: number[] = [];
+    const hasAllSameStatus = todos.every(t => t.completed);
```
raulriato: This is not actually a mistake, just an possible improvement.

add a `Complete` to the name of the variable. You want it to have the value `true` if all the `todos` have status `completed`. When you name the varibale `hasAllSameStatus` it gives the impression that if the status of all the `todos` is `not completed`, it would be true as well.

raulriato (changes requested):
You're in the way. Good job!

Just need a few changes

src/App.tsx
```diff
@@ -1,26 +1,109 @@
...
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { ErrorComponent } from './components/ErrorComponent';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [editingTodosId, setEditingTodosId] = useState<number[]>([]);
+  const [isFooterActive, setIsFooterActiveFilter] = useState(false);
```
raulriato, line 15: Try to keep the name of the variable and the setter with the same name.

raulriato (approved):
Approved. Good job! Just an advice for next projects.

PR #1524, opened 2024-11-27
add task solution

sophiasph:
[DEMO LINK](https://sophiasph.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,158 @@
-/* eslint-disable max-len */
+/* eslint-disable react-hooks/rules-of-hooks */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import * as todoService from './api/todos';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { Todo } from './types/Todo';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Filter } from './types/Filter';
+import { ErrorNotification } from './components/Error';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<Filter>(Filter.all);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todosInProcess, setTodosInProcess] = useState<number[]>([]);
+  const [newTitleTodo, setNewTitleTodo] = useState('');
+
+  useEffect(() => {
+    todoService
+      .getTodos()
+      .then(setTodos)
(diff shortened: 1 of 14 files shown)


src/api/todos.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
+};
+
+export const postTodos = ({ userId, title, completed }: Omit<Todo, 'id'>) => {
+  return client.post<Todo>('/todos', { userId, title, completed });
+};
+
+export const deleteTodos = (postId: number) => {
```
TarasHoliuk:
```suggestion
export const deleteTodo = (todoId: number) => {
```
You delete (and add, update) only one todo
Also, it's not a post - it's todo)

src/api/todos.ts
```diff
@@ -0,0 +1,20 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 1873;
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
+};
+
+export const postTodos = ({ userId, title, completed }: Omit<Todo, 'id'>) => {
```
TarasHoliuk:
```suggestion
export const addTodo = ({ userId, title, completed }: Omit<Todo, 'id'>) => {
```
you can name function depending on what does it do. It shouldn't reflect a request method in the name

src/App.tsx
```diff
@@ -1,26 +1,158 @@
...
+    setTodosInProcess(currentId => [...currentId, postId]);
+
+    return todoService
+      .deleteTodos(postId)
+      .then(() => {
+        setTodos(currentTodo => currentTodo.filter(todo => todo.id !== postId));
+      })
+      .catch(() => {
+        setErrorMessage('Unable to delete a todo');
+        setTimeout(() => setErrorMessage(''), 3000);
```
TarasHoliuk: You can create a separate useEffect for it - check if error exist - if yes setTimeout to clear it

src/components/Header.tsx
```diff
@@ -0,0 +1,112 @@
...
+};
+
+export const Header: React.FC<Props> = ({
+  todos,
+  onErrorMessage,
+  onSubmit,
+  setTempTodo,
+  todosInProcess,
+  updateTodo,
+  newTitleTodo: newTodoTitle,
```
TarasHoliuk: nit: you can name props like it instead of renaming

src/App.tsx
```diff
@@ -1,26 +1,158 @@
-/* eslint-disable max-len */
+/* eslint-disable react-hooks/rules-of-hooks */
```
TarasHoliuk: Why do you need to disable this rule? Better not do it and figure the problem out

TarasHoliuk (changes requested):
GJ 👍 
Added a few minor comments

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1525, opened 2024-11-29
todo app withh api solution

ZupaFly:
[DEMO LINK](https://zupafly.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,6 +7,7 @@ import TodoList from './components/TodoList';
 import Footer from './components/Footer';
 import ErrorNotification from './components/ErrorNotification';
 import classNames from 'classnames';
+import { FilterType } from './types/Filter';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,26 +1,229 @@
...
+      sethaveId([]);
+    }
+  };
+
+  const filteredTodos = todos.filter(todo => {
+    if (filter === 'active') {
+      return !todo.completed;
+    }
+
+    if (filter === 'completed') {
```
volodymyr-soltys97: Create a enum for 'all', 'active', 'completed' and use it everywhere

volodymyr-soltys97 (changes requested):
Good job!
You need to add this data-cy attribute for fix failed test, if you need some help feel free in the fe_chat 
image

etojeDenys (approved):
good job, only one comment

1. after adding the todo you need to focus the input

PR #1529, opened 2024-12-02
add task solution

vneholiuk:
[DEMO LINK](https://vneholiuk.github.io/react_todo-app-with-api/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to hide the error message by default 
image

And fix the delete logic 
image

volodymyr-soltys97 (changes requested):
Hi, the demo link still not fixed from the previous review, you need to make deploy again and check it after on the demo

VitaliyBondarenko1982 (approved):
Good job.

PR #1532, opened 2024-12-11
solution 0.01 one test failed

StanislavKapytsia:
Знімок екрана 2024-12-11 210335

[DEMO LINK](https://StanislavKapytsia.github.io/react_todo-app-with-api/)

## Toggling a todo status

Toggle the `completed` status on `TodoStatus` change:
- Install Prettier Extention and use this [VSCode settings](https://mate-academy.github.io/fe-program/tools/vscode/settings.json) to enable format on save.
- covered the todo with a loader overlay while waiting for API response;
- the status should be changed on success;
- show the `Unable to update a todo` notification in case of API error.

Add the ability to toggle the completed status of all the todos with the `toggleAll` checkbox:

- `toggleAll` button should have `active` class only if all the todos are completed;
- `toggleAll` click changes its status to the opposite one, and sets this new status to all the todos;
- it should work the same as several individual updates of the todos which statuses were actually changed;
- do send requests for the todos that were not changed;

## Renaming a todo

Implement the ability to edit a todo title on double click:

- show the edit form instead of the title and remove button;
- cancel editing on `Esс` key `keyup` event;
- if the title was changed show the loader while waiting for the API response;
- update the todo title on success;
- show `Unable to update a todo` in case of API error;
- or the deletion error message if we tried to delete the todo.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,15 @@
-import React, { useEffect, useMemo, useRef, useState } from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import { addTodo, delTodo, get, updTodo } from './api/todos';
 import { TodoInterface } from './types/Todo';
 import { Filter } from './types/filter';
-import { TodoList } from './components/todoList/todoList';
-import { FilteredTodoList } from './components/footer/filteredTodoList';
+import { TodoList } from './components/todoList/TodoList';
+import { FilteredTodoList } from './components/footer/FilteredTodoList';
 import classNames from 'classnames';
 
 export const App: React.FC = () => {
@@ -12,17 +18,16 @@ export const App: React.FC = () => {
   const [value, setValue] = useState('');
   const [tempTodo, setTempTodo] = useState<TodoInterface | null>(null);
   const [errorMessage, setErrorMessage] = useState('');
-  const [active, setActive] = useState(false);
   const [todosForModify, setTodosForModify] = useState<TodoInterface[]>([]);
 
   const inputForFocusRef = useRef<HTMLInputElement>(null);
   const notificationRef = useRef<HTMLDivElement>(null);
 
-  const hideNotification = () => {
+  const hideNotification = useCallback(() => {
     if (notificationRef.current) {
       notificationRef.current.classList.add('hidden');
     }
-  };
+  }, []);
 
   const errorHandling = (error: Error) => {
     if (notificationRef.current) {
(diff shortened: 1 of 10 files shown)


StanislavKapytsia:
i can't understand failed test,
1) i made delay 3000 for every request.
2) wait till couple of them was deleted
3) then switch offline
4) all todos before switch was delete and after all todos was saved 
5) i got error message once after switch
 help

src/App.tsx
```diff
@@ -1,26 +1,297 @@
...
+    return todos.every(item => item.completed);
+  }, [todos]);
+
+  useEffect(() => {
+    if (allTodosComplited) {
+      setActive(true);
+    } else {
+      setActive(false);
+    }
+  }, [allTodosComplited, active]);
```
etojeDenys:
```suggestion
  const allTodosComplited = useMemo(() => {
    return todos.every(item => item.completed);
  }, [todos]);
```
it looks redundant, you can just use allTodosComplited var

src/components/footer/FilteredTodoList.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  setTodosForModify,
+}) => {
+  const countNotCompletedItem = useCallback(() => {
+    const filtered = todos.filter(todo => !todo.completed);
+
+    return filtered.length;
+  }, [todos]);
+
+  const notCompletedItem = countNotCompletedItem();
```
etojeDenys:
```suggestion
  const notCompletedItem = useMemo(() => {
    const filtered = todos.filter(todo => !todo.completed);

    return filtered.length;
  }, [todos]);
```

etojeDenys (changes requested):
on completed all click and toggle all click you need to update all todos immediately

https://github.com/user-attachments/assets/a8f26bab-e0c6-4971-8660-213742302c2b

src/App.tsx
StanislavKapytsia: Знімок екрана 2024-12-13 162017

src/components/footer/FilteredTodoList.tsx
StanislavKapytsia: Знімок екрана 2024-12-13 163222

StanislavKapytsia: Знімок екрана 2024-12-13 162837

src/App.tsx
```diff
@@ -1,26 +1,300 @@
...
+      content = copyTodos.map(item => ({ ...item, completed: false }));
+    } else {
+      content = copyTodos
+        .filter(todo => !todo.completed)
+        .map(item => ({ ...item, completed: true }));
+    }
+
+    setTodosForModify(content);
+
+    await Promise.allSettled(content.map(todo => updateTodos([todo])));
```
anastasiiavorobiova:
```suggestion
    await Promise.allSettled(content.map(todo => updateTodos([todo]))).then(here you should filter just successfully removed todos);
```

anastasiiavorobiova (changes requested):
Great work! Almost done

src/App.tsx
StanislavKapytsia: Знімок екрана 2024-12-15 164959

VitaliyBondarenko1982 (approved):
Looks good

PR #1541, opened 2024-12-19
solution

save-v:
https://save-v.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,14 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, {
-  useEffect,
-  useRef,
-  useState,
-  Dispatch,
-  SetStateAction,
-  FormEvent,
-  useMemo,
-} from 'react';
-import { USER_ID } from './api/todos';
+import React, { useEffect, useRef, useState, FormEvent, useMemo } from 'react';
+import {
+  deleteTodos,
+  getTodos,
+  patchTodos,
+  postTodos,
+  USER_ID,
+} from './api/todos';
 import { Todo } from './types/Todo';
-import { client } from './utils/fetchClient';
 import cN from 'classnames';
 import { TodoItem } from './components/TodoItem';
 import { Filter } from './types/Filter';
@@ -27,10 +24,9 @@ export const App: React.FC = () => {
   const [errorMessage, setErrorMessage] = useState<Error>(Error.Default);
   const [activeFilter, setActiveFilter] = useState<Filter>(Filter.All);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
-  const [todoToDeleteIds, setTodoToDeleteIds] = useState<number[]>([]);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
 
   const [inputText, setInputText] = useState('');
-  const [statusChangeId, setStatusChangeId] = useState<number[]>([]);
 
   const addTodoField = useRef<HTMLInputElement>(null);
   const errorTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
@@ -70,36 +66,35 @@ export const App: React.FC = () => {
   }
 
   function filterToBool(filter: Filter) {
-    let boolFilter = null;
+    let boolFilter;
 
-    if (filter === Filter.Active) {
-      boolFilter = false;
-    } else if (filter === Filter.Completed) {
-      boolFilter = true;
+    switch (filter) {
+      case Filter.Active:
+        boolFilter = false;
+        break;
+      case Filter.Completed:
+        boolFilter = true;
+        break;
+      default:
+        boolFilter = null;
+        break;
     }
 
     return boolFilter;
   }
 
-  function changeState(
-    id: number,
-    todosState: Dispatch<SetStateAction<Todo[]>>,
-    updatedTodo: Todo,
-  ) {
-    todosState(prev => {
-      const changed = prev.map(todo => (todo.id === id ? updatedTodo : todo));
-
-      return changed;
-    });
+  function changeTodos(fetchedTodo: Todo) {
+    setTodos(prevTodos =>
+      prevTodos.map(todo => (todo.id === fetchedTodo.id ? fetchedTodo : todo)),
+    );
   }
 
-  function handleTitleChange(newTitle: string, editingTodoId: number | null) {
-    const updateStatus = { title: newTitle };
+  function handleTitleChange(editingTodoId: number | null, newTitle: string) {
+    const updateTitle = { title: newTitle };
 
-    return client
-      .patch<Todo>(`/todos/${editingTodoId}`, updateStatus)
+    return patchTodos(editingTodoId, updateTitle)
       .then(fetchedTodo => {
-        changeState(editingTodoId as number, setTodos, fetchedTodo);
+        changeTodos(fetchedTodo);
       })
       .catch(error => {
         showError(Error.UpdateError);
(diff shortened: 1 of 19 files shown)


Welbrn (changes requested):
"Good job! However, when I press the 'Toggle All' button in the header, it changes the status of the items, but does not reflect the change visually. Additionally, the update is only applied locally and not on the server."

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
-    return <UserWarning />;
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<Error>(Error.Default);
+  const [activeFilter, setActiveFilter] = useState<Filter>(Filter.All);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todoToDeleteIds, setTodoToDeleteIds] = useState<number[]>([]);
+
+  const [inputText, setInputText] = useState('');
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [statusChangeId, setStatusChangeId] = useState<number[]>([]);
```
BudnikOleksii:
```suggestion
  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
```
One state for todo ids is enough

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+  }
+
+  function filterToBool(filter: Filter) {
+    let boolFilter = null;
+
+    if (filter === Filter.Active) {
+      boolFilter = false;
+    } else if (filter === Filter.Completed) {
+      boolFilter = true;
+    }
```
BudnikOleksii: Use switch case instead. And you can simplify this condition

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+    updatedTodo: Todo,
+  ) {
+    const filter = filterToBool(activeFilter);
+
+    todosState(prev => {
+      let changed = prev.map(todo => (todo.id === id ? updatedTodo : todo));
+
+      if (todosState === setFilteredTodos) {
+        changed = filterTodosByStatus(changed, filter);
+      }
```
BudnikOleksii: I have no idea what is happening here. Try to simplify

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+
+      return changed;
+    });
+  }
+
+  function handleTitleChange(newTitle: string, editingTodoId: number | null) {
+    const updateStatus = { title: newTitle };
```
BudnikOleksii: Why do you call it status when actually store a title?

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+
+    return client
+      .patch<Todo>(`/todos/${editingTodoId}`, updateStatus)
```
BudnikOleksii: Create a function to do it. Your component shouldn't know anything about URL where you are sending the request

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+      .then(fetchedTodo => {
+        changeState(editingTodoId as number, setFilteredTodos, fetchedTodo);
+        changeState(editingTodoId as number, setTodos, fetchedTodo);
```
BudnikOleksii: ???
If I understand correct `handleTitleChange` function should update the todo so just do it when you get it from the API.
```typescript
setTodos(prevTodos => prevTodos.map(todo => todo.id === fetchedTodo.id ? fetchedTodo : todo));
```

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+  function handleTodoStatusChange(id: number, newStatus: boolean) {
+    setStatusChangeId(prev => [...prev, id]);
+    const updateStatus = { completed: newStatus };
+
+    return client
+      .patch<Todo>(`/todos/${id}`, updateStatus)
+      .then(fetchedTodo => {
+        changeState(id, setFilteredTodos, fetchedTodo);
+        changeState(id, setTodos, fetchedTodo);
+      })
```
BudnikOleksii: The same, fix in all places

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+        setTodos(fetchedTodos);
+        setFilteredTodos(fetchedTodos);
+        setFocusOnAddInput();
+      })
+      .catch(() => showError(Error.LoadError));
+  }, []);
+
+  useEffect(() => {
+    if (tempTodo === null) {
+      /*tempTodo === null для того, не виконувати це два рази (бо стейт tempTodo спочатку змінюється на об'єкт а потім змінюється на null)*/
```
BudnikOleksii: Remove all comments. If you simplify your code and provide good names everyone understand what your code is doing without any comments

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+      (acc, todo) => (todo.completed ? [...acc, todo.id] : acc),
+      [] as number[],
+    );
+
+    setTodoToDeleteIds(completedTodoIds);
+
+    const promises: Promise<void>[] = [];
+
+    completedTodoIds.forEach(id => {
+      promises.push(onDelete(id));
```
BudnikOleksii: Use `map` instead

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
-  if (!USER_ID) {
-    return <UserWarning />;
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<Error>(Error.Default);
+  const [activeFilter, setActiveFilter] = useState<Filter>(Filter.All);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todoToDeleteIds, setTodoToDeleteIds] = useState<number[]>([]);
+
+  const [inputText, setInputText] = useState('');
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
BudnikOleksii: You don't need one more state for todos. `filteredTodos` are dynamic and depends on `activeFilter` so just store in the variable

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+    completedTodoIds.forEach(id => {
+      promises.push(onDelete(id));
+    });
+
+    Promise.all(promises).then(() => {
+      setFocusOnAddInput();
+    });
+  }
+
+  //96 221
```
BudnikOleksii: ???

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+    setFilteredTodos(filterTodosByStatus(todos, filter));
+    setActiveFilter(filterParam);
+  }
+
+  useEffect(() => {
+    handleFilter(activeFilter); //&&&
+  });
+
+  function changeStatusAll() {
+    const status = isAllCompleted ? false : true;
```
BudnikOleksii: It's already boolean

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
-        </a>
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
```
BudnikOleksii: Create a header component

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+              type="text"
+              value={inputText}
+              className="todoapp__new-todo"
+              placeholder="What needs to be done?"
+              onChange={event => setInputText(event.target.value)}
+              disabled={tempTodo !== null}
+            />
+          </form>
+        </header>
+        <section className="todoapp__main" data-cy="TodoList">
```
BudnikOleksii: Create a TodoList component

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+            />
+          ))}
+          {tempTodo !== null && (
+            <TodoItem key={tempTodo.id} todo={tempTodo} isTemp={true} />
+          )}
+        </section>
+
+        {/* Hide the footer if there are no todos +++*/}
+        {todos.length !== 0 && (
+          <footer className="todoapp__footer" data-cy="Footer">
```
BudnikOleksii: Create a Footer component

src/api/todos.ts
```diff
@@ -0,0 +1,10 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 2148;
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
+};
+
+// Add more methods here
```
BudnikOleksii: // Add more methods here

BudnikOleksii (changes requested):
Good work 👍 
Let's improve and simplify your solution

StasSokolov1 (approved):
It looks much more better, I'm approving it!

PR #1544, opened 2024-12-19
Develop

SeemsGood78:
[DEMO LINK](https://SeemsGood78.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,7 +2,14 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { getTodos, USER_ID, deleteTodo, addTodo, completeTodo, renameTodo } from './api/todos';
+import {
+  getTodos,
+  USER_ID,
+  deleteTodo,
+  addTodo,
+  completeTodo,
+  renameTodo,
+} from './api/todos';
 import { Todo } from './types/Todo';
 import classNames from 'classnames';
 
@@ -58,6 +65,7 @@ export const App: React.FC = () => {
 
     if (!trimmedTitle) {
       setErrorMessage('Title should not be empty');
+
       return;
     }
 
@@ -131,7 +139,10 @@ export const App: React.FC = () => {
 
   const handleComplete = (todoId: number) => {
     const currentTodo = todos.find(todo => todo.id === todoId);
-    if (!currentTodo) return;
+
+    if (!currentTodo) {
+      return;
+    }
 
     const newCompletionState = !currentTodo.completed;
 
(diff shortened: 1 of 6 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It is bad practice to write all the code in one component, you need to separate the logic into different components, for example `Header, Footer, TodoItem, TodoList`

SeemsGood78:
There is no such requirement in the specification, and I don't see the point in spending a couple of hours splitting the code into separate components if it's not a mandatory part of the task.

vadiimvooo (changes requested):
Seems like there are no changes from last review. Please feel free to ask for help in fe_chat.

etojeDenys (changes requested):
one of the main reasons you submit tasks for review is to learn how to write clean and understandable code for other developers. 

If(when) you get a job at an IT company and, after being asked to break down a 500-line component into smaller components (without overdoing it), respond with, "It wasn't in the requirements, so I won't do it," — trust me, it won't end well


also don't forget to add loading logic on toggle all and clear completed as in the working example

SeemsGood78:
Thank you for the advice I didn't ask for. When my employer asks me to break down the code, I will do it, but if you think it's necessary, you can add a requirement for ALL students to break down the code, and I will do it. However, for now, it's just a personal recommendation that doesn't really concern me. Code cleanliness is good, but not for the second task in the series. Two of your colleagues don't see this as an issue and have already accepted the assignments, so you should decide whether it's necessary or not.

SeemsGood78:
So if you really want me to break the code down, then add that to the assignment and talk to the people who accepted my assignment done 'incorrectly'. Also, thank you for the advice on logic loading that was given to me after the third review a month later.

etojeDenys (changes requested):
I appreciate your response, but this isn't just a personal recommendation—it's about maintaining general code standards, which are important regardless of who reviews the assignment. The fact that someone overlooked this earlier doesn’t mean it’s not relevant. If consistency in reviewing is an issue, it should be addressed separately, but that doesn’t change the importance of clean and maintainable code.

regarding your comment about the third review a month later—I’m not sure what you mean. You didn’t request a review or make any changes for an entire month, so it’s not that we delayed the process. If something needed to be addressed earlier, you could have followed up at any time.

SeemsGood78:
As I mentioned, months later, I meant that the other two mentors didn’t notice the problem, and I consider it their mistake. That’s why I’m going to file a complaint against them. In any case, it’s not your fault. Thank you for the response. I won’t be able to fix the code right now, but I’ll do it later.

anastasiiavorobiova (changes requested):
Try to fix previous comments 
https://www.linkedin.com/pulse/react-anti-patterns-god-component-jarrod-van-doren-jz8zc
https://dilankam.medium.com/the-god-object-anti-pattern-in-software-architecture-b2b7782d6997
UPD: https://github.com/mate-academy/react_todo-app-add-and-delete/pull/1534 
There is the same issue in the previous task
Feel free to ask for help in the chat

VitaliyBondarenko1982 (approved):
Looks good.

PR #1556, opened 2024-12-20
solution

Krrampuss:
[DEMO LINK](https://Krrampuss.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -49,30 +49,6 @@ export const App = () => {
         return todos;
     }
   }, [todos, filteredBy]);
-  //   setLoadingTodoId(todoId);
-  //   deleteTodo(todoId)
-  //     .then(() => {
-  //       setTodos(currentTodos =>
-  //         currentTodos.filter(todo => todo.id !== todoId),
-  //       );
-  //     })
-  //     .catch(() => {
-  //       setErrorMessage(Error.Delete);
-  //       setTimeout(() => setErrorMessage(Error.Default), 3000);
-  //     })
-  //     .finally(() => {
-  //       setLoadingTodoId(null);
-  //       inputRef.current?.focus();
-  //     });
-  // }, []);
-
-  // const deleteCompleted = useCallback(() => {
-  //   const completedIds = todos
-  //     .filter(todo => todo.completed)
-  //     .map(todo => todo.id);
-
-  //   completedIds.forEach(id => deletePost(id));
-  // }, [todos, deletePost]);
 
   useEffect(() => {
     loadTodos();
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,21 +1,6 @@
-/* eslint-disable react-hooks/rules-of-hooks */
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, {
-  useCallback,
-  useEffect,
-  useMemo,
-  useRef,
-  useState,
-} from 'react';
-import { UserWarning } from './UserWarning';
-import {
-  getTodos,
-  USER_ID,
-  deleteTodo,
-  addTodo,
-  updateTodo,
-} from './api/todos';
+/* eslint-disable react-hooks/exhaustive-deps */
+import { useEffect, useMemo, useRef, useState } from 'react';
+import { getTodos } from './api/todos';
 import { Todo } from './types/Todo';
 import { Error } from './types/enumError';
 import { FilterName } from './types/enumFilterName';
@@ -26,17 +11,22 @@ import { ErrorMes } from './components/ErrorMes';
 
 export const App = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [errorMessage, setErrorMessage] = useState<Error>(Error.Default);
-  const [filteredBy, setFilteredBy] = useState<FilterName>(FilterName.ALL);
+  const [errorMessage, setErrorMessage] = useState(Error.Default);
+  const [filteredBy, setFilteredBy] = useState(FilterName.ALL);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [newTodoTitle, setNewTodoTitle] = useState('');
   const [loadingTodoId, setLoadingTodoId] = useState<number | null>(null);
   const [isInputDisabled, setIsInputDisabled] = useState(false);
   const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
-  const [editingTitle, setEditingTitle] = useState<string>('');
+  const [editingTitle, setEditingTitle] = useState('');
 
-  const activeCount = todos.filter(todo => !todo.completed).length;
-  const completedTodos = todos.filter(todo => todo.completed);
+  const activeCount = useMemo(() => {
+    return todos.filter(todo => !todo.completed).length;
+  }, [todos]);
+
+  const completedTodos = useMemo(() => {
+    return todos.filter(todo => todo.completed);
+  }, [todos]);
 
   const inputRef = useRef<HTMLInputElement>(null);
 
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,26 +1,314 @@
...
+export const App = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<Error>(Error.Default);
+  const [filteredBy, setFilteredBy] = useState<FilterName>(FilterName.ALL);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [loadingTodoId, setLoadingTodoId] = useState<number | null>(null);
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
+  const [editingTitle, setEditingTitle] = useState<string>('');
```
VolodymyrKirichenko: if u use (string, number, boolean) as the initial value - u don't need a generic type
```suggestion
  const [editingTitle, setEditingTitle] = useState('');
```

src/App.tsx
```diff
@@ -1,26 +1,314 @@
...
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  useEffect(() => {
+    if (!isInputDisabled && inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [isInputDisabled]);
```
VolodymyrKirichenko: useEffect should be at the bottom, before rendering.
Thus - all useEffects will be in one place and you can easily manage them

src/App.tsx
```diff
@@ -1,26 +1,314 @@
...
+  const [filteredBy, setFilteredBy] = useState<FilterName>(FilterName.ALL);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [loadingTodoId, setLoadingTodoId] = useState<number | null>(null);
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
+  const [editingTitle, setEditingTitle] = useState<string>('');
+
+  const activeCount = todos.filter(todo => !todo.completed).length;
+  const completedTodos = todos.filter(todo => todo.completed);
```
VolodymyrKirichenko: useMemo?

src/App.tsx
```diff
@@ -1,26 +1,314 @@
...
+
+  const deleteCompleted = useCallback(() => {
+    const completedIds = todos
+      .filter(todo => todo.completed)
+      .map(todo => todo.id);
+
+    completedIds.forEach(id => deletePost(id));
+  }, [todos, deletePost]);
+
+  function togglCompleted(todoId: number) {
```
VolodymyrKirichenko: syntax error
```suggestion
  function toggleCompleted(todoId: number) {
```

src/App.tsx
```diff
@@ -1,26 +1,314 @@
...
+
+    const failedIds: number[] = [];
+
+    for (const todo of todosToUpdate) {
+      try {
+        await updateTodo(todo.id, newCompletionState, todo.title);
+      } catch {
+        failedIds.push(todo.id);
+      }
+    }
```
VolodymyrKirichenko: useCallback?

src/App.tsx
```diff
@@ -1,26 +1,314 @@
...
+  );
+
+  useEffect(() => {
+    if (errorMessage) {
+      const timer = setTimeout(() => setErrorMessage(Error.Default), 3000);
+
+      return () => clearTimeout(timer);
+    }
+
+    return;
```
VolodymyrKirichenko: do u need return?why?

src/App.tsx
```diff
@@ -1,26 +1,314 @@
...
+      const trimmedTitle = editingTitle.trim();
+      const currentTodo = todos.find(todo => todo.id === todoId);
+
+      if (currentTodo && trimmedTitle === currentTodo.title) {
+        setEditingTodoId(null);
+
+        return;
+      }
+
+      if (trimmedTitle === '') {
```
VolodymyrKirichenko: the same syntax, but more readable
```suggestion
      if (!trimmedTitle.length) {
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,53 @@
...
+            {link}
+          </a>
+        ))}
+      </nav>
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        onClick={() => deleteCompleted()}
```
VolodymyrKirichenko: the same
```suggestion
        onClick={deleteCompleted}
```

src/components/Header.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todos: Todo[];
+  completeAll: () => void;
+  activeCount: number;
+  handleSubmit: (event: React.FormEvent) => void;
+  inputRef: React.RefObject<HTMLInputElement>;
+  newTodoTitle: string;
+  setNewTodoTitle: React.Dispatch<React.SetStateAction<string>>;
```
VolodymyrKirichenko: will be easier, what do u think?
```suggestion
  setNewTodoTitle: (title: string) => void;
```

src/App.tsx
VolodymyrKirichenko: why all the logic in the parent component?
Let's unload ur component and move it a little in the child?(delete, edit, etc..)
ur component contains 314 lines..

VolodymyrKirichenko (changes requested):
Looks good, but let's make some fixes 🥹

src/App.tsx
```diff
@@ -1,26 +1,147 @@
...
+  //     });
+  // }, []);
+
+  // const deleteCompleted = useCallback(() => {
+  //   const completedIds = todos
+  //     .filter(todo => todo.completed)
+  //     .map(todo => todo.id);
+
+  //   completedIds.forEach(id => deletePost(id));
+  // }, [todos, deletePost]);
```
StasSokolov1: let's remove comments

src/api/todos.ts
```diff
@@ -0,0 +1,34 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 2129;
+
+export const getTodos = () => {
+  const baseUrl = `/todos?userId=${USER_ID}`;
```
StasSokolov1: move `/todos` into const and reuse

src/components/Header.tsx
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  function handleSubmit(event: React.FormEvent) {
+    event.preventDefault();
+
+    const trimmedTitle = newTodoTitle.trim();
+
+    if (!trimmedTitle) {
+      setErrorMessage(Error.Empty);
+      setTimeout(() => setErrorMessage(Error.Default), 3000);
```
StasSokolov1: don't forget to clear setTimeout, for ex. user click submit twice we need to clear previous timeout in this case

src/components/Header.tsx
```diff
@@ -0,0 +1,138 @@
...
+      title: trimmedTitle,
+      completed: false,
+      id: 0,
+    };
+
+    setTempTodo(loadingTodo);
+    addPost(loadingTodo);
+  }
+
+  const completeAll = useCallback(async () => {
```
StasSokolov1:
```suggestion
  const handleCompleteAll = useCallback(async () => {
```

let's use better naming for functions

StasSokolov1 (changes requested):
Almost done, let's improve your code a bit more!

StasSokolov1 (approved):
Looks like a masterpiece!

PR #1563, opened 2024-12-22
final version

AsyaDev14:
[DEMO](https://asyadev14.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,167 @@
 /* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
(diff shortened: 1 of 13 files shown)


Fridrif27 (approved):
I approved but you need to sign up for Q&A to solve tests

src/App.tsx
```diff
@@ -1,26 +1,168 @@
...
+
+  const todosCompletedNum = useMemo(
+    () => todoList.filter(todo => todo.completed).length,
+    [todoList],
+  );
+
+  const allTodosCompleted = useMemo(
+    () => todoList.length === todosCompletedNum,
+    [todoList],
+  );
```
BudnikOleksii, line 49: React Hook useMemo has a missing dependency: 'todosCompletedNum'. Either include it or remove the dependency array

Actually you don't need to use `useMemo` here as it's a primitive value

BudnikOleksii (approved):
Good work 👍

PR #1567, opened 2024-12-23
Solution

Bajkowsky:
[DEMO LINK](https://bajkowsky.github.io/react_todo-app-with-api/)


natalia-klonowska (changes requested):
when multiple errors occur one after another the time they remain visible should increase because currently some of the later errors are displayed only for a split second

natalia-klonowska (approved):
Great work! 🚀

PR #1569, opened 2024-12-23
Solution

LiliiaVol:
[DEMO LINK](https://LiliiaVol.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,16 +24,23 @@ export const App: React.FC = () => {
   const [filterType, setFilterType] = useState<FilterType>(FilterType.All);
   const inputRef = useRef<HTMLInputElement>(null);
   const inputEditRef = useRef<HTMLInputElement>(null);
+  const errorTimeoutRef = useRef<NodeJS.Timeout | null>(null);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [isAllCompleted, setIsAllCompleted] = useState(false);
   const [todoChangedTitle, setTodoChangedTitle] = useState<Todo | null>(null);
   const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([0]);
 
-  const setError = (errorType: ErrorType) => {
-    setErrorState(errorType);
+  const trimmedTitle = title.trim();
 
-    setTimeout(() => {
+  const setError = (errorType: ErrorType) => {
+    if (errorTimeoutRef.current) {
+      clearTimeout(errorTimeoutRef.current);
+    }
+
+    setErrorState(errorType);
+    errorTimeoutRef.current = setTimeout(() => {
       setErrorState(ErrorType.Default);
+      errorTimeoutRef.current = null;
     }, 3000);
   };
 
@@ -45,7 +52,7 @@ export const App: React.FC = () => {
     event.preventDefault();
     setIsAllCompleted(false);
 
-    if (!title.trim().length) {
+    if (!trimmedTitle.length) {
       setError(ErrorType.Input);
 
       return;
@@ -53,7 +60,7 @@ export const App: React.FC = () => {
 
     const newTodo = {
       userId: USER_ID,
-      title: title.trim(),
+      title: trimmedTitle,
       completed: false,
     };
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,5 @@
 /* eslint-disable max-len */
 import React, { useEffect, useMemo, useRef, useState } from 'react';
-import { UserWarning } from './UserWarning';
 import {
   deleteTodo,
   getTodos,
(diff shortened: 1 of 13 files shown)


src/components/TodoComponent/TodoComponent.tsx
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  onDoubleClickTodo: (todo: Todo) => void;
```
AnderZerfall: You have many props here. Some could be created within associated components, so your component won't rely on all of them from the outside.

src/components/TodoComponent/TodoComponent.tsx
```diff
@@ -0,0 +1,111 @@
...
+            ×
+          </button>
+        </>
+      )}
+
+      <div
+        data-cy="TodoLoader"
+        className={classNames('modal overlay', {
+          'is-active':
+            idDeletedTodo === id ||
```
AnderZerfall: For this statement, you could create some variable, which combines all possible scenarios. For example: `isLoading =todo.id === 0 || isDeleting === todo.id || isUpdating === todo.id || loadingTodos.includes(todo.id) ....` Something like that. You can create this variable in your component, outside the render.

AnderZerfall (approved):
Good job! I left some comments below. Maybe that will be useful.
Also,  I couldn't open App.tsx in review mode, so I've written some notes here:

- You can drop `finally {}` in try\catch block. The thing is, that everything after catch{} block will execute automatically regardless the result.

- Your statements in `if(...)` (I guess, I saw that in handleUpdate func and all sorts of it) look a bit too complicated - I guess, that's the result of having too many states and variables. Try to reduce anything that seems useless. You can even reuse your own code here. That will definitely clean it up a bit and make statements easier.

Anyway, that's just small suggestions. The todo app works as indented and that's the point! Good job!

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+      )}
+    >
+      <button data-cy="HideErrorButton" type="button" className="delete" />
+      {!todos.length && `Unable to load todos`}
+      {errorState[ErrorType.Input] && `Title should not be empty`}
+      {errorState[ErrorType.Add] && `Unable to add a todo`}
+      {(errorState[ErrorType.Delete] ||
+        errorState[ErrorType.CompletedDelete]) &&
+        `Unable to delete a todo`}
+      {errorState[ErrorType.Update] && `Unable to update a todo`}
```
BudnikOleksii: You can just store a message in an error state. As it's a bad idea to do like you, you have to add more and more conditions here for each case

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+        <a
+          href="#/active"
+          className={classNames('filter__link', {
+            selected: filterType === FilterType.Active,
+          })}
+          data-cy="FilterLinkActive"
+          onClick={() => {
+            setFilterType(FilterType.Active);
+          }}
```
BudnikOleksii: The links look almost the same. You can transform `FilterType` into an array with `Object.values/keys/entries` and `map` through them to generate links

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,55 @@
...
+    onSubmit,
+    inputRef,
+    title,
+    setTitle,
+    isInputDisabled,
+  } = props;
+
+  return (
+    <header className="todoapp__header">
+      {!todos.length || (
```
BudnikOleksii:
```suggestion
      {!!todos.length && (
```

src/components/TodoComponent/TodoComponent.tsx
BudnikOleksii: > You have many props here. Some could be created within associated components, so your component won't rely on all of them from the outside.

Good point

src/components/TodoComponent/TodoComponent.tsx
BudnikOleksii: > For this statement, you could create some variable, which combines all possible scenarios. For example: `isLoading =todo.id === 0 || isDeleting === todo.id || isUpdating === todo.id || loadingTodos.includes(todo.id) ....` Something like that. You can create this variable in your component, outside the render.

I think it should be simplified, just one condition is enough for example store loading todo ids and as [AnderZerfall](https://github.com/AnderZerfall) said `loadingTodoIds.includes(todo.id)`

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,106 @@
...
+      inputEditRef.current.focus();
+      inputEditRef.current.addEventListener('keydown', handleKeyDown);
+    }
+
+    return () => {
+      if (inputEditRef?.current) {
+        inputEditRef.current.removeEventListener('keydown', handleKeyDown);
+      }
+    };
+  }, [todoChangedTitle, inputEditRef]);
```
BudnikOleksii, line 58: Fix linter errors. And it's better to avoid DOM manipulations in react

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution. Also, I don't see any loaders when I click `Clear completed` button, check the working example

src/App.tsx
```diff
@@ -1,26 +1,302 @@
...
+  const [isAllCompleted, setIsAllCompleted] = useState(false);
+  const [todoChangedTitle, setTodoChangedTitle] = useState<Todo | null>(null);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([0]);
+
+  const setError = (errorType: ErrorType) => {
+    setErrorState(errorType);
+
+    setTimeout(() => {
+      setErrorState(ErrorType.Default);
+    }, 3000);
```
StasSokolov1: please don't forget to clear setTimeout

src/App.tsx
```diff
@@ -1,26 +1,302 @@
...
+
+  const filteredTodos = useMemo((): Todo[] => {
+    return filterTodos(todos, filterType);
+  }, [filterType, todos]);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    setIsAllCompleted(false);
+
+    if (!title.trim().length) {
```
StasSokolov1: please move `title.trim()` to const and reuse

src/api/todos.ts
```diff
@@ -0,0 +1,20 @@
+import { Todo, TodoResponse } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 2179;
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
```
StasSokolov1: let's move `/todos` to const and reuse

src/types/ErrorType.ts
```diff
@@ -0,0 +1,9 @@
+export enum ErrorType {
+  Default = '',
+  Input = 'Title should not be empty',
+  Add = 'Unable to add a todo',
+  Delete = 'Unable to delete a todo',
+  // CompletedDelete = 'Unable to delete a completed todo',
```
StasSokolov1: please remove redundant comment

StasSokolov1 (changes requested):
Overall looks good, let's improve your solution a bit more!

StasSokolov1 (approved):
Looks good to me, I'm approving it!

PR #1571, opened 2024-12-26
Todo App

SemenVodolazskij:
[DEMO LINK](https://SemenVodolazskij.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,293 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  addTodos,
+  deleteTodos,
+  getTodos,
+  updateTodos as updateTodo,
+  USER_ID,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
 
-const USER_ID = 0;
+import { Filter } from './types/FilterButton';
+import { Error } from './types/ErrorMessage';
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,26 +1,293 @@
...
+
+              <button
+                className={classNames('filter__link', {
+                  selected: filter === 'completed',
+                })}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setFilter(Filter.Completed)}
+              >
+                Completed
+              </button>
```
StasSokolov1, line 257: let's show these elements using .map()

src/api/todos.ts
```diff
@@ -0,0 +1,20 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 2171;
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
```
StasSokolov1, line 7: move `/todos` into const and reuse

src/App.tsx
```diff
@@ -1,26 +1,293 @@
...
+
+      activeTodos.forEach(todo => {
+        onUpdateTodo({ ...todo, completed: true });
+      });
+      setIsToogleAll(true);
+    } else {
+      todos.forEach(todo => {
+        onUpdateTodo({ ...todo, completed: false });
+      });
+      setIsToogleAll(false);
```
StasSokolov1, line 160: seems very similar can let's write one time and just change true or false depends on unCompletedTodos counter

StasSokolov1 (changes requested):
Almost done, let's improve your solution a bit more!

PR #1588, opened 2025-01-06
solution

VasylPylypchynets:
[DEMO LINK](https://VasylPylypchynets.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,506 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useEffect, useRef, useState } from 'react';
+import {
+  deleteTodo,
+  getTodos,
+  addTodo,
+  updateTodo,
+  USER_ID,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { Error } from './components/ErrorMessage/ErrorMessage';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+  const [query, setQuery] = useState<string>('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [itemsLeft, setItemsLeft] = useState<number>(0);
+  const [sortBy, setSortBy] = useState('all');
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [deleteTodoId, setDeleteTodoId] = useState<number | null>(null);
+  const [isLoadingChange, setIsLoadingChange] = useState(false);
+  const [cleanCompleted, setCleanCompleted] = useState<boolean>(false);
+  const [newTask, setNewTask] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isSubmiting, setIsSubmiting] = useState<boolean>(false);
+  const [isUpdating, setIsUpdating] = useState<number | null>(null);
+  const [itemEditingId, setItemEditingId] = useState<number | null>(null);
+  const [newTitle, setNewTitle] = useState<string>('');
+  const [todosLength, setTodosLength] = useState<number>(0);
+  const [allTodos, setAllTodos] = useState<Todo[]>([]);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  function handleFilter(sort: string) {
+    switch (sort) {
+      case 'all':
+        setTodos(currentTodos => currentTodos);
+        break;
+      case 'active':
+        setTodos(currentTodos => currentTodos.filter(todo => !todo.completed));
+        break;
+      case 'completed':
+        setTodos(currentTodos => currentTodos.filter(todo => todo.completed));
+        break;
+    }
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">
-          React Todo App - Add and Delete
-        </a>
-      </p>
+  function handleUpadateNewTitle(id: number, title: string) {
+    const todo = todos.find(currentTodo => currentTodo.id === id);
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    if (todo?.title === title) {
+      setItemEditingId(null);
+      setNewTitle('');
+
+      return;
+    }
+
+    if (title) {
+      setIsUpdating(id);
+      setErrorMessage(null);
+
+      if (!todo) {
+        return;
+      }
+
+      const updatedTitle = { title: title.trim() };
+
+      updateTodo(id, updatedTitle)
+        .then(() => {
+          setTodos(currentTodos =>
+            currentTodos.map(item =>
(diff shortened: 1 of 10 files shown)


src/api/todos.ts
```diff
@@ -0,0 +1,25 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 2085;
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
+};
+
+// Add more methods here
```
TarasHoliuk: you can delete this comment now :)

src/App.tsx
```diff
@@ -1,26 +1,506 @@
...
+        .catch(() => {
+          setErrorMessage('Unable to update a todo');
+        })
+        .finally(() => {
+          setIsUpdating(null);
+        });
+    }
   }
 
+  function handleUpdateAllTodos() {
```
TarasHoliuk: Only the last todo has loader when I click toggle all:
image

src/App.tsx
```diff
@@ -1,26 +1,506 @@
...
+              return null;
+            } finally {
+              setIsUpdating(null);
+            }
+          }
+
+          return null;
+        });
+
+        const updatedTodo = await Promise.all(updatedTodos);
```
```suggestion
        const updatedTodos = await Promise.all(updatedTodos);
```
But Promise.all will be rejected if at least one todo update was failed. You should handle result for each todo separately

src/App.tsx
```diff
@@ -1,26 +1,506 @@
...
+  const [isLoadingChange, setIsLoadingChange] = useState(false);
+  const [cleanCompleted, setCleanCompleted] = useState<boolean>(false);
+  const [newTask, setNewTask] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isSubmiting, setIsSubmiting] = useState<boolean>(false);
+  const [isUpdating, setIsUpdating] = useState<number | null>(null);
+  const [itemEditingId, setItemEditingId] = useState<number | null>(null);
+  const [newTitle, setNewTitle] = useState<string>('');
+  const [todosLength, setTodosLength] = useState<number>(0);
+  const [allTodos, setAllTodos] = useState<Todo[]>([]);
```
TarasHoliuk: You don't need todos and allTodos states. Keep only one (name `todos`, but store all todos there) state and compute filtered todos based on other states and save filtered array into a variable

Also do filtering inside useMemo callback

src/App.tsx
```diff
@@ -1,26 +1,506 @@
...
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { Error } from './components/ErrorMessage/ErrorMessage';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+  const [query, setQuery] = useState<string>('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [itemsLeft, setItemsLeft] = useState<number>(0);
```
TarasHoliuk: General rule - do not create state if you can compute it based on other state or props values

src/App.tsx
```diff
@@ -1,26 +1,506 @@
...
+  const [sortBy, setSortBy] = useState('all');
```
TarasHoliuk: Add type - it shouldn't be just a string

src/App.tsx
```diff
@@ -1,26 +1,506 @@
...
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [deleteTodoId, setDeleteTodoId] = useState<number | null>(null);
+  const [isLoadingChange, setIsLoadingChange] = useState(false);
+  const [cleanCompleted, setCleanCompleted] = useState<boolean>(false);
+  const [newTask, setNewTask] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isSubmiting, setIsSubmiting] = useState<boolean>(false);
+  const [isUpdating, setIsUpdating] = useState<number | null>(null);
```
TarasHoliuk: use the same logic and state that you use to display loader for deleting todos

src/App.tsx
```diff
@@ -1,26 +1,506 @@
...
+    e.preventDefault();
+    if (query.trim() !== '') {
+      setNewTask(query);
+      setIsSubmiting(true);
+    } else {
+      setErrorMessage('Title should not be empty');
+    }
+  }
+
+  function handleDeleteTodo(id: number) {
```
TarasHoliuk: such functions are redundant - just use `setDeleteTodoId` everywhere instead of `handleDeleteTodo`

src/App.tsx
```diff
@@ -1,26 +1,506 @@
...
+    setDeleteTodoId(id);
+  }
+
+  function handleInput(e: React.ChangeEvent<HTMLInputElement>) {
+    setQuery(e.target.value);
+  }
+
+  function handleCleanCompleted() {
+    setCleanCompleted(true);
```
TarasHoliuk: this state seems redundant
Do something directly inside handler instead of changing state and do it in useEffect callback

src/App.tsx
```diff
@@ -1,26 +1,506 @@
...
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [deleteTodoId, setDeleteTodoId] = useState<number | null>(null);
+  const [isLoadingChange, setIsLoadingChange] = useState(false);
+  const [cleanCompleted, setCleanCompleted] = useState<boolean>(false);
+  const [newTask, setNewTask] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isSubmiting, setIsSubmiting] = useState<boolean>(false);
+  const [isUpdating, setIsUpdating] = useState<number | null>(null);
+  const [itemEditingId, setItemEditingId] = useState<number | null>(null);
+  const [newTitle, setNewTitle] = useState<string>('');
```
TarasHoliuk: why do you need query, newTask and newTitle here?

For editing move state (and part of related logic, for example `itemEditingId` state - you can make it just boolean inside of that component) into TodoItem component

For input in header you need only one state

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,77 @@
...
+          data-cy="FilterLinkActive"
+          onClick={e => {
+            e.preventDefault();
+            onSortBy('active');
+          }}
+        >
+          Active
+        </a>
+
+        <a
```
TarasHoliuk: each link is almost the same - try to get rid of code duplication

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,135 @@
...
+      onSetNewTitle(todo.title);
+    }
+  }
+
+  useEffect(() => {
+    if (itemEditingId === todo.id && inputRef.current) {
+      inputRef.current.focus();
+      onSetNewTitle(todo.title);
+    }
+  }, [itemEditingId, todo.id]);
```
TarasHoliuk: fix lint:
image

TarasHoliuk (changes requested):
You have too much states - try to remove part of them and simplify logic

Feels free to ask for help in the chat if you need 🙂

TarasHoliuk (changes requested):
.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+            key={filter}
+            href={`#/${filter}`}
+            className={classNames('filter__link', {
+              selected: sortBy === filter,
+            })}
+            data-cy={`FilterLink${filter.charAt(0).toUpperCase() + filter.slice(1)}`}
+            onClick={e => {
+              e.preventDefault();
+              onSortBy(filter);
+            }}
```
volodymyr-soltys97:
```suggestion
            onClick={() => onSortBy(filter)}
```

volodymyr-soltys97 (approved):
GJ!

PR #1612, opened 2025-01-24
Final code

nikalaiii:
some tests were failed, but i have no suggestions, why - app structure & logic seems correct

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,7 +21,8 @@ export const App: React.FC = () => {
     FilterMethods.All,
   );
   const [loadTodo, setLoadTodo] = useState<Todo | null>(null);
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,26 +1,148 @@
...
+
+      return () => clearTimeout(timer);
+    }
+  }, [error]);
+
+  const countOfItemsLeft = (elements: Todo[]) => {
+    const filtered = elements.filter(el => el.completed === false);
+
+    return filtered.length;
+  };
```
etojeDenys, line 66:
```suggestion
  const countOfItemsLeft = elements.filter(el => !el.completed).lenght
```

etojeDenys (approved):

PR #1620, opened 2025-01-29
add task solution

Svimpik:
https://Svimpik.github.io/react_todo-app-with-api/


src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
+                <TodoItem
+                  todo={tempTodo}
+                  onUpdateTodo={handleUpdateTodo}
+                  onDeleteTodo={handleDeleteTodo}
+                  isLoading={isAddingLoading}
+                />
+              </CSSTransition>
+            )}
+          </TransitionGroup>
+        </section>
```
volodymyr-soltys97, line 202: Consider will be better if you move this logic to the separate component, for example `TodoList`

volodymyr-soltys97 (approved):
Good job 👍

PR #1623, opened 2025-01-31
added solution

AlNeon:
[DEMO LINK](https://AlNeon.github.io/react_todo-app-with-api/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
When we edit a todo, the loader should be shown only for that todo.
image

maxim2310 (approved):
Well done!

PR #1648, opened 2025-02-22
add task solution

romazh1988:


romazh1988:
[DEMO LINK](https://romazh1988.github.io/react_todo-app-with-api/)

src/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+        </a>
+
+        <a
+          href="#/completed"
+          className={`filter__link ${filter === FilterEnum.Completed ? 'selected' : ''}`}
+          onClick={() => setFilter(FilterEnum.Completed)}
+          data-cy="FilterLinkCompleted"
+        >
+          Completed
+        </a>
```
volodymyr-soltys97: Use `Object.values(FilterEnum)` and render these options with `map()` method

src/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {activeTodosCount} items left
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={`filter__link ${filter === FilterEnum.All ? 'selected' : ''}`}
```
volodymyr-soltys97: Use classnames library for add classes with condition, fix it everywhere

src/types/filterEnum.ts
```diff
@@ -0,0 +1,5 @@
+export enum FilterEnum {
```
volodymyr-soltys97:
```suggestion
export enum FilterType {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The edit mode looks different as on the working example, need to fix it 
image

Also need to fix this failed test, if you need some help feel free in the fe_chat 
image

anastasiiavorobiova (approved):

PR #1650, opened 2025-02-24
Develop

oleg19790807:
https://oleg19790807.github.io/react_todo-app-with-api/
I've tried to resolve that issue in different ways, but all my attempts were unsuccessful.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,8 +24,7 @@ export const App: React.FC = () => {
   const [isAddingTodo, setIsAddingTodo] = useState(false);
   const inputRef = useRef<HTMLInputElement>(null);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
-  const [deletingTodoId, setDeletingTodoId] = useState<number | null>(null);
-  const [updatingTodoId, setUpdatingTodoId] = useState<number | null>(null);
+  const [loadingTodoId, setLoadingTodoId] = useState<number | null>(null);
 
   const loadTodos = async () => {
     setIsLoading(true);
@@ -107,9 +106,10 @@ export const App: React.FC = () => {
   };
 
   const handleDeleteTodo = async (id: number) => {
-    setDeletingTodoId(id);
+    setLoadingTodoId(id);
     setErrorMessage('');
     setShowError(false);
+
     try {
       await deleteTodo(id);
       setTodos(prevTodos => prevTodos.filter(todo => todo.id !== id));
@@ -118,7 +118,7 @@ export const App: React.FC = () => {
       setShowError(true);
       setTimeout(() => setShowError(false), 3000);
     } finally {
-      setDeletingTodoId(null);
+      setLoadingTodoId(null);
       if (inputRef.current) {
         inputRef.current.focus();
       }
@@ -133,9 +133,7 @@ export const App: React.FC = () => {
 
     const deletePromises = completedTodos.map(todo =>
       deleteTodo(todo.id)
-        .then(() => {
-          return todo.id;
-        })
+        .then(() => todo.id)
         .catch(() => {
           setErrorMessage('Unable to delete a todo');
           setShowError(true);
@@ -144,11 +142,11 @@ export const App: React.FC = () => {
         }),
     );
 
-    const results: (number | null)[] = await Promise.all(deletePromises);
+    const results = await Promise.allSettled(deletePromises);
 
-    const successfullyDeletedIds = results.filter(
-      (id): id is number => id !== null,
-    );
+    const successfullyDeletedIds = results
+      .filter(result => result.status === 'fulfilled' && result.value !== null)
+      .map(result => (result as PromiseFulfilledResult<number>).value);
 
     setTodos(prevTodos =>
       prevTodos.filter(todo => !successfullyDeletedIds.includes(todo.id)),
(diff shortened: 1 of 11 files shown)


volodymyr-soltys97 (changes requested):
Hi, there is no demo link. Also you need to add the demo link to the PR description 
image

oleg19790807:
https://github.com/oleg19790807/react_todo-app-with-api?tab=readme-ov-file

TarasHoliuk:
> https://github.com/oleg19790807/react_todo-app-with-api?tab=readme-ov-file

It's not a demo link ⬆️ 

That would be the correct demo link:
https://oleg19790807.github.io/react_todo-app-with-api/

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,172 @@
...
+}) => {
+  const [isEditing, setIsEditing] = useState(false);
+  const [editTitle, setEditTitle] = useState(todo.title);
+  const [isLoading, setIsLoading] = useState(false);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (isEditing && inputRef.current) {
+      inputRef.current.focus();
+      inputRef.current.select();
```
TarasHoliuk: Do we need `select` here? I believe it's enough to `focus` the element

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,172 @@
...
+          disabled={isDeleting}
+        >
+          ×
+        </button>
+      )}
+
+      <div
+        data-cy="TodoLoader"
+        className={classNames('modal overlay', {
+          'is-active': isDeleting || isUpdating || isLoading,
```
TarasHoliuk: Isn't `isUpdating` and `isLoading` reflect the same state (when a user tries to update the todo)? I guess we can get rid of local isLoading state

src/components/Footer.tsx
```diff
@@ -0,0 +1,60 @@
...
+      })}
+      data-cy={`FilterLink${status.charAt(0).toUpperCase() + status.slice(1)}`}
+      onClick={() => setFilterStatus(status)}
+    >
+      {status.charAt(0).toUpperCase() + status.slice(1)}
+    </a>
+  ));
+
+  return (
+    <footer className={className} data-cy={dataCy}>
```
TarasHoliuk: You can write correct data-cy here (we do not use dynamic values in different cases anywhere) - it's always a static value

src/api/todos.ts
```diff
@@ -0,0 +1,48 @@
...
+//   });
+// };
+
+// export const deleteTodo = (id: number): Promise<void> => {
+//   return client.delete(`/todos/${id}`);
+// };
+
+// export const updateTodo = (id: number, data: Partial<Todo>): Promise<Todo> => {
+//   return client.patch<Todo>(`/todos/${id}`, data);
+// };
```
TarasHoliuk: Delete the commented code

src/App.tsx
```diff
@@ -1,26 +1,390 @@
...
+        })
+        .catch(() => {
+          setErrorMessage('Unable to delete a todo');
+          setShowError(true);
+
+          return null;
+        }),
+    );
+
+    const results: (number | null)[] = await Promise.all(deletePromises);
```
TarasHoliuk: To pass all tests you should handle each result separately. Promise.allSettled will be better for it

src/App.tsx
```diff
@@ -1,26 +1,390 @@
...
+      setTimeout(() => setShowError(false), 3000);
+    } finally {
+      setDeletingTodoId(null);
+      if (inputRef.current) {
+        inputRef.current.focus();
+      }
+    }
+  };
+
+  const handleClearCompleted = async () => {
```
TarasHoliuk, line 128: Would be nice to show loading animation when deleting/toggling status for all todos

src/App.tsx
```diff
@@ -1,26 +1,390 @@
...
+  const [showError, setShowError] = useState(false);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>(
+    FilterStatus.All,
+  );
+  const [isAddingTodo, setIsAddingTodo] = useState(false);
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletingTodoId, setDeletingTodoId] = useState<number | null>(null);
+  const [updatingTodoId, setUpdatingTodoId] = useState<number | null>(null);
```
TarasHoliuk: Users cannot delete and update todo at the moment. Only one action is possible. So you can get rid of one state and rename another to `loadingTodoId`

src/App.tsx
```diff
@@ -1,26 +1,390 @@
...
+
+        const updatedTodos = await Promise.all(updatePromises);
+
+        setTodos(prevTodos =>
+          prevTodos.map(
+            todo =>
+              updatedTodos.find(updated => updated.id === todo.id) || todo,
+          ),
+        );
+      } else {
```
TarasHoliuk: you duplicate part of the code in `else` block - try to avoid duplication

TarasHoliuk (changes requested):
Try to fix tests. Feel free for help in the chat if you need 🙂

volodymyr-soltys97 (approved):
Good job! 👍

PR #1651, opened 2025-02-24
First version

NickDevX:
[DEMO LINK](https://NickDevX.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,26 +1,361 @@
...
 
 export const App: React.FC = () => {
+  const input = useRef<HTMLInputElement>(null);
+  const hideErrorTimer = useRef<number | null>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletingTodoId, setDeletingTodoId] = useState<number[] | null>(null);
+  const [filter, setFilter] = useState(Filter.All);
+  const [title, setTitle] = useState('');
+  const [loadingError, setLoadingError] = useState({
```
TarasHoliuk: It's easier to save the correct error (enum value) or empty string if there is no error (empty string is a falsy value)

src/App.tsx
```diff
@@ -1,26 +1,361 @@
...
+      queryError: false,
+      addError: false,
+      todosError: false,
+      deleteError: false,
+      updateError: false,
+    });
+  }, []);
+
+  const showError = useCallback(
+    (errorType: Erors) => {
```
```suggestion
    (errorType: Errors) => {
```
use singular form for types

src/App.tsx
```diff
@@ -1,26 +1,361 @@
...
+      setLoadingError(prev => ({ ...prev, [errorType]: true }));
+      hideErrorTimer.current = window.setTimeout(() => {
+        clearError();
+        hideErrorTimer.current = null;
+      }, 3000);
+    },
+    [clearError],
+  );
+
+  const filterTodos = useCallback((todosList: Todo[], filterBy: string) => {
```
TarasHoliuk: you don't need this function separately. just do this logic in the `useMemo` callback below

src/App.tsx
```diff
@@ -1,26 +1,361 @@
...
+  const handleClick = (
+    event: React.MouseEvent<HTMLAnchorElement>,
+    filterBy: Filter,
+  ) => {
+    event.preventDefault();
+    setFilter(filterBy);
+  };
+
+  const complateTodo = (todo: Todo) => {
+    setDeletingTodoId([todo.id]);
```
TarasHoliuk: `loadingTodoId` instead of `deletingTodoId` to reflect the value better

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,64 @@
...
+        { hidden: !anyExistingError },
+      )}
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className={clsx('delete')}
+        onClick={clearError}
+      />
+      {loadingError.todosError && (
```
TarasHoliuk: It would be better to use some helper function with a switch block inside that returns the correct message. But it won't be necessary if you just save the correct error as a state value

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+            {value}
+          </a>
+        ))}
+      </nav>
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        onClick={removeCompletedTodos}
+        disabled={!checkTodoCompleted() && todos.length > 0}
```
TarasHoliuk: why it's a function? `checkTodoCompleted`. Instead just pass the correct boolean value `hasCompletedTodos`. it should also include `todos.length > 0` this

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,60 @@
...
+      ))}
+      {tempTodo && (
+        <CSSTransition key={0} timeout={300} classNames="temp-item">
+          <TodoItem
+            {...{
+              todoItem: tempTodo,
+              removeTodo,
+              complateTodo,
+              deletingTodoId,
+              isTemp: true,
```
TarasHoliuk: instead of a separate prop for it you can do the following:
- also delete deletingTodoId prop
- create isLoading prop
pass the correct boolean value for isLoading: `todoItem.id === loadingTodoId` and always pass true for temporary todo

src/App.tsx
```diff
@@ -1,26 +1,361 @@
...
+  };
+
+  const removeTodo = (id: number) => {
+    setDeletingTodoId([id]);
+    const removeTodoFromServer = async () => {
+      try {
+        const response = (await deleteTodo(id)) as number;
+
+        if (response !== 1) {
+          throw new Error();
```
TarasHoliuk: Do you catch this error somewhere? If not it will lead to page crash. Generally - do not throw errors again if you don't catch and handle them later

TarasHoliuk (changes requested):
Great job! Everything works perfectly. Left a few comments to improve your code

volodymyr-soltys97 (approved):
GJ!

PR #1656, opened 2025-03-05
solution

SerhiiUnhurian:
[DEMO LINK](https://SerhiiUnhurian.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,352 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import {
+  createTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+  USER_ID,
+} from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
+import { Footer } from './components/Footer';
+import { Filter } from './utils/Enums';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>(todos);
+  const [title, setTitle] = useState('');
+  const [loading, setLoading] = useState(false);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<Set<number>>(new Set());
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState(Filter.All);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [titleEditingId, setTitleEditingId] = useState<null | number>(null);
+
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const completedTodos = todos.filter(todo => todo.completed);
+  const notCompletedTodos = todos.filter(todo => !todo.completed);
+
+  // eslint-disable-next-line @typescript-eslint/no-shadow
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,26 +1,355 @@
...
+  const [filter, setFilter] = useState(Filter.All);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [titleEditingId, setTitleEditingId] = useState<null | number>(null);
+
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const completedTodos = todos.filter(todo => todo.completed);
+  const notCompletedTodos = todos.filter(todo => !todo.completed);
+
+  // eslint-disable-next-line @typescript-eslint/no-shadow
+  function filterTodos(todos: Todo[], filter: string) {
```
volodymyr-soltys97: You must follow one way of creating functions, fix it everywhere
```suggestion
  const filterTodos = (todos: Todo[], filter: string) => {
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,195 @@
...
+          {/* overlay will cover the todo while it is being deleted or updated */}
+          <TodoLoader loadingTodoIds={loadingTodoIds} todoId={0} />
+        </div>
+      )}
+
+      {/* This todo is being edited */}
+      {/* <div data-cy="Todo" className="todo"> */}
+      {/* <label className="todo__status-label">
+          <input
+            data-cy="TodoStatus"
```
volodymyr-soltys97: Remove all comments

src/components/TodoLoader.tsx
```diff
@@ -0,0 +1,20 @@
+import classNames from 'classnames';
+
+type Props = {
+  todoId: number;
+  loadingTodoIds: Set<number>;
```
volodymyr-soltys97, line 5: Do you need use Set here? You can use array

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/components/TodoLoader.tsx
SerhiiUnhurian: I used an array at first but when we have a delay then the loaders did not work as they should, they jumped to a new location
I came up with the idea to store all the IDs this way

src/components/TodoList.tsx
```diff
@@ -0,0 +1,146 @@
...
+                onSubmit={event => handleSubmit(event, todo)}
+                onBlur={event => handleSubmit(event, todo)}
+                onKeyUp={event => {
+                  if (event.key === 'Escape') {
+                    setTitleEditingId(null);
+                  }
+                }}
+              >
+                <input
+                  data-cy="TodoTitleField"
```
volodymyr-soltys97: Move this logic to the `TodoItem` component

src/components/TodoList.tsx
```diff
@@ -0,0 +1,146 @@
...
+          </label>
+          <span data-cy="TodoTitle" className="todo__title">
+            {tempTodo.title}
+          </span>
+          <button type="button" className="todo__remove" data-cy="TodoDelete">
+            ×
+          </button>
+          <TodoLoader loadingTodoIds={loadingTodoIds} todoId={0} />
+        </div>
+      )}
```
volodymyr-soltys97: Move it to the separate component

volodymyr-soltys97 (changes requested):
Almost done!
Let's make your code better

src/components/TodoList.tsx
SerhiiUnhurian: is this really necessary to do?

SerhiiUnhurian: it turns out too much to transfer now, just like that, it was done as needed according to the condition

SerhiiUnhurian:
I'm already doing the next task

SerhiiUnhurian:
done

vadiimvooo (approved):
Well done

PR #1677, opened 2025-03-15
Develop

Beliyandr:
https://beliyandr.github.io/react_todo-app-with-api/

only for audit, and correction for future decompose app
I can't pass all test


volodymyr-soltys97 (changes requested):
Hi, if you need help with failed tests feel free in the fe_chat 
image

Also need to separate the logic into different components.

vadiimvooo (approved):
Well done

PR #1682, opened 2025-03-18
some commit message

tuhusova:
[DEMO LINK](https://tuhusova.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -38,7 +38,7 @@ export const TodoList: React.FC = () => {
   }, []);
 
   useEffect(() => {
-    if (inputRef.current) {
+    if ((error === ErrorType.AddTodo || error === ErrorType.NoTitle || error === ErrorType.DeleteTodo) && inputRef.current) {
       inputRef.current.focus();
     }
   }, [error, newTodoTitle]);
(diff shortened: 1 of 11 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,369 @@
...
+    event.preventDefault();
+    const trimmedTitle = newTodoTitle.trim();
+    if (!trimmedTitle) {
+      setError(ErrorType.NoTitle);
+      setTimeout(() => setError(null), 3000);
+      return;
+    }
+
+    const newTempTodo = {
+      id: -1,
```
volodymyr-soltys97, line 46: Id looks strange

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to fix this failed test 
image
image

volodymyr-soltys97 (changes requested):
.

src/components/TodoList/TodoList.tsx
tuhusova: just for unique id, ok?

vadiimvooo: Why not just use 0

vadiimvooo (approved):
Well done.

src/components/TodoList/TodoList.tsx
tuhusova: When I put 0 this lead to a problem with newTempTodo.

PR #1683, opened 2025-03-21
add task solution

Anton777148:
[DEMO LINK](https://Anton777148.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -3,6 +3,7 @@ import React, { useEffect, useState, useMemo } from 'react';
 import { Todo } from '../types/Todo';
 import { FilterType } from '../types/FilterType';
 import { deleteTodo } from '../api/todos';
+import { ErrorType } from '../types/ErrorType';
 
 type Props = {
   selectedLink: FilterType;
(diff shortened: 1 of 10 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+      <nav className="filter" data-cy="Filter">
+        {Object.values(FilterType).map(type => (
+          <a
+            href="#/"
+            key={type}
+            className={classNames('filter__link', {
+              selected: selectedLink === type,
+            })}
+            data-cy={type === 'All' ? 'FilterLinkAll' : `FilterLink${type}`}
```
VitaliyBondarenko1982:
```suggestion
            data-cy={type === FilterType.All ? 'FilterLinkAll' : `FilterLink${type}`}
```
Use FilterType instead hardcode string.
In geneeral such ternary looks redundant

VitaliyBondarenko1982 (changes requested):
Good job.
Try to fix tests.
Current test failed, because during update todo (click on checkbox) should be visible loader during between request  and response on server.
Loader should be on each todo which on processing on server.

volodymyr-soltys97 (changes requested):
Hi, you need to show loader for each todo in toggle all mode. If you need some help with failed tests feel free in the fe_chat 
image

vadiimvooo (approved):
Well done

PR #1684, opened 2025-03-22
sol

dvdmsk:
- [DEMO LINK](https://dvdmsk.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,26 +1,225 @@
...
+    case FilterType.completed:
+      return filteredTodos.filter(todo => todo.completed);
+    default:
+      return filteredTodos;
+  }
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setLoading] = useState(true);
```
volodymyr-soltys97, line 28:
```suggestion
  const [isLoading, setLoading] = useState(false);
```

src/App.tsx
```diff
@@ -1,26 +1,225 @@
...
+    return todos.filter(todo => !todo.completed).length;
+  };
+
+  useEffect(() => {
+    getTodos()
+      .then(response => {
+        setTodos(response);
+        setLoading(false);
+      })
+      .catch(() => setError('Unable to load todos'));
```
volodymyr-soltys97, line 48: If the request is unsuccessful, the loader state will not change.
```suggestion
    setLoading(true)
    getTodos()
      .then(setTodos)
      .catch(() => setError('Unable to load todos'))
      .finally(() => setLoading(false));
```

volodymyr-soltys97 (approved):

PR #1706, opened 2025-04-01
add solution

PiekharievaK:
[DEMO LINK](https://PiekharievaK.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,118 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+// import { UserWarning } from './UserWarning';
+import React, { useCallback, useEffect, useState } from 'react';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { Header } from './components/Header/Header';
+import { ErrorField } from './components/ErrorField/ErrorField';
+import { FilterBy, ERROR } from './types/enums';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterQwery, setFilterQwery] = useState(FilterBy.All);
+  const [errorMessage, setErrorMessage] = useState<ERROR | string>(
+    ERROR.default,
+  );
+  const [todosLoading, setTodosLoading] = useState<number[]>([]);
+
+  // if (!USER_ID) {
+  //   return <UserWarning />;
+  // }
+
+  const loadTodos = useCallback(async () => {
+    try {
+      setErrorMessage(ERROR.default);
+      const data = await getTodos();
+
+      setTodos(data);
+    } catch (err) {
+      setErrorMessage(ERROR.todos);
+    } finally {
+    }
+  }, []);
+
+  useEffect(() => {
+    setTodosLoading([0]);
+    loadTodos();
+    setTodosLoading([]);
+  }, []);
+
+  const adIdToLoadingList = useCallback((id: number) => {
+    setTodosLoading(prev => {
+      if (prev.length > 0) {
+        return [...prev, id];
+      }
+
+      return [id];
+    });
+  }, []);
+
+  const removeIdFromLoadingList = useCallback((id: number | null) => {
+    setTodosLoading(prev => {
+      if (prev.length > 1 && id !== null) {
+        return prev.filter(item => item !== id);
+      }
+
+      return [];
+    });
+  }, []);
+
+  const filteredTodos = (qwery: FilterBy): Todo[] => {
+    return todos.filter(item => {
+      if (qwery === FilterBy.Active) {
+        return !item.completed;
+      }
+
+      if (qwery === FilterBy.Completed) {
+        return item.completed;
+      }
+
+      return true;
+    });
+  };
+
+  const visibleTodos = filteredTodos(filterQwery);
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">
-          React Todo App - Add and Delete
-        </a>
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+      <div className="todoapp__content">
+        <Header
+          setErrorMessage={setErrorMessage}
+          setTodosLoading={setTodosLoading}
+          setTodos={setTodos}
+          loading={{ adIdToLoadingList, removeIdFromLoadingList }}
+          todos={todos}
+        />
+        {todos.length > 0 && (
+          <TodoList
+            todos={visibleTodos}
+            loading={{ adIdToLoadingList, removeIdFromLoadingList }}
+            setErrorMessage={setErrorMessage}
+            setTodos={setTodos}
+            todosLoading={todosLoading}
+          />
+        )}
+        {/* Hide the footer if there are no todos */}
+        {todos.length > 0 && (
+          <Footer
+            todos={todos}
+            filterQwery={filterQwery}
+            setFilterQwery={setFilterQwery}
+            loading={{ adIdToLoadingList, removeIdFromLoadingList }}
+            setErrorMessage={setErrorMessage}
+            setTodos={setTodos}
+            todosLoading={todosLoading}
+          />
+        )}
+      </div>
+      <ErrorField
+        errorMessage={errorMessage}
+        setErrorMessage={setErrorMessage}
+      />
+    </div>
   );
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,13 +2,15 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 // import { UserWarning } from './UserWarning';
 import React, { useCallback, useEffect, useState, useMemo } from 'react';
-import { getTodos } from './api/todos';
-import { Todo } from './types/Todo';
+
 import { TodoList } from './components/TodoList/TodoList';
 import { Footer } from './components/Footer/Footer';
 import { Header } from './components/Header/Header';
 import { ErrorField } from './components/ErrorField/ErrorField';
+
+import { Todo } from './types/Todo';
 import { FilterBy, ERROR } from './types/enums';
+import { getTodos } from './api/todos';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
@@ -16,28 +18,21 @@ export const App: React.FC = () => {
   const [errorMessage, setErrorMessage] = useState<ERROR | string>(
     ERROR.no_error,
   );
-  const [todosLength, setTodosLength] = useState<number>(0);
+  const [tempTodo, setTempTodo] = useState<{
+    type: 'add' | 'remove';
+    todo?: Todo;
+    todoId?: Todo['id'];
+  } | null>(null);
   const [todosLoading, setTodosLoading] = useState<number[]>([]);
   // if (!USER_ID) {
   //   return <UserWarning />;
   // }
 
-  useEffect(() => {
-    if (
-      (todosLoading.length === 0 && todosLength !== todos.length) ||
-      errorMessage
-    ) {
-      setTodosLength(todos.length);
-    }
-  }, [todosLoading]);
-
   const loadTodos = useCallback(async () => {
     try {
-      setErrorMessage(ERROR.no_error);
       const data = await getTodos();
 
       setTodos(data);
-      setTodosLength(data.length);
     } catch (err) {
       setErrorMessage(ERROR.todos);
     } finally {
@@ -56,8 +51,8 @@ export const App: React.FC = () => {
     setTodosLoading(prev => prev.filter(item => item !== id));
   }, []);
 
-  const filteredTodos = (qwery: FilterBy): Todo[] => {
-    return todos.filter(item => {
+  const filteredTodos = (qwery: FilterBy, list: Todo[]): Todo[] => {
+    return list.filter(item => {
       switch (qwery) {
         case FilterBy.Active:
           return !item.completed;
@@ -69,21 +64,38 @@ export const App: React.FC = () => {
     });
   };
 
-  const visibleTodos = useMemo(
-    () => filteredTodos(filterQwery),
-    [todos, filterQwery],
-  );
+  const visibleTodos = useMemo(() => {
+    let todoList = [...todos];
+
+    if (tempTodo) {
+      switch (tempTodo.type) {
+        case 'add':
+          if (tempTodo.todo) {
+            todoList = [...todoList, tempTodo.todo];
+          }
+
+          break;
+        case 'remove':
+          todoList = todoList.filter(item => item.id !== tempTodo.todoId);
+          break;
+        default:
+      }
+    }
+
+    return filteredTodos(filterQwery, todoList);
+  }, [todos, filterQwery, tempTodo]);
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
       <div className="todoapp__content">
         <Header
-          todosLength={todosLength}
+          todosLength={todos.length}
           setErrorMessage={setErrorMessage}
           setTodosLoading={setTodosLoading}
           setTodos={setTodos}
           loadingIdsState={{ adIdToLoadingList, removeIdFromLoadingList }}
+          setTempTodo={setTempTodo}
           todos={todos}
         />
         {todos.length > 0 && (
@@ -93,6 +105,7 @@ export const App: React.FC = () => {
             setErrorMessage={setErrorMessage}
             setTodos={setTodos}
             todosLoading={todosLoading}
+            setTempTodo={setTempTodo}
           />
         )}
         {todos.length > 0 && (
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,26 +1,118 @@
...
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterQwery, setFilterQwery] = useState(FilterBy.All);
+  const [errorMessage, setErrorMessage] = useState<ERROR | string>(
+    ERROR.default,
+  );
```
fromjkeee16:
```suggestion
  const [errorMessage, setErrorMessage] = useState<ERROR>(
    ERROR.no_error,
  );
```

src/App.tsx
```diff
@@ -1,26 +1,118 @@
...
+  const [todosLoading, setTodosLoading] = useState<number[]>([]);
+
+  // if (!USER_ID) {
+  //   return <UserWarning />;
+  // }
+
+  const loadTodos = useCallback(async () => {
+    try {
+      setErrorMessage(ERROR.default);
```
fromjkeee16: move removing errorMessage to `useEffect` of `ErrorField` component

src/App.tsx
```diff
@@ -1,26 +1,118 @@
...
+      setErrorMessage(ERROR.todos);
+    } finally {
+    }
+  }, []);
+
+  useEffect(() => {
+    setTodosLoading([0]);
+    loadTodos();
+    setTodosLoading([]);
+  }, []);
```
fromjkeee16: u could use state like `isListLoading => true / false`

src/App.tsx
```diff
@@ -1,26 +1,118 @@
...
+
+  const adIdToLoadingList = useCallback((id: number) => {
+    setTodosLoading(prev => {
+      if (prev.length > 0) {
+        return [...prev, id];
+      }
+
+      return [id];
+    });
```
fromjkeee16: it can be simplified, spreading empty array wont cause an error

```suggestion
    setTodosLoading(prev => [...prev, id]);
```

src/App.tsx
```diff
@@ -1,26 +1,118 @@
...
+
+  const removeIdFromLoadingList = useCallback((id: number | null) => {
+    setTodosLoading(prev => {
+      if (prev.length > 1 && id !== null) {
+        return prev.filter(item => item !== id);
+      }
+
+      return [];
+    });
+  }, []);
```
fromjkeee16:
```suggestion
  const removeIdFromLoadingList = useCallback((id: number | null) => {
   setTodosLoading(prev=> prev.filter(item => item !== id));
  }, []);
```

src/App.tsx
```diff
@@ -1,26 +1,118 @@
...
+        return !item.completed;
+      }
+
+      if (qwery === FilterBy.Completed) {
+        return item.completed;
+      }
+
+      return true;
+    });
+  };
```
fromjkeee16: move this function outside of the component as it is re-creating on each re-render and pass it todo list as argument

src/components/ErrorField/ErrorField.tsx
```diff
@@ -0,0 +1,48 @@
...
+    setTimerId(
+      setTimeout(() => {
+        setErrorMessage(ERROR.default);
+      }, 3000),
+    );
+
+    return () => {
+      clearTimeout(timerId);
+    };
+  }, [errorMessage]);
```
fromjkeee16: check the github actions' warning

create `handleClearError` function in App.tsx, wrap it in useCallback and pass it as prop
```javascript
const handleClearError = useCallback(
    () => setErrorMessage(ERROR.no_error),
    [],
  );
  ```

File: src/components/ErrorField/ErrorField.tsx [hunk 8]
```diff
@@ -0,0 +1,48 @@
...
+type Props = {
+  errorMessage: string;
+  setErrorMessage: (message: ERROR) => void;
+};
+
+export const ErrorField: React.FC<Props> = ({
+  errorMessage,
+  setErrorMessage,
+}) => {
+  const [timerId, setTimerId] = useState<NodeJS.Timeout | undefined>();
```
fromjkeee16: i think its better to use `useRef` hook to carry timer id across re-renders without causing another re-render on each value change

src/components/ErrorField/ErrorField.tsx
```diff
@@ -0,0 +1,48 @@
...
+    return () => {
+      clearTimeout(timerId);
+    };
+  }, [errorMessage]);
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={`notification is-danger is-light has-text-weight-normal ErrorNotification ${!errorMessage && 'hidden'}`}
+    >
```
fromjkeee16: use `classNames`
```suggestion
      className={classNames(
          'notification is-danger is-light has-text-weight-normal',
          {
            hidden: !message,
          },
        )}
    >
```

src/App.tsx
```diff
@@ -1,26 +1,118 @@
...
+          setErrorMessage={setErrorMessage}
+          setTodosLoading={setTodosLoading}
+          setTodos={setTodos}
+          loading={{ adIdToLoadingList, removeIdFromLoadingList }}
+          todos={todos}
+        />
+        {todos.length > 0 && (
+          <TodoList
+            todos={visibleTodos}
+            loading={{ adIdToLoadingList, removeIdFromLoadingList }}
```
fromjkeee16: you could create `handler`s functions to handle events like toggling completion status, deleting todos and so on in `App.tsx` and passing those functions as props to your components, this will simplify the code as functions will be in one place & they will have access to your state variables of main component

it will look something like
```javascript
const handleDeleteTodo = async (id: Todo['id']) => {
  // optionally: memorize initial list into a variable to revert the state back to its initial look
  // const initialProcessingList = [...processingIds]
  setProcessingIds(prev => [...prev, id]);

  try {
    await deleteTodo(id);
  } catch {
    setErrorMessage(ERROR.deleting);
    //add `throw new Error(ERROR.deleting)` if needed
  } finally {
    setProcessingIds(prev => prev.filter(item => item !== id));
    // or if initial list was memorized:
    // setProcessingIds(initialProcessingList);
  }
};

```

src/api/todos.ts
```diff
@@ -0,0 +1,24 @@
...
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const changeTodoParams = (
+  todoId: number,
+  param: string,
+  paramValue: Todo['title'] | Todo['completed'],
+) => {
+  return client.patch(`/todos/${todoId}`, { [param]: paramValue });
+};
```
fromjkeee16: you can pass an object with all the properties into the `.patch()` directly
```suggestion
export const changeTodoParams = (
  todoId: Todo['id'],
 newData: Partial<Todo>,
) => {
  return client.patch(`/todos/${todoId}`, newData);
};
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,109 @@
...
+      try {
+        await deleteTodo(item.id);
+        setTodos(prev => prev.filter(todo => todo.id !== item.id));
+      } catch (error) {
+        onError();
+      } finally {
+        loading.removeIdFromLoadingList(item.id);
+      }
+    });
+  };
```
fromjkeee16: u can lift this function into `App` component as:
```javascript
const handleDeleteCompleted = async() => {
const todoIdsToDelete = todos.filter(todo => todo.completed).map(todo=>todo.id);

setProcessingIds(prev => [...prev, ...todoIdsToDelete]

//do the stuff for all items from `todoIdsToDelete`
//it is possible to map the list and return promises and use Promise.allSettled(%promise list from map%) 
// and handle the errored promises
}
```
example (from my code)
```javascript
try {
      const results = await Promise.allSettled(ids.map(id => deleteTodo(id)));

      const successfullyDeleted = ids.filter(
        (_, index) => results[index].status === 'fulfilled',
      );

      setTodos(current =>
        current.filter(todo => !successfullyDeleted.includes(todo.id)),
      );

      const failedToDelete = ids.length - successfullyDeleted.length;

      if (failedToDelete > 0) {
        setErrorMessage(ErrorMessage.FAIL_DELETING);
      }
    } finally {
      setTodoIdsInProcess(current => current.filter(id => !ids.includes(id)));
    }
``` 
but that may be complicated....

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+          return (
+            <a
+              href="#/"
+              key={item}
+              className={`filter__link ${filterQwery === filterValue && 'selected'}`}
+              data-cy={filterValue}
+              onClick={e =>
+                filterChange(e.currentTarget.dataset.cy as FilterBy)
+              }
```
fromjkeee16: why u didnt use `filterValue` in `onClick` if u are passing the same value but as value of attribute in DOM

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,109 @@
...
+};
+
+const areLoaded = (prevProps: Props, nextProps: Props) => {
+  const todosChanged = prevProps.todos !== nextProps.todos;
+  const todosLoadingEmpty = nextProps.todosLoading.length !== 0;
+
+  return todosChanged && todosLoadingEmpty;
+};
+
+export const Footer = React.memo(footerComponent, areLoaded);
```
fromjkeee16: u could wrap each `handle` function into `useCallback` and components into `React.memo` and it would work fine, just remember not to miss dependencies arrays (linter highlights if some required items are not included in it)

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,143 @@
...
+  setErrorMessage,
+  setTodosLoading,
+  setTodos,
+  loading,
+  todos,
+}) => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [completed, setCompleted] = useState(
+    todos.every(item => item.completed === true),
+  );
```
fromjkeee16: you can simplify this to 
```javascript
const completed = todos.every(todo => todo.completed);
```

and it will work because on update request you change the `todos` state which will cause a re-render of all your components that have this `todos` passed as prop and value will be re-calculated on each re-render

fromjkeee16 (changes requested):
it works, but code is hard to read :(

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,137 @@
...
+    todos,
+    setTempTodo,
+  }) => {
+    const [isLoading, setIsLoading] = useState(false);
+    const completed = !todos.every(item => item.completed === true);
+    const inputField = useRef<HTMLInputElement>(null);
+
+    useEffect(() => {
+      inputField.current?.focus();
+    }, [todosLength, isLoading]);
```
fromjkeee16: you are passing the `todos` array, so you can get the length of todos array in-place, there's no need to pass it as a separate prop
```javascript
}, [todos, isLoading]);
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+    useEffect(() => {
+      if (todos?.length < 0) {
+        return;
+      }
+    }, [todos]);
```
fromjkeee16: remove unused code

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+    const addItem = async (event: React.FormEvent<HTMLFormElement>) => {
```
fromjkeee16: it is better to write such functions in root component (`App` in this case) since you are using a lot of its states and functions and passing them down to child components

if you write a function inside of `App` then you will just pass it down as a single prop that will do all the stuff, the only thing that you will need to do is that create a function inside of the component to handle form submit like 
```javascript
const handleFormSubmit = async () => {
  const newTitle = inputField.current?.value.trim() || ''; //this syntax can be used to omit `if` statement
  try {
     await onTodoCreate(newTitle); // function passed down as a prop from `App` which takes the new title
     // validates it, tries to create a todo and throws an error if attempt was unsuccessful
     
     // focus the input on success
     // and the other stuff like `catch() {}`
}
```
so your `<form>` will have something like
```jsx
<form onSubmit={handleFormSubmit}>
```
this is needed to separate logic of functions into smaller ones
your form has a function which handles the logic of form submission
this function calls `onTodoCreate` function which is supposed to send new todo to the server, add the id to array of ids that are being processed, set error message and so on

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,137 @@
...
+          const res = await addTodo(todo);
+
+          setTodos(prev => [...prev, res as Todo]);
+
+          inputField.current.focus();
+          inputField.current.value = '';
+        } catch (err) {
+          const error = err as Error;
+
+          setErrorMessage(error.message || ERROR.add);
```
fromjkeee16:
```javascript
setErrorMessage(ERROR.add);
// add `throw err;` if it is a handler of adding todos
/* throwing an error in the `catch()` block is useful in the 3rd part 
of the task (updateTodo, deleteTodo) as you can track the result of the 
operation in your component with a `try {} catch{}` block (where the task
 is to focus the input on some condition, like: 
*/
 const handleFormSubmit = async (
      event: React.FormEvent<HTMLFormElement>,
    ) => {
      event.preventDefault();

      const newTitle = inputTitleRef.current?.value.trim() || '';

      if (newTitle === todo.title) {
        setIsEditing(false);

        return;
      }

      if (onRename) {
        setIsProcessing(true);
        try {
          await onRename(todo, newTitle);
          setIsEditing(false);
        } finally {
          setIsProcessing(false);
        }
      }
    };
```
where `onRename` is a handler function in `App` passed down as prop:
```javascript
const handleRenameTodo = useCallback(
    async (todoToUpdate: Todo, newTitle: Todo['title']) => {
      if (!newTitle) {
        return handleDeleteTodo(todoToUpdate.id); // this handler uses `throw error` in `catch` block 
        // if there was an error. if not - then it wont interrupt `try{} catch{}` block of components where
        //this handleRenameTodo was called (passed down as prop of a name `onRename`
        // this is an example from 3rd part of the task
      }

      return handleUpdateTodo(todoToUpdate, { title: newTitle }); // the same thing as for update
    },
    [handleDeleteTodo, handleUpdateTodo],
  );
```

fromjkeee16 (approved):
it works. it looks better

 code can be improved further by lifting the logic of todo processing (e.g. : create new todo, update todo, delete todo) to the root component (`App`, App.tsx), removing the need to pass down all the necessary (for that functions) values and setStates

the logic is: pass to the components only the things that will be displayed or will affect the UI (like adding `.is-active` class to the loaders) and functions that will just do all the logic that was implemented before where you only will need to pass them correct arguments and handle their results (if necessary)

src/components/Header/Header.tsx
PiekharievaK: You are right, I don't need to pass a separate prop for this.
This part is left over from the time when I change the list several times for each request on the server, adding a temporary element to it. Now the logic of adding a temporary element has changed and I can remove it.
I will remove the separate variable, but I will leave tracking specifically the length of the todo list in order to get the focus like on mate demo version.
Thanks for notice it.

src/components/Header/Header.tsx
PiekharievaK: Oops, missed that part. Thanks.

src/App.tsx
```diff
@@ -1,26 +1,128 @@
...
+  );
+  const [tempTodo, setTempTodo] = useState<{
+    type: 'add' | 'remove';
+    todo?: Todo;
+    todoId?: Todo['id'];
+  } | null>(null);
+  const [todosLoading, setTodosLoading] = useState<number[]>([]);
+  // if (!USER_ID) {
+  //   return <UserWarning />;
+  // }
```
ArturErmolenko: remove commented

src/App.tsx
```diff
@@ -1,26 +1,128 @@
...
+
+  const loadTodos = useCallback(async () => {
+    try {
+      const data = await getTodos();
+
+      setTodos(data);
+    } catch (err) {
+      setErrorMessage(ERROR.todos);
+    } finally {
+    }
```
ArturErmolenko:
```suggestion
    }
```

src/App.tsx
```diff
@@ -1,26 +1,128 @@
...
+
+  const filteredTodos = (qwery: FilterBy, list: Todo[]): Todo[] => {
+    return list.filter(item => {
+      switch (qwery) {
+        case FilterBy.Active:
+          return !item.completed;
+        case FilterBy.Completed:
+          return item.completed;
+        default:
+          return true;
```
ArturErmolenko: ALL case?

src/App.tsx
```diff
@@ -1,26 +1,128 @@
...
+      }
+    });
+  };
+
+  const visibleTodos = useMemo(() => {
+    let todoList = [...todos];
+
+    if (tempTodo) {
+      switch (tempTodo.type) {
+        case 'add':
```
ArturErmolenko: don't rely on unknown string , create enum

src/App.tsx
```diff
@@ -1,26 +1,128 @@
...
+          if (tempTodo.todo) {
+            todoList = [...todoList, tempTodo.todo];
+          }
+
+          break;
+        case 'remove':
+          todoList = todoList.filter(item => item.id !== tempTodo.todoId);
+          break;
+        default:
```
ArturErmolenko: default should do smthing

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,129 @@
...
+        } finally {
+          setTempTodo(null);
+          setIsLoading(false);
+          setTodosLoading([]);
+        }
+      }
+    };
+
+    const CompleteAll = () => {
```
ArturErmolenko:
```suggestion
    const completeAll = () => {
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,129 @@
...
+  }) => {
+    const [isLoading, setIsLoading] = useState(false);
+    const completed = !todos.every(item => item.completed === true);
+    const inputField = useRef<HTMLInputElement>(null);
+
+    useEffect(() => {
+      inputField.current?.focus();
+    }, [todos.length, isLoading]);
+
+    const addItem = async (event: React.FormEvent<HTMLFormElement>) => {
```
ArturErmolenko: strongly agree with prev comment, move handler to APP
thats how it should work

ArturErmolenko (changes requested):
Almost done 🔥 
Let's try improve a bit

ArturErmolenko (approved):
Well done 👍

PR #1709, opened 2025-04-01
Solution

fromjkeee16:
[DEMO LINK](https://fromjkeee16.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,273 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { Todo } from './types/Todo';
+import { TodoHeader } from './components/Todo/Header/Header';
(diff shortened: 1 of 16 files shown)


mmaryblackk:
Добкин2005

maponomarenko (approved):
вроді працює як треба 👍

src/App.tsx
```diff
@@ -1,26 +1,273 @@
...
+  getTodos,
+  updateTodo,
+  USER_ID,
+} from './api/todos';
+import { NetworkStatus } from './types/AppNetworkStatus';
+import { getInitialFilterFromHash } from './utils/getUrlHash';
 
-const USER_ID = 0;
+const prepareTodos = (list: Todo[], filterBy: FilterOption) => {
+  let copy = [...list];
```
alexander-ignatow: When the filtering is not set you always return a copy.
Why?

alexander-ignatow (approved):
Looks good, but still, I see an issue here

You are overusing memo and useCallback. In some cases it does more harm.

In your case - it makes sense to memo the TodoItem (that is used many times) and no sense in memoing Header and Footer (they are used only once)

Remember - memo is not free
1. It adds some calculations costs and memory usage
2. It adds complexity to the code
3. It requires using hooks, like useCallback, that in turn has it's own cost

PR #1715, opened 2025-04-02
task solution

radykhovska:
[DEMO LINK](https://radykhovska.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
 import { UserWarning } from './UserWarning';
 
@@ -8,8 +8,13 @@ import { Todo } from './types/Todo';
 import { FilteredBy } from './types/filteredBy';
 
 import { errorNotification } from './constants/errors';
-import { deleteTodo, getTodos, USER_ID, updateTodo } from './api/todos';
-import { focusTodoInput } from './helpers/inputFocus';
+import {
+  deleteTodo,
+  getTodos,
+  USER_ID,
+  updateTodo,
+  addTodos,
+} from './api/todos';
 import { filterTodos } from './helpers/filter';
 
 import { FooterTodo } from './components/FooterTodo';
@@ -26,6 +31,8 @@ export const App: React.FC = () => {
   const [isTempTodoCreating, setIsTempTodoCreating] = useState(false);
   const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
 
+  const inputField = useRef<HTMLInputElement>(null);
+
   const handleDelete = async (id: number): Promise<void> => {
     setDeletingTodoIds(prev => [...prev, id]);
 
@@ -37,7 +44,40 @@ export const App: React.FC = () => {
       throw error;
     } finally {
       setDeletingTodoIds(prev => prev.filter(todoId => todoId !== id));
-      focusTodoInput();
+      inputField.current?.focus();
+    }
+  };
+
+  const showTitleError = () => {
+    setErrorMessage(errorNotification.title);
+
+    setTimeout(() => {
+      setErrorMessage('');
+    }, 3000);
+  };
+
+  const handleAddTodo = async (todo: Todo): Promise<boolean> => {
+    const newTodo = {
+      ...todo,
+      userId: USER_ID,
+    };
+
+    setTempTodo(newTodo);
+    setIsTempTodoCreating(true);
+
+    try {
+      const created = await addTodos(newTodo);
+
+      setTodos(prev => [...prev, created]);
+
+      return true;
+    } catch (error) {
+      setErrorMessage(errorNotification.add);
+
+      return false;
+    } finally {
+      setTempTodo(null);
+      setIsTempTodoCreating(false);
     }
   };
 
@@ -130,12 +170,11 @@ export const App: React.FC = () => {
       <div className="todoapp__content">
         <HeaderTodo
           todos={todos}
-          setErrorMessage={setErrorMessage}
-          setTodos={setTodos}
-          setTempTodo={setTempTodo}
+          onInvalidTitle={showTitleError}
           isTempTodoCreating={isTempTodoCreating}
-          setIsTempTodoCreating={setIsTempTodoCreating}
+          inputField={inputField}
           toggleAllTodos={toggleAllTodos}
+          onAddTodo={handleAddTodo}
         />
 
         {!isTodosLoading && (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,170 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+import { Todo } from './types/Todo';
+import { FilteredBy } from './types/filteredBy';
+
+import { errorNotification } from './constants/errors';
+import { deleteTodo, getTodos, USER_ID, updateTodo } from './api/todos';
+import { focusTodoInput } from './helpers/inputFocus';
+import { filterTodos } from './helpers/filter';
+
+import { FooterTodo } from './components/FooterTodo';
+import { HeaderTodo } from './components/HeaderTodo';
+import { TodoList } from './components/TodoList';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isTodosLoading, setIsTodosLoading] = useState(false);
+  const [filteredBy, setFilteredBy] = useState(FilteredBy.ALL);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isTempTodoCreating, setIsTempTodoCreating] = useState(false);
+  const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
+
+  const handleDelete = async (id: number): Promise<void> => {
+    setDeletingTodoIds(prev => [...prev, id]);
+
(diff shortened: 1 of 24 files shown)


glebanty19 (approved):
good job

src/helpers/inputFocus.ts
```diff
@@ -0,0 +1,3 @@
+export const focusTodoInput = () => {
+  document.querySelector<HTMLInputElement>('[data-cy="NewTodoField"]')?.focus();
+};
```
alexander-ignatow: This works, but the react way is preferred

Please, use `useRef` instead

src/components/HeaderTodo/HeaderTodo.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export const HeaderTodo: FC<Props> = ({
+  todos,
+  setErrorMessage,
+  setTodos,
+  setIsTempTodoCreating,
+  setTempTodo,
+  isTempTodoCreating,
+  toggleAllTodos,
```
alexander-ignatow: You have too many props
The component duplicates some functionality and becomes too complex

I would suggest refactor it to have fewer props:
```suggestion
  toggleAllTodos: () => void,
  onAddTodo: (todo: Todo) => void
```

And take away all complex todo-creating logic

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,164 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { FC, useEffect, useRef, useState } from 'react';
+
+interface Props {
+  todo: Todo;
+  isLoading?: boolean;
```
alexander-ignatow: Why optional?

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+interface Props {
+  todos: Todo[];
+  tempTodo: Todo | null;
+  isTempTodoCreating: boolean;
+  handleDelete: (id: number) => Promise<void>;
+  deletingTodoIds: number[];
+  isLoading: boolean;
+  toggleTodo?: (todo: Todo) => Promise<void>;
+  renameTodo?: (todo: Todo) => Promise<void>;
```
alexander-ignatow, line 16: Same as with header, the props and component logic are too complicated

src/components/FooterTodo/FooterTodo.tsx
```diff
@@ -0,0 +1,80 @@
...
+  filteredBy,
+  setFilteredBy,
+  handleDelete,
+}: Props) => {
+  const countActiveTodos = todos.filter(todo => !todo.completed).length;
+  const hasTodosCompleted = todos.some(todo => todo.completed === true);
+
+  const deleteCompleted = () => {
+    todos.filter(todo => todo.completed).forEach(todo => handleDelete(todo.id));
+  };
```
alexander-ignatow, line 24: You should have a props like onDeleteCompleted: () => void

and just call it here

alexander-ignatow (approved):
Overal - your solution is not bad

The main problem is that you split and duplicate the app logic across the components, that makes an App hard to maintain 

In real-life project you would need to move all Todo-manipulation logic to a single place and refactor your components, making them as dumb as possible (see the comments bellow)

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
+  useEffect(() => {
+    setIsTodosLoading(true);
+    setErrorMessage('');
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage(errorNotification.load);
+      })
+      .finally(() => setIsTodosLoading(false));
```
dothearts: Various processing methods used ( await and dotnotation with then)

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
   }
 
+  const filteredTodos = filterTodos(todos, filteredBy);
+
+  const toggleTodo = (updatedTodo: Todo): Promise<void> => {
+    return updateTodo({
+      ...updatedTodo,
+      completed: !updatedTodo.completed,
+    })
+      .then(todoToUpdate => {
```
dothearts: also using then/await in all places

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
+      .finally(() => setIsTodosLoading(false));
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const filteredTodos = filterTodos(todos, filteredBy);
+
+  const toggleTodo = (updatedTodo: Todo): Promise<void> => {
```
dothearts: Also you can create function with try catch wich accept callbacks. 
In this case you remove try/catch duplication in App component. 
Not critical*

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
+
+    updatedTodos
+      .filter((todo, index) => todo.completed !== todos[index].completed)
+      .forEach(todo => {
+        updateTodo(todo).catch(() => {
+          setTodos(prevTodos);
+
+          setErrorMessage(errorNotification.update);
+        });
+      });
```
dothearts, line 154: You can devide this code using variable. 
Chaining in this case bad for code reading (for all case)

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
+          todos={todos}
+          onInvalidTitle={showTitleError}
+          isTempTodoCreating={isTempTodoCreating}
+          inputField={inputField}
+          toggleAllTodos={toggleAllTodos}
+          onAddTodo={handleAddTodo}
+        />
+
+        {!isTodosLoading && (
+          <TodoList
```
dothearts, line 186: Also, you can try move some logic inside todoList ( not App.tsx).

src/components/FooterTodo/FooterTodo.tsx
```diff
@@ -0,0 +1,80 @@
+import classNames from 'classnames';
+import React from 'react';
+import { FilteredBy } from '../../types/filteredBy';
```
dothearts: In future use global imports for paths

src/components/FooterTodo/FooterTodo.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <span className="todo-count" data-cy="TodosCounter">
+        {countActiveTodos} items left
+      </span>
+
+      {/* Active link should have the 'selected' class */}
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={classNames('filter__link', {
+            selected: filteredBy === 'All',
```
dothearts, line 37: Why you don't use enum in this case?

src/components/FooterTodo/FooterTodo.tsx
```diff
@@ -0,0 +1,80 @@
...
+          data-cy="FilterLinkAll"
+          onClick={() => setFilteredBy(FilteredBy.ALL)}
+        >
+          All
+        </a>
+
+        <a
+          href="#/active"
+          className={classNames('filter__link', {
+            selected: filteredBy === 'Active',
```
dothearts, line 48: Also in this case

src/components/FooterTodo/FooterTodo.tsx
```diff
@@ -0,0 +1,80 @@
...
+          data-cy="FilterLinkActive"
+          onClick={() => setFilteredBy(FilteredBy.ACTIVE)}
+        >
+          Active
+        </a>
+
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: filteredBy === 'Completed',
```
dothearts, line 59: And this case

src/components/TodoItem/TodoItem.tsx
dothearts: Loading need to be boolean and required

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,164 @@
...
+  handleDelete?: () => Promise<void>; // 🔧 змінено тип функції
+  isDeleting: boolean;
+  onToggle?: () => Promise<void>;
+  onRename?: (title: string) => Promise<void>;
```
dothearts, line 12: Also optional function is not good idea. 
In which case do we not have these methods?

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,164 @@
...
+  isDeleting,
+  onToggle,
+  onRename,
+}: Props) => {
+  const [isToggleLoading, setIsToggleLoading] = useState(false);
+  const [isFormActive, setIsFormActive] = useState(false);
+  const [isLoadingChanges, setIsLoadingChanges] = useState(false);
+  const [enterTriggeredSave, setEnterTriggeredSave] = useState(false);
+  const [title, setTitle] = useState(todo.title);
+  const inputRef = useRef<HTMLInputElement | null>(null);
```
dothearts, line 28: Too many states. Need to check if we really need them.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,164 @@
...
+      }
+
+      setIsFormActive(false);
+    } catch {
+      focusInput();
+    } finally {
+      setIsLoadingChanges(false);
+      setEnterTriggeredSave(false);
+    }
+  };
```
dothearts, line 83: Try to think how we can optimaize this code for reading and better understanding. Also you need use one of way in work with promise async/await or than/catch notation

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,164 @@
...
+      ) : (
+        <>
+          <span data-cy="TodoTitle" className="todo__title">
+            {todo.title}
+          </span>
+          <button
+            type="button"
+            className="todo__remove"
+            data-cy="TodoDelete"
+            onClick={handleDelete} // 🔧 передаємо вже обгорнуту функцію
```
dothearts, line 145: comment

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,164 @@
...
+            ×
+          </button>
+        </>
+      )}
+
+      <div
+        data-cy="TodoLoader"
+        className={classNames('modal overlay', {
+          'is-active':
+            isLoading || isToggleLoading || isDeleting || isLoadingChanges,
```
dothearts, line 156: can write const for it

DanilWeda (changes requested):

PR #1720, opened 2025-04-04
add task solution

nataliaso23:
[DEMO LINK] (https://nataliaso23.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,3 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
 import React, {
   useCallback,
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,26 +1,194 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
raulriato: Your code should work with the eslint. Do not disable it.

raulriato (changes requested):
Your code is nice, but you should work with eslint, not disable it.

raulriato (approved):
Well done!

PR #1722, opened 2025-04-04
Dev

Iwantschenko:
[DEMO LINK](https://Iwantschenko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,192 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+import { useEffect, useRef, useState } from 'react';
 
-export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+import './styles/todoapp.scss';
+
+import { Todo } from './types/Todo';
+
+import { ErrorNotification } from './components/ErrorNotification';
+import { TodoList } from './components/TodoList';
+import { TodoFilter } from './components/TodoFilter';
+import { FilterType } from './types/FilterType';
+import { todosService, USER_ID } from './api/todos';
+
+import { TodoItem } from './components/TodoItem';
+import { ErrorMessages } from './types/ErrorMessages';
+import classNames from 'classnames';
+import { TodoInput } from './components/TodoInput/TodoInput';
+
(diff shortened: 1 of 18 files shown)


maxbilfed (approved):
Seems like a good work

src/App.tsx
```diff
@@ -1,26 +1,276 @@
...
+    return todoList.filter(todo => !todo.completed).length;
+  };
+
+  const hasComplitedTodos = () => {
+    if (todoList.length === 0) {
+      return true;
+    }
+
+    return todoList.some(todo => todo.completed);
+  };
```
alexander-ignatow, line 53: Minor - but it makes sense to useMemo on these values

src/App.tsx
```diff
@@ -1,26 +1,276 @@
...
+      setErrorMessage(ErrorMessages.updateError);
+      throw new Error();
+    } finally {
+      setTodosToLoading([]);
+    }
+  };
+
+  const isToggleAll = () => {
+    return todoList.every(todo => todo.completed);
+  };
```
alexander-ignatow, line 166: Mind the naming

For instance: 
```suggestion
  const areAllTodosCompleted = () => {
    return todoList.every(todo => todo.completed);
  };
```

etc.

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,45 @@
...
+    if (!errorMessage) {
+      return;
+    }
+
+    const timer = setTimeout(() => {
+      removeError();
+    }, 2000);
+
+    return () => clearTimeout(timer);
+  }, [errorMessage, removeError]);
```
alexander-ignatow, line 26: Remove error is a dependency for "remove message" hook
So you need to make sure it is used with the useCallback hook (not to trigger the effect)

Though in your case it is probably can't be reproduced

src/components/TodoInput/TodoInput.tsx
```diff
@@ -0,0 +1,64 @@
+import { useEffect, useRef, useState } from 'react';
+import { ErrorMessages } from '../../types/ErrorMessages';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  onAddTodo: (title: string) => Promise<Todo | null>;
+  setCurrentInputElement: (input: HTMLInputElement | null) => void;
+  setErrorMessage: (message: ErrorMessages) => void;
```
alexander-ignatow, line 8: I would call it "onError" - ust to use more generic name

alexander-ignatow (approved):
Good job! 

PLease consider the comments for the future

PR #1738, opened 2025-04-18
solution without checking conciseness of code

NadiiaBulmak:
[DEMO LINK](https://NadiiaBulmak.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,26 +1,303 @@
...
-  if (!USER_ID) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+  const [newTodoTitle, setTodoTitle] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [showError, setShowError] = useState(false);
+  const [loading, setLoading] = useState(true);
+  const [filteredBy, setFilteredBy] = useState(Filter.All);
+  const [todosCounter, setTodosCounter] = useState(0);
```
etojeDenys: looks redundant

etojeDenys (approved):

src/App.tsx
NadiiaBulmak: You're absolutely right — `todosCounter` does look redundant, I can calculate it with useMemo, so I’ll remove that.
As for this state(17 line), I decided to keep it because when I reset the error, a small empty error window appears for a few milliseconds. That’s why I use showError to control the visibility more precisely.

PR #1743, opened 2025-04-29
solution

VitoorMF:
[DEMO LINK](https://VitoorMF.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,26 +1,503 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
raulriato: It's not a good practice to disable eslint.

Make your code match it, instead of disabling.

raulriato (changes requested):
Your code works. But you should not disable eslint

raulriato (changes requested):
Forgot to request changes

pedro-ruas (approved):

PR #1757, opened 2025-05-10
Develop

ruslan-ihnatenko:
[DEMO LINK](https://ruslan-ihnatenko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,26 +13,19 @@ import { ToDoItem } from './components/ToDoItem';
 export const App: React.FC = () => {
   // #region loadToDOs
   const [todos, setToDos] = useState<Todo[]>([]);
-  const [, setLoading] = useState<boolean>(false);
-  const [loadingTodoIds, setloadingTodoIds] = useState<number[]>([]);
+  const [loadingTodoId, setLoadingTodoId] = useState<number | null>(null);
   const [errorMessage, setErrorMessage] = useState('');
   const [filter, setFilter] = useState<Filter>(Filter.All);
   const [tempToDo, setTempToDo] = useState<Todo | null>(null);
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
 
   const inputRef = useRef<HTMLInputElement>(null);
 
-  const loadToDos = () => {
-    setLoading(true);
-
+  useEffect(() => {
     postService
       .getTodos(USER_ID)
       .then(setToDos)
-      .catch(() => setErrorMessage('Unable to load todos'))
-      .finally(() => setLoading(false));
-  };
-
-  useEffect(() => {
-    loadToDos();
+      .catch(() => setErrorMessage('Unable to load todos'));
   }, []);
 
   useEffect(() => {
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,26 +1,229 @@
...
+import { Filter } from './types/Filter';
+import { Header } from './components/Header';
+import { TempToDo } from './components/TempToDo';
+import { Footer } from './components/Footer';
+import { ToDoItem } from './components/ToDoItem';
 
 export const App: React.FC = () => {
+  // #region loadToDOs
+  const [todos, setToDos] = useState<Todo[]>([]);
+  const [, setLoading] = useState<boolean>(false);
```
VitaliyBondarenko1982: During update or delete multiply todos loader should be on each todo which in processing on server.
Instead boolean state use state whick will contain array of todo ids, whick now in processing on server.

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comment and link to [DEMO LINK](https://ruslan-ihnatenko.github.io/react_todo-app-with-api/) in PR description (now not works).

src/App.tsx
```diff
@@ -1,26 +1,228 @@
...
```
TarasHoliuk: This state is redundant - you don't use its value. Just delete it

src/App.tsx
```diff
@@ -1,26 +1,228 @@
...
+  const [loadingTodoIds, setloadingTodoIds] = useState<number[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState<Filter>(Filter.All);
+  const [tempToDo, setTempToDo] = useState<Todo | null>(null);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const loadToDos = () => {
```
TarasHoliuk: In fact,  once you remove `setLoading` from this function you can entirely remove the function and just do this inside of useEffect callback:

```
postService
      .getTodos(USER_ID)
      .then(setToDos)
      .catch(() => setErrorMessage('Unable to load todos'))
      .finally(() => setLoading(false));
```

src/App.tsx
```diff
@@ -1,26 +1,228 @@
...
+        title: newTodo.title,
+        completed: newTodo.completed,
+        userId: newTodo.userId,
+      });
+
+      setToDos(currentTodos => [...currentTodos, createdTodo]);
+      setTempToDo(null);
+    } catch (error) {
+      setErrorMessage('Unable to add a todo');
+      setTempToDo(null);
```
TarasHoliuk, line 76: do it in `finally` block

src/App.tsx
```diff
@@ -1,26 +1,228 @@
...
+      throw error;
```
TarasHoliuk, line 77: No need to throw it again and catch the second time to handle in **_the same_** way in ToDoForm:

Знімок екрана 2025-05-11 о 13 13 37

src/App.tsx
```diff
@@ -1,26 +1,228 @@
...
+        ...todoToUpdate,
+        ...updatedFields,
+      });
+
+      setToDos(currentTodos =>
+        currentTodos.map(todo => (todo.id === todoId ? updatedTodo : todo)),
+      );
+    } catch (error) {
+      setErrorMessage('Unable to update a todo');
+      throw error;
```
TarasHoliuk, line 137: Same - do not throw error again

TarasHoliuk (changes requested):
Great job! Added a few small comments to improve your code 🙂

src/App.tsx
ruslan-ihnatenko: There i throw error to refocuse back to the input field when we failed on todo updating submit. If you can advise me how to handle that without error throwing - i would be glad for that.

volodymyr-soltys97 (approved):
Great! 🔥

PR #1758, opened 2025-05-12
tests done

DrDakka:
Unfortunately, I am unable to deploy my solution due to some technical issues


DrDakka:
deploy complete

DrDakka:
https://drdakka.github.io/react_todo-app-with-api/

src/components/ErrorMsg/ErrorMsg.tsx
```diff
@@ -0,0 +1,43 @@
...
+    if (error === '') {
+      return;
+    }
+
+    const timer = setTimeout(() => {
+      setError('');
+    }, 3000);
+
+    return () => clearTimeout(timer);
+  }, [error]);
```
volodymyr-soltys97, line 20: Fix it 
image

volodymyr-soltys97 (changes requested):
Hi, need to fix all comments as in this pull request 
https://github.com/mate-academy/react_todo-app-add-and-delete/pull/1747

src/components/ErrorMsg/ErrorMsg.tsx
DrDakka: I've committed changes you've highlighted.
About useEffect dependencies: same answer, I'll just duplicate it here:
"It will cause setting timeout twice. Error here is a state and it is being pushed to the element by props right from the hooks file. The dependencies are correct, no need to use a setError as dependency here. Also, according to https://legacy.reactjs.org/docs/hooks-reference.html#usestate state "React guarantees that setState function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the useEffect or useCallback dependency list.""

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,55 @@
...
+  clear: (event: React.MouseEvent<HTMLButtonElement>) => Promise<void>;
+};
+
+export const Footer: React.FC<Props> = ({
+  filter,
+  setFilter,
+  todos,
+  clear,
+}) => {
+  const completed = todos.filter(todo => todo.completed === true);
```
volodymyr-soltys97, line 18:
```suggestion
  const completed = todos.filter(todo => todo.completed);
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,137 @@
...
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={completed}
+          onClick={handleCheckbox}
+        />
+      </label>
+
+      {editFlag === false ? (
```
volodymyr-soltys97, line 91:
```suggestion
      {!editFlag ? (
```

volodymyr-soltys97 (approved):
GJ! 👍

PR #1760, opened 2025-05-17
React_todo-app-with-api

IOherhi:
[DEMO LINK](https://IOherhi.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,308 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+/* eslint-disable max-len */
 
-const USER_ID = 0;
+// #region Components
+
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { TodoList } from './components/TodoList';
+import { MessageError } from './components/MessageError';
+
+// #endregion
+
+// #region allImport
+
+import { Todo } from './types/Todo';
+import { getTodos, postTodos, deleteTodos, patchTodos } from './api/todos';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
+
+// #endregion
+
+export type TypeNewTodo = {
+  id: null;
+  userId: number;
+  title: string;
+  completed: boolean;
+};
+
+const USER_ID = 2881;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  // #region AddFocus
+  const inputRef = useRef<HTMLInputElement>(null);
+  const useRefTimer = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  const focusForInput = () => {
+    inputRef.current?.focus();
+  };
+  // #endregion
+
+  // #region AddLoader
+  const [LoderId, setLoderId] = useState<number | null>(null);
+
+  const showLoader = (id: number) => {
+    setLoderId(id);
+  };
+  // #endregion
+
+  // #region State
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState(false);
+  const [theError, setTheError] = useState<string>('');
+  const [inputValue, setInputValue] = useState<string>('');
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
+  const [NewTodo, setCreateNewTodo] = useState<TypeNewTodo | null>(null);
+  const [doDisable, setDoDisable] = useState(false);
+
+  // #endregion
+
+  // #region ShowErro
+
+  const errorTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  const showError = (message: string) => {
+    setHasError(true);
+    setTheError(message);
+
+    if (errorTimerRef.current) {
+      clearTimeout(errorTimerRef.current);
+    }
+
+    errorTimerRef.current = setTimeout(() => {
+      setHasError(false);
+    }, 3000);
+  };
+
+  // #endregion
+
+  // #region DoFetch
+
+  const doPost = (titleValue: string) => {
+    if (titleValue.trim().length === 0) {
+      showError('Title should not be empty');
+
+      return;
+    }
+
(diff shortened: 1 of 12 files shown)


etojeDenys (changes requested):
1. please fix all tests or feel free to ask for some help in the chat
image

2. toggle all button doesn't work correctly
3. clear completed delete todos without loaders
4. fix styles on edit input
image

5. if edit todo is empty delete todo without this message
image

src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
+
+  // #region State
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState(false);
+  const [theError, setTheError] = useState<string>('');
+  const [inputValue, setInputValue] = useState<string>('');
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
+  const [NewTodo, setCreateNewTodo] = useState<TypeNewTodo | null>(null);
+  const [doDisable, setDoDisable] = useState(false);
```
VitaliyBondarenko1982, line 57:
```suggestion
  const [newTodo, setNewTodo] = useState<Todo | null>(null);
```
use camelCase for variables naming, and general convention for naming variables which returns useState (state, setState)

doDisable looks redundant. use newTodo state for it.

TypeNewTodo redundant type. Should be Todo

src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
+  const showLoader = (id: number) => {
+    setLoderId(id);
+  };
+  // #endregion
+
+  // #region State
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState(false);
+  const [theError, setTheError] = useState<string>('');
```
VitaliyBondarenko1982, line 53: why do you need to states for error?
enough  theError (better to rename errror) state.

src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
+    errorTimerRef.current = setTimeout(() => {
+      setHasError(false);
+    }, 3000);
+  };
+
+  // #endregion
+
+  // #region DoFetch
+
+  const doPost = (titleValue: string) => {
```
VitaliyBondarenko1982, line 82:
```suggestion
  const addTodo = (titleValue: string) => {
```
addTodo or createTodo will be better naming. we work with todo entity, not with post.

src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
+      showError('Unable to update a todo');
+      throw new Error('Update failed');
+    } finally {
+      setShowLoaderCurrentPatch(p => p.filter(currentId => currentId !== id));
+    }
+  };
+
+  // #endregion
+
+  const deletePost = (id: number) => {
```
VitaliyBondarenko1982, line 153:
```suggestion
  const deleteTodo = (id: number) => {
```

src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
+      .then(() => {
+        setTodos(todos.filter(todo => todo.id !== id));
+      })
+      .catch(() => {
+        showError('Unable to delete a todo');
+      })
+      .finally(() => focusForInput());
+  };
+
+  const deleteCompletedPost = (completedTodos: Todo[]) => {
```
VitaliyBondarenko1982, line 164:
```suggestion
  const deleteCompletedTodos = (completedTodos: Todo[]) => {
```

src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
+
+  const doPost = (titleValue: string) => {
+    if (titleValue.trim().length === 0) {
+      showError('Title should not be empty');
+
+      return;
+    }
+
+    const createNewTodo: TypeNewTodo = {
+      id: null,
```
VitaliyBondarenko1982, line 90:
```suggestion
      id: 0,
```
better to make consistent types for todo

src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
+    return todos.filter(todo => {
+      if (filter === 'active') {
+        return !todo.completed;
+      }
+
+      if (filter === 'completed') {
+        return todo.completed;
+      }
+
+      return true;
```
VitaliyBondarenko1982, line 251: here will be redundant array iteration for case 'all'
you can return todos immediately in useMemo

src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
+        // #endregion AddFocus
+      });
+  };
+
+  // #region UpdatePost
+
+  /* eslint-disable @typescript-eslint/indent */
+  const [showLoaderCurrentPatch, setShowLoaderCurrentPatch] = useState<
+    number[]
+  >([]);
```
VitaliyBondarenko1982, line 128: keep all states at the top of component

src/components/TodoList.tsx
```diff
@@ -0,0 +1,184 @@
...
+  NewTodo,
+  LoderId,
+  showLoader,
+  toggleAll,
+  uppdatePost,
+  deletePost,
+}) => {
+  const [isEditing, setIsEditing] = useState<number>(-1);
+  const [editedTitle, setEditedTitle] = useState<string>('');
+
```
VitaliyBondarenko1982, line 35: isEditing naming means that expect boolean type.
about isEditing or not and abotu editedTitle should now TodoItem component (you should make it and keep all logic which related only for todo in that component)

src/components/TodoList.tsx
```diff
@@ -0,0 +1,184 @@
...
+                    if (newValue !== todo.title) {
+                      try {
+                        await uppdatePost(newValue, todo.id);
+                        setIsEditing(-1); // ← тільки якщо успішно
+                      } catch (error) {
+                        throw new Error('Update failed');
+                      }
+                    } else {
+                      setIsEditing(-1); // нічого не змінилось — закриваємо
+                    }
```
VitaliyBondarenko1982, line 100: it is bad practice to create anonymous arrow function in jsx.
make it in component and pass reference here.
check other places.

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest to improve code and make it cleaner and understandable.

vadiimvooo (changes requested):
Looks like you forgot to push last changes. Please feel free to ask for help in fe_chat.

anastasiiavorobiova (changes requested):
Make sure to fix previous comments

PR #1784, opened 2025-06-15
Develop

keex1k:
[DEMO LINK](https://keex1k.github.io/react_todo-app-with-api/)


src/UserTodosList.tsx
```diff
@@ -0,0 +1,72 @@
...
+          >
+            ×
+          </button>
+
+          <div
+            data-cy="TodoLoader"
+            className={classNames('modal overlay', {
+              'is-active': isLoading || isDeleting,
+            })}
+          >
```
natalia-klonowska: loader depends only on a boolean value which causes all todos to show loader when just one's being deleted:
image
Instead you should maintain array of id for all todos currently being deleted. This will also allow you to delete multiple todos at the same time

src/App.tsx
```diff
@@ -1,26 +1,248 @@
...
+
+        <section className="todoapp__main" data-cy="TodoList">
+          <UserTodosList
+            todos={finalTodos}
+            //onChecked={checkTodo}
+            onDeleted={delTodo}
+            isDeleting={isDeleting}
+          />
+          {isCreating && (
+            <UserTodosList
```
natalia-klonowska: Create TodoItem component instead of using UserTodosList twice. Also it should be based on presence of tempTodo rather than value of isCreating:
"show an independent TodoItem after the list if tempTodo is not null"

src/App.tsx
```diff
@@ -1,26 +1,248 @@
...
+        })
+        .catch(() => {
+          setError('add');
+          setTempTodo({ id: 0, userId: USER_ID, title: '', completed: false });
+          setTimeout(() => inputRef.current?.focus(), 0);
+        })
+        .finally(() => {
+          setIsCreating(false);
+        });
+    }
```
natalia-klonowska: set tempTodo before sending request instead of using setIsCreating

natalia-klonowska (changes requested):

keex1k:
could someone help me with this error?

src/App.tsx
```diff
@@ -1,26 +1,352 @@
...
+
+    setTempTodo({
+      id: 0,
+      title: title.trim(),
+      completed: false,
+      userId: USER_ID,
+    });
+
+    addTodos({
+      id: 0,
```
zibi-dev, line 83: You should not send id to the BE. It probably creates unique id

src/App.tsx
```diff
@@ -1,26 +1,352 @@
...
+    getTodos()
+      .then(data => {
+        setTodos(data ?? null);
+      })
+      .catch(() => setError('load'));
+  }, []);
+
+  useEffect(() => {
+    inputRef.current?.focus();
+  }, [todos]);
```
zibi-dev: I see that you are calling focus in handler function. Do you need that effect then?

src/ErrorMessage.tsx
```diff
@@ -0,0 +1,51 @@
...
+      errorMessage = 'Unable to load todos';
+      break;
+    case 'add':
+      errorMessage = 'Unable to add a todo';
+      break;
+    case 'delete':
+      errorMessage = 'Unable to delete a todo';
+      break;
+    case 'update':
+      errorMessage = 'Unable to update a todo';
```
zibi-dev, line 26: You should create interface with errors. Enums or object that keeps the case as key and as value description. And use it in error state also.

src/App.tsx
```diff
@@ -1,26 +1,352 @@
...
+import { TodoList } from './TodoList';
+import classNames from 'classnames';
 
-const USER_ID = 0;
+export type TodoInput = Omit<Todo, 'id'>;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [title, setTitle] = useState<string>('');
+  const [error, setError] = useState<string>('');
```
zibi-dev, line 21: It should be typed

src/App.tsx
```diff
@@ -1,26 +1,352 @@
...
+import { Footer } from './Footer';
+import { TodoList } from './TodoList';
+import classNames from 'classnames';
 
-const USER_ID = 0;
+export type TodoInput = Omit<Todo, 'id'>;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [title, setTitle] = useState<string>('');
```
zibi-dev, line 20: Putting state that changes every time you type, really high up in your app's structure, can make it slow.

src/App.tsx
```diff
@@ -1,26 +1,352 @@
...
+  const [error, setError] = useState<string>('');
+  const [filter, setFilter] = useState<string>('all');
```
zibi-dev, line 22: This one also should be typed

src/App.tsx
```diff
@@ -1,26 +1,352 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [loadingIds, setLoadingIds] = useState<number[]>([]);
```
zibi-dev, line 24: Are you sure you need both of these states?

src/App.tsx
```diff
@@ -1,26 +1,352 @@
...
+
+    addTodos({
+      id: 0,
+      title: title.trim(),
+      completed: false,
+      userId: USER_ID,
+    })
+      .then(newTodoFromAPI => {
+        setTodos(prev => (prev ? [...prev, newTodoFromAPI] : [newTodoFromAPI]));
+        setTimeout(() => inputRef.current?.focus(), 0);
```
zibi-dev: is that setTimeout necessarry?

src/App.tsx
```diff
@@ -1,26 +1,352 @@
...
+          if (!prev) {
+            return null;
+          }
+
+          return prev.map(todo => ({
+            ...todo,
+            completed: newCompletedStatus,
+          }));
+        });
+        todos.map(todo => {
```
zibi-dev, line 249: If it happens in both then and catch then maybe move it into finally

src/App.tsx
```diff
@@ -1,26 +1,352 @@
...
+
+          return prev.map(todo =>
+            todo.id === todoId ? { ...todo, title: newTitle } : todo,
+          );
+        });
+        setLoadingIds(prev => prev.filter(id => id !== todoId));
+      })
+      .catch(() => {
+        setError('update');
+        setLoadingIds(prev => prev.filter(id => id !== todoId));
```
zibi-dev: Again the same action in both cases -> move it to finally block

Zibi95 (changes requested):

danon321 (approved):
git jest

PR #1800, opened 2025-06-17
add solution

yelyzavetaz:
[DEMO LINK](https://yelyzavetaz.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/components/TodoItem/TodoItem.tsx
+++ b/src/components/TodoItem/TodoItem.tsx
@@ -10,6 +10,7 @@ import { Todo } from '../../types/Todo';
 import cn from 'classnames';
 import { client } from '../../utils/fetchClient';
 import { ErrorMessage } from '../../types/ErrorStatusType';
+import { deleteTodo } from '../../api/todos';
 
 type TodoItemProps = {
   todo: Todo;
@@ -201,8 +202,7 @@ export const TodoItem: React.FC<TodoItemProps> = ({
       setTodoIdsToDelete([...todoIdsToDelete, todo.id]);
     }
 
-    client
-      .delete(`/todos/${todo.id}`)
+    deleteTodo(todo.id)
       .then(() => {
         if (todos && setTodos) {
           setTodos(todos.filter(todoItem => todoItem.id !== todo.id));
--- a/src/components/TodoItem/TodoItem.tsx
+++ b/src/components/TodoItem/TodoItem.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
-import React, { ChangeEvent } from 'react';
+import React, { ChangeEvent, FormEvent, useState } from 'react';
 import { Todo } from '../../types/Todo';
 import cn from 'classnames';
 import { client } from '../../utils/fetchClient';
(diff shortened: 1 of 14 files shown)


jasialazyrka:
Супер!

jasialazyrka (approved):
Молодець!

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,321 @@
...
+      return;
+    }
+
+    if (value.length === 0) {
+      if (setTodoIdsToDelete && todoIdsToDelete) {
+        setTodoIdsToDelete([...todoIdsToDelete, todo.id]);
+      }
+
+      client
+        .delete(`/todos/${todo.id}`)
```
ArturErmolenko: use method from api

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,321 @@
...
+      setIsEditing(false);
+      if (setTodoIdsToUpdate) {
+        setTodoIdsToUpdate([]);
+      }
+
+      return;
+    }
+
+    if (value.length === 0) {
+      if (setTodoIdsToDelete && todoIdsToDelete) {
```
ArturErmolenko: can we get rid of nested if?

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,321 @@
...
+        setTodoIdsToDelete([...todoIdsToDelete, todo.id]);
+      }
+
+      client
+        .delete(`/todos/${todo.id}`)
```
ArturErmolenko: we should use routes like this

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,321 @@
...
+
+      return;
+    }
+
+    if (setTodoIdsToUpdate && todoIdsToUpdate) {
+      setTodoIdsToUpdate([...todoIdsToUpdate, todo.id]);
+    }
+
+    client
+      .patch<Todo>(`/todos/${todo.id}`, { title: value.trim() })
```
ArturErmolenko: you have api file use it

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,50 @@
...
+            setErrorMessage={setErrorMessage}
+            todoIdsToDelete={todoIdsToDelete}
+            isTempTodo={false}
+            todoIdsToUpdate={todoIdsToUpdate}
+            setTodoIdsToDelete={setTodoIdsToDelete}
+            setTodoIdsToUpdate={setTodoIdsToUpdate}
+          />
+        );
+      })}
+      {tempTodo && <TodoItem todo={tempTodo} isTempTodo={true} />}
```
ArturErmolenko:
```suggestion
      {tempTodo && <TodoItem todo={tempTodo} isTempTodo />}
```

src/hooks/useTodo.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ErrorMessage } from '../types/ErrorStatusType';
+import { getVisibleTodos } from '../utils/getVisibleTodos';
+
+export function useTodo() {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<ErrorMessage | null>(null);
+  const [filterStatus, setFilterStatus] = useState(FilterStatusType.All);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todoIdsToDelete, setTodoIdsToDelete] = useState<number[]>([]);
+  const [todoIdsToUpdate, setTodoIdsToUpdate] = useState<number[]>([]);
```
ArturErmolenko, line 14: what difference between this states? i believe we should use just one state like processingTodoIds or smth like this
Let's simplify
KISS DRY ;)

src/hooks/useTodo.ts
```diff
@@ -0,0 +1,50 @@
...
+  useEffect(() => {
+    getTodos()
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+        setErrorMessage(null);
+      })
+      .catch(() => {
+        setErrorMessage(ErrorMessage.LoadTodos);
+      });
+  }, []);
```
ArturErmolenko, line 31: where is loading state? will be great to use only one loading state too

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,321 @@
...
+        if (setTodoIdsToUpdate && todoIdsToUpdate) {
+          setTodoIdsToUpdate(todoIdsToUpdate.filter(id => id !== todo.id));
+        }
+      });
+  };
+
+  const isLoading =
+    todoIdsToUpdate?.includes(todo.id) ||
+    todoIdsToDelete?.includes(todo.id) ||
+    isTempTodo;
```
ArturErmolenko, line 176: we should manage loading on every async operation

ArturErmolenko (changes requested):
Nice 👍 
Let's improve a bit

src/hooks/useTodo.ts
yelyzavetaz: I followed the Single Responsibility Principle, which is why I have two separate states for two different purposes. If I merged them into a single state, the user might encounter issues with the todo list if the API takes too long to process requests. For example, what if the user edits a todo title and immediately deletes another todo while the API is still processing the editing action?

src/components/TodoItem/TodoItem.tsx
yelyzavetaz: I believe they’re already managing the loading state for each async operation. I implemented it this way because I need to track the loading state for each individual item. For example, if item 1 has finished updating but item 2 is still being processed, the loading should only be shown for item 2.

Знімок екрана 2025-06-20 161744
Знімок екрана 2025-06-20 161759

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+    if (value === todo.title) {
+      setIsEditing(false);
+      setTodoIdsToUpdate?.([]);
+
+      return;
+    }
+
+    if (value.length === 0) {
+      setTodoIdsToDelete?.([...(todoIdsToDelete || []), todo.id]);
```
BudnikOleksii, line 63: When you need to update state depending on the previous one, use functional `setState`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+      deleteTodo(todo.id)
+        .then(() => {
+          if (todos && setTodos) {
+            setTodos?.(todos.filter(todoItem => todoItem.id !== todo.id));
```
BudnikOleksii, line 68: The same, fix in all places

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,298 @@
...
+      .finally(() => {
+        setTodoIdsToUpdate?.(
+          (todoIdsToUpdate || []).filter(id => id !== todo.id),
+        );
+      });
+  };
+
+  const isLoading =
+    todoIdsToUpdate?.includes(todo.id) ||
+    todoIdsToDelete?.includes(todo.id) ||
```
BudnikOleksii, line 175: One state for both will be enough

src/components/ToggleAll/ToggleAll.tsx
```diff
@@ -0,0 +1,119 @@
...
+              setErrorMessage(ErrorMessage.UpdateTodos);
+            }
+          }
+
+          if (todos) {
+            const newTodos = todos.map(currentTodo => {
+              const todosToUpdate = updatedTodos.find(
+                updatedTodo => updatedTodo.id === currentTodo.id,
+              );
+
```
BudnikOleksii, line 53: Could be simplified, you do almost the same in both cases

src/hooks/useTodo.ts
BudnikOleksii: > I followed the Single Responsibility Principle, which is why I have two separate states for two different purposes. If I merged them into a single state, the user might encounter issues with the todo list if the API takes too long to process requests. For example, what if the user edits a todo title and immediately deletes another todo while the API is still processing the editing action?

You don't violate SRP if you use the same state, because they are responsible only for one thing - loading state.
I'm not sure if understand correct. But if the user edits a todo title and immediately deletes another todo you should show loaders for both by checking if their id's are in an array.

And after operation remove only corresponding loaders.
```
setLoadingTodoIds(prev => prev.filter(id => (
  id !== todo.id // filter only todos were are updated/deleted
)));
```

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments

PR #1816, opened 2025-06-23
third part solution

solyaqw:
[DEMO LINK](https://solyaqw.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import classNames from 'classnames';
 
 import { UserWarning } from './UserWarning';
@@ -27,15 +27,16 @@ export const App: React.FC = () => {
   const [todoValue, setTodoValue] = useState<string>('');
   const [isAdding, setIsAdding] = useState<boolean>(false);
 
-  const [deletingTodos, setDeletingTodos] = useState<number[]>([]);
-  const [updatingTodos, setUpdatingTodos] = useState<number[]>([]);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+
+  const inputRef = useRef<HTMLInputElement>(null);
 
   const filteredTodos = todos.filter(todo => {
-    if (filter === 'active') {
+    if (filter === StatusFilter.Active) {
       return !todo.completed;
     }
 
-    if (filter === 'completed') {
+    if (filter === StatusFilter.Completed) {
       return todo.completed;
     }
 
@@ -47,7 +48,7 @@ export const App: React.FC = () => {
 
     const trimmedValue = todoValue.trim();
 
-    if (trimmedValue === '') {
+    if (!trimmedValue) {
       setError('Title should not be empty');
 
       return;
@@ -62,6 +63,8 @@ export const App: React.FC = () => {
       userId: USER_ID,
     });
 
+    setLoadingTodoIds(prev => [...prev, 0]);
+
     createTodo(trimmedValue)
       .then(newTodo => {
         setTodos([...todos, newTodo]);
@@ -73,18 +76,15 @@ export const App: React.FC = () => {
       .finally(() => {
         setTempTodo(null);
         setIsAdding(false);
+        setLoadingTodoIds(ids => ids.filter(id => id !== 0));
         setTimeout(() => {
-          (
-            document.querySelector(
-              '[data-cy="NewTodoField"]',
-            ) as HTMLInputElement
-          )?.focus();
+          inputRef.current?.focus();
         }, 0);
       });
   };
 
   const handleDelete = (todoId: number, onSuccess?: () => void) => {
-    setDeletingTodos([...deletingTodos, todoId]);
+    setLoadingTodoIds([...loadingTodoIds, todoId]);
 
     deleteTodo(todoId)
       .then(() => {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState('');
+  const [filter, setFilter] = useState<StatusFilter>(StatusFilter.All);
+
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todoValue, setTodoValue] = useState<string>('');
+  const [isAdding, setIsAdding] = useState<boolean>(false);
+
+  const [deletingTodos, setDeletingTodos] = useState<number[]>([]);
+  const [updatingTodos, setUpdatingTodos] = useState<number[]>([]);
```
TarasHoliuk: You cannot update and delete simultaneously, so you can keep only one state, renamed to `loadingTodoIds`

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+
+  const filteredTodos = todos.filter(todo => {
+    if (filter === 'active') {
```
TarasHoliuk: use enum value in comparing (everywhere)

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+
+    return true;
+  });
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    const trimmedValue = todoValue.trim();
+
+    if (trimmedValue === '') {
```
TarasHoliuk: nit: empty string is a falsy value

```suggestion
    if (!trimmedValue) {
```

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+      })
+      .catch(() => {
+        setError('Unable to add a todo');
+      })
+      .finally(() => {
+        setTempTodo(null);
+        setIsAdding(false);
+        setTimeout(() => {
+          (
+            document.querySelector(
```
TarasHoliuk: use ref. We do not query elements like this in React

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+
+    todosToUpdate.forEach(todo => {
+      handleToggle(todo.id);
+    });
+  };
+
+  const handleUpdate = (
+    todoId: number,
+    newTitle: string,
+    onSuccess?: () => void,
```
TarasHoliuk: Why optional?

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+    setIsAdding(true);
+
+    setTempTodo({
+      id: 0,
+      title: trimmedValue,
+      completed: false,
+      userId: USER_ID,
+    });
+
+    createTodo(trimmedValue)
```
TarasHoliuk, line 68: I cannot create todo now
You should send to server the `completed` field as well

src/components/TodoList.tsx
```diff
@@ -0,0 +1,137 @@
...
+          if (event.key === 'Enter') {
+            handleSave();
+          }
+
+          if (event.key === 'Escape') {
+            setEditingId(null);
+          }
+        };
+
+        const handleBlur = () => {
```
TarasHoliuk, line 63: redundant function, just use `handleSave` in the place you use `handleBlur `

src/components/Header.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    <form onSubmit={handleSubmit}>
+      <input
+        data-cy="NewTodoField"
+        type="text"
+        className="todoapp__new-todo"
+        placeholder="What needs to be done?"
+        onChange={e => setTodoValue(e.target.value)}
+        value={todoValue}
+        autoFocus
```
TarasHoliuk, line 45: use ref and focus it using ref

src/components/TodoList.tsx
```diff
@@ -0,0 +1,137 @@
...
+          handleSave();
+        };
+
+        return (
+          <div
```
TarasHoliuk, line 68: You should use `TodoItem` component here

src/components/TodoList.tsx
```diff
@@ -0,0 +1,137 @@
...
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todos: Todo[];
+  deletingTodos: number[];
+  updatingTodos: number[];
+  onDelete: (todoId: number, onSuccess?: () => void) => void;
+  onToggle: (todoId: number) => void;
+  isTemp?: boolean;
```
TarasHoliuk, line 10: the entire list `isTemp`? - sounds bad. That's because List and Item are not properly separated

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+          <TodoList
+            todos={[tempTodo]}
+            deletingTodos={[0]}
+            updatingTodos={[0]}
+            onDelete={() => {}}
+            onToggle={() => {}}
+            onUpdate={() => {}}
+            isTemp
+          />
+        )}
```
TarasHoliuk: You shouldn't render 2 lists. Handle tempTodo inside of TodoList and render or not additional tempTodo Item if exist

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState('');
+  const [filter, setFilter] = useState<StatusFilter>(StatusFilter.All);
+
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todoValue, setTodoValue] = useState<string>('');
+  const [isAdding, setIsAdding] = useState<boolean>(false);
```
TarasHoliuk, line 28: do you need both: `isAdding` and `loading` (and 2 arrays of ids)? I gueess tou can keep one boolean and one array. Or even only one array and const boolean: `const isLoading = !!loadingTodoIds.length`

TarasHoliuk (changes requested):
GJ! But some changes required to improve your code

VitaliyBondarenko1982 (approved):
Looks good.

PR #1861, opened 2025-08-01
please help me to fix it. I can`t fight this anymore

Knysh19:
[DEMO LINK](https://Knysh19.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,10 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
+/* eslint-disable @typescript-eslint/indent */
+
 import React, { useEffect, useState, useRef } from 'react';
 import { Todo } from './types/Todo';
-import {
-  getTodos,
-  createTodo,
-  deleteTodo,
-  clearCompletedTodos,
-  USER_ID,
-} from './api/todos';
+import { getTodos, createTodo, deleteTodo, USER_ID } from './api/todos';
 import { Header } from './components/Header/header';
 import { TodoList } from './components/TodoList/todoList';
 import { Footer } from './components/Footer/footer';
@@ -25,7 +21,9 @@ export const App: React.FC = () => {
   const [adding, setAdding] = useState(false);
   const [tempId, setTempId] = useState<number | null>(null);
   const inputRef = useRef<HTMLInputElement>(null);
-  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+  const [loadingTodoIds, setLoadingTodoIds] = React.useState<
+    (number | string)[]
+  >([]);
 
   useEffect(() => {
     if (!adding && inputRef.current) {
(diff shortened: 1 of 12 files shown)


vitalii-bondarenko-ua (changes requested):
You need to pass all tests.

For fix current failed test you need to create state tempTodo (type will be Todo | null).
Initial will be null. Before request for add todo on server - set to tempTodo state new todo. In finally block set null.
Screenshot 2025-08-02 at 10 32 34


Based on tempTodo state render TodoItem with loader after list of existing todos. 
It should be rendered only when tempTodo state not null.

Also check demo link - now it looks such.
Screenshot 2025-08-02 at 10 32 54

In general feel free to ask questions in chat. It will be faster to solve any problems
Screenshot 2025-08-02 at 10 33 48

src/App.tsx
```diff
@@ -1,26 +1,242 @@
...
+    return () => clearTimeout(t);
+  }, [error]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const filtered = todos.filter(t =>
+    filter === FilterType.All
+      ? true
```
h-amster, line 57: Use the "if" instead of the ternary operator

src/App.tsx
```diff
@@ -1,26 +1,242 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [filter, setFilter] = useState<FilterType>(FilterType.All);
+  const [title, setTitle] = useState('');
+  const [adding, setAdding] = useState(false);
+  const [tempId, setTempId] = useState<number | null>(null);
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [loadingTodoIds, setLoadingTodoIds] = React.useState<
+    (number | string)[]
```
h-amster, line 25: Correct the logic so that you have one type, either strings or numbers, and not both.

src/components/Footer/footer.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+interface Props {
+  filter: FilterType;
+  setFilter: React.Dispatch<React.SetStateAction<FilterType>>;
+  activeCount: number;
+  hasCompleted: boolean;
+  onClearCompleted: () => void;
+}
+
+const labels: Record<FilterType, string> = {
```
h-amster, line 12: You can capitalize the values in the FilterType enum, and then use the following
`const labels = object.values(FilterType);`

src/components/Header/header.tsx
```diff
@@ -0,0 +1,47 @@
...
+  onToggleAll,
+  isAdding,
+  inputRef,
+  showToggleAll,
+}) => (
+  <header className="todoapp__header">
+    {showToggleAll && (
+      <button
+        type="button"
+        className={`todoapp__toggle-all ${allCompleted ? 'active' : ''}`}
```
h-amster, line 28: It is better to use the classnames library.

h-amster (approved):

h-amster:
Good job. Approved.
However, please pay attention to the comments I left and correct your code so that it is more readable.

PR #1864, opened 2025-08-03
add task solution

lublubuterbrodi:
https://lublubuterbrodi.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,7 +28,6 @@ export const App: React.FC = () => {
 
   const inputRef = useRef<HTMLInputElement>(null);
 
-  // авто-скрытие ошибки через 3 сек
   const hideErrorTimer = useRef<number | null>(null);
 
   const clearError = useCallback(() => {
@@ -64,7 +63,7 @@ export const App: React.FC = () => {
     setLoading(true);
     clearError();
     try {
-      const data = await getTodos(); // без аргумента
+      const data = await getTodos();
 
       setTodos(data);
     } catch {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,376 @@
...
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [errorMsg, setErrorMsg] = useState<string | null>(null);
+  const [filter, setFilter] = useState<FilterType>(FilterType.All);
+  const [newTitle, setNewTitle] = useState('');
```
vitalii-bondarenko-ua: state newTitle should be in TodoItem component.

src/App.tsx
```diff
@@ -1,26 +1,376 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [savingIds, setSavingIds] = useState<number[]>([]);
+  const [editingId, setEditingId] = useState<number | null>(null);
+  const [editingTitle, setEditingTitle] = useState('');
```
vitalii-bondarenko-ua: editingId state looks like redundant (all data which related only for editing title should be in TodoItem component).
edititngTitle should be moved into TodoItem component.

vitalii-bondarenko-ua (changes requested):
Good job.
Consider some refactoring.

src/App.tsx
lublubuterbrodi: Hi! I fixed, thanks!

maxim2310 (changes requested):
Good job, but plz fix couople points:
1. fix this gap 
image

2. when i delete couple tasks loading should be on each of deleting todos, but in you case you just hide whole view
image
image

3. same when you click here 
image

Anton-Kuchmasov (approved):
LGTM! Approved

PR #1869, opened 2025-08-08
add task solution

yuliafito:
- [DEMO LINK](https://yuliafito.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,307 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { getTodos, addTodo, USER_ID, deleteTodo, editTodo } from './api/todos';
+// eslint-disable-next-line max-len
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
+import { Filter } from './components/Filter/Filter';
+import { FilterStatus } from './types/FilterStatus';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/Todo';
+import { ErrorMessage } from './types/ErrorMessage';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<ErrorMessage | ''>('');
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>('all');
+  const [isAdding, setIsAdding] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [title, setTitle] = useState('');
+  const [loadingTodosIDs, setLoadingTodosIDs] = useState<number[]>([]);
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<ErrorMessage | ''>('');
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>('all');
+  const [isAdding, setIsAdding] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [title, setTitle] = useState('');
+  const [loadingTodosIDs, setLoadingTodosIDs] = useState<number[]>([]);
+  const [editingId, setEditingId] = useState<number | null>(null);
```
vitalii-bondarenko-ua: it can be one state (processingTodoIds for example)

src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/Todo';
+import { ErrorMessage } from './types/ErrorMessage';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<ErrorMessage | ''>('');
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>('all');
+  const [isAdding, setIsAdding] = useState(false);
```
vitalii-bondarenko-ua: Why do you need this state?
Looks like tempTodo will be enough.

src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
+    try {
+      const createdTodo = await addTodo(newTodo);
+
+      setTodos(currentTodos => [...currentTodos, createdTodo]);
+      setTitle('');
+    } catch {
+      setErrorMessage(ErrorMessage.AddTodo);
+    } finally {
+      setIsAdding(false);
+      setTempTodo(null);
```
vitalii-bondarenko-ua, line 100: You manage both states in the same time. You do not need boolean state isAdding - will be enough tempTodo which also can be truthy of falsy (Todo or null)

vitalii-bondarenko-ua (changes requested):
Good job.
Consider some refactoring.

src/App.tsx
```diff
@@ -1,26 +1,301 @@
...
+    return () => {
+      if (timerId) {
+        clearTimeout(timerId);
+      }
+    };
+  }, [errorMessage]);
+
+  const visibleTodos = todos.filter(todo => {
+    switch (filterStatus) {
+      case 'active':
```
h-amster, line 68: Consider using enum instead of strings for all filter values.

src/App.tsx
```diff
@@ -1,26 +1,301 @@
...
+    try {
+      const results = await Promise.all(updatePromises);
+
+      setTodos(currentTodos =>
+        currentTodos.map(todo => {
+          const updated = results.find(
+            updatedTodo => updatedTodo.id === todo.id,
+          );
+
+          return updated ? updated : todo;
```
h-amster, line 221: return updated ?? todo;

h-amster (approved):
Nice work!

PR #1871, opened 2025-08-08
add task solution

iFedyna:
[DEMO LINK](https://iFedyna.github.io/react_todo-app-with-api/)


etojeDenys (changes requested):
1. after adding a new task, the loader takes up the entire screen, which it should not do

2. Remove comments

src/App.tsx
```diff
@@ -1,26 +1,493 @@
...
+                      value={editedTitle}
+                      onChange={e => setEditedTitle(e.target.value)}
+                      onBlur={() => handleUpdateTodoTitle(todo)}
+                      onKeyUp={event => {
+                        if (event.key === 'Escape') {
+                          setEditingTodoId(null);
+                        }
+                      }}
+                      ref={editTodoFieldRef}
+                    />
```
Anton-Kuchmasov, line 386: Fix styles here (Check the initial code)
image

Anton-Kuchmasov (changes requested):
Good job, but there are a couple issues needed to be fixed.

1. One "super-component" (like `App.tsx` in your solution) is bad practice. Please, divide your project into separate files.
2. Check the UI-bug in comments below

PR #1873, opened 2025-08-09
Solution React Todo App with API (complete)

Ivanyn732:
[DEMO LINK](https://Ivanyn732.github.io/react_todo-app-with-api/)

https://prnt.sc/ILskW4gtvgYy

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,274 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  addTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+  USER_ID,
+} from './api/todos';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Error } from './components/Error';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
 
-const USER_ID = 0;
+type StatusFilter = Filter;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState('');
+  const [statusFilter, setStatusFilter] = useState<StatusFilter>(Filter.All);
+  const [title, setTitle] = useState('');
+  const [isCreating, setIsCreating] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const completedCount = todos.filter(todo => todo.completed).length;
+
+  const loadTodos = async () => {
+    setLoading(true);
+    setError('');
+
+    try {
+      const data = await getTodos();
+
+      setTodos(data);
+    } catch (err) {
+      setError('Unable to load todos');
+
+      setTimeout(() => {
+        setError('');
+      }, 3000);
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  const handleFilterChange = (newFilter: Filter) => {
+    setStatusFilter(newFilter);
+  };
+
+  const handleSubmit = async (event: React.FormEvent) => {
+    event.preventDefault();
+
+    const trimmedTitle = title.trim();
+
+    if (!trimmedTitle) {
+      setError('Title should not be empty');
+      setTimeout(() => setError(''), 3000);
+
+      return;
+    }
+
+    const newTodo = {
+      userId: USER_ID,
+      title: trimmedTitle,
+      completed: false,
+    };
+
+    setIsCreating(true);
+    setTempTodo({ ...newTodo, id: 0 });
+
+    try {
+      const createdTodo = await addTodo(newTodo);
+
+      setTodos(current => [...current, createdTodo]);
+      setTitle('');
+    } catch {
+      setError('Unable to add a todo');
+      setTimeout(() => setError(''), 3000);
+    } finally {
+      setIsCreating(false);
+      setTempTodo(null);
+    }
+  };
+
+  const handleDelete = async (todoId: number) => {
+    setDeletingTodoIds(ids => [...ids, todoId]);
+
+    try {
+      await deleteTodo(todoId);
+      setTodos(current => current.filter(todo => todo.id !== todoId));
+      inputRef.current?.focus();
+    } catch {
+      setError('Unable to delete a todo');
+      setTimeout(() => setError(''), 3000);
+    } finally {
+      setDeletingTodoIds(ids => ids.filter(id => id !== todoId));
+    }
+  };
+
+  const onClearCompleted = () => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    const deletePromises = completedTodos.map(todo => {
+      setDeletingTodoIds(current => [...current, todo.id]);
+
+      return deleteTodo(todo.id)
+        .then(() => {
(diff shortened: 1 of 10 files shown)


TarasHoliuk:
You shouldn't show the loader for all todos when updating only one todo

src/App.tsx
```diff
@@ -1,26 +1,269 @@
...
+} from './api/todos';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Error } from './components/Error';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
 
-const USER_ID = 0;
+type StatusFilter = Filter;
```
TarasHoliuk: why this? Just use `Filter`

src/App.tsx
```diff
@@ -1,26 +1,269 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState('');
+  const [statusFilter, setStatusFilter] = useState<StatusFilter>(Filter.All);
+  const [title, setTitle] = useState('');
+  const [isCreating, setIsCreating] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
```
TarasHoliuk: You don't need two arrays. Use one `loadingTodoIds` for both - deleting end editing operations

src/App.tsx
```diff
@@ -1,26 +1,269 @@
...
+import { Filter } from './types/Filter';
 
-const USER_ID = 0;
+type StatusFilter = Filter;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState('');
+  const [statusFilter, setStatusFilter] = useState<StatusFilter>(Filter.All);
```
```suggestion
  const [statusFilter, setStatusFilter] = useState<Filter>(Filter.All);
```

src/App.tsx
```diff
@@ -1,26 +1,269 @@
...
+import { Error } from './components/Error';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
 
-const USER_ID = 0;
+type StatusFilter = Filter;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
```
TarasHoliuk: Do you need it? you can check `!!loadingTodoIds.length` instead

src/components/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+export const Footer: React.FC<Props> = ({
+  todos,
+  statusFilter,
+  onFilterChange,
+  onClearCompleted,
+  completedCount,
+}) => {
+  const notCompletedTodos = todos.filter(todo => !todo.completed);
+
+  const filterLinks = [
```
TarasHoliuk, line 22: Create static variables outside of the component

src/components/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+};
+
+export const Footer: React.FC<Props> = ({
+  todos,
+  statusFilter,
+  onFilterChange,
+  onClearCompleted,
+  completedCount,
+}) => {
+  const notCompletedTodos = todos.filter(todo => !todo.completed);
```
TarasHoliuk, line 20: You can pass the number as a prop instead of todos array, as you do with `completedCount`

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,185 @@
+import React, { useEffect, useRef, useState } from 'react';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todo: Todo;
+  isTemp?: boolean;
```
TarasHoliuk, line 6: redundant, for tempTodo you should just always pass `isLoading` true

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,185 @@
...
+  onDelete: (todoId: number) => void;
+  isDeleting?: boolean;
+  onToggleStatus: (todoId: number, newStatus: boolean) => void;
+  isLoading?: boolean;
+  loading?: boolean;
```
TarasHoliuk:
```suggestion
  isLoading?: boolean;
```
You can get rid of additional loading prop

src/App.tsx
```diff
@@ -1,26 +1,269 @@
...
+
+      setTodos(prevTodos =>
+        prevTodos.map(todo =>
+          todo.id === todoId ? { ...todo, title: updatedTodo.title } : todo,
+        ),
+      );
+    } catch (err) {
+      setError('Unable to update a todo');
+      setTimeout(() => setError(''), 3000);
+      throw err;
```
TarasHoliuk, line 217: Don't throw error here. You catch it in `handleEditSubmit` and just do the same: `setError('Unable to update a todo');` which is redundant since you've already done it here

TarasHoliuk: Fix similar cases

src/App.tsx
```diff
@@ -1,26 +1,269 @@
...
+          toggleAll={toggleAll}
+          isAllCompleted={todos.every(todo => todo.completed)}
+          todos={todos}
+          isLoadingTodos={loading}
+        />
+
+        <TodoList
+          todos={filteredTodos}
+          tempTodo={tempTodo}
+          setError={setError}
```
TarasHoliuk, line 243: no need to pass it if you handle errors in the handler functions already

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,185 @@
...
+  onToggleStatus,
+  isLoading,
+  loading,
+  editingTodoId,
+  setEditingTodoId,
+  onRename,
+  setError,
+  setLoading,
+}) => {
+  const isEditing = editingTodoId === id;
```
TarasHoliuk: Don't pass it as a prop
You need it only here - create a local state `isEditing` here and delete `editingTodoId` in the App.tsx

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,185 @@
...
+  };
+
+  return (
+    <div
+      data-cy="Todo"
+      className={`todo ${completed ? 'completed' : ''}`}
+      key={id}
+    >
+      <label className="todo__status-label" htmlFor={`todo-status-${id}`}>
+        {/* {} */}
```
TarasHoliuk, line 100: delete it

src/App.tsx
TarasHoliuk: UPD: if you need to focus `editingInput` you can throw here and catch the error in `handleEditSubmit` to focus

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,185 @@
...
+      try {
+        setLoading(true);
+        await onDelete(id);
+
+        return true;
+      } catch {
+        setError('Unable to delete a todo');
+        setTimeout(() => setError(''), 3000);
+
+        return false;
```
TarasHoliuk: Instead of returning something just focus input here

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,185 @@
...
+        />
+      </label>
+
+      {isEditing ? (
+        <form
+          onSubmit={async e => {
+            e.preventDefault();
+            const ok = await handleEditSubmit();
+
+            if (ok) {
```
TarasHoliuk: handleEditSubmit should accept event as argument, do ` e.preventDefault();` inside of the function

Instead of returning `true` in `try` block just set correct state value there

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,185 @@
...
+          <input
+            ref={editInputRef}
+            type="text"
+            data-cy="TodoTitleField"
+            className="todo__title-field"
+            value={editedTitle}
+            onChange={handleTitleChange}
+            onKeyUp={handleKeyUp}
+            onBlur={async () => {
+              await new Promise(resolve => setTimeout(resolve, 0));
```
TarasHoliuk: redundant
just do this:

onBlur={handleEditSubmit}

TarasHoliuk (changes requested):

vitalii-bondarenko-ua (approved):
Looks good.

PR #1879, opened 2025-08-14
Develop

yuliimar:
https://yuliimar.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,274 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState, useRef } from 'react';
+import {
+  getTodos,
+  USER_ID,
+  createTodo,
+  deleteTodo,
+  updateTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
 
-const USER_ID = 0;
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,26 +1,307 @@
...
+            />
+          </>
+        )}
+      </div>
+
+      {tempTodo && (
+        <div data-cy="TodoItem" style={{ position: 'relative' }}>
+          <TodoItem todo={tempTodo} onDelete={() => {}} />
+        </div>
+      )}
```
anastasiiavorobiova: Temporary todo should be right after the todo list

anastasiiavorobiova (changes requested):
Great work! 
Loaders look a bit broken when toggling one todo state

src/App.tsx
```diff
@@ -1,26 +1,318 @@
...
+              style={{ position: 'relative' }}
+            >
+              <TodoItem
+                todo={tempTodo}
+                onDelete={() => {}}
+                onUpdate={() => Promise.resolve()}
+                onError={setError}
+              />
+            </div>
+          )}
```
FaiHamid: This code should be placed in the TodoList component. Also, you should remove the div that wraps the TodoItem for the temporary todo.
image

FaiHamid (changes requested):
Well done!! 
I added just one comment to improve the code.

src/App.tsx
yuliimar: ok, thanks.

vadiimvooo (approved):
Well done

PR #1884, opened 2025-08-17
Add task

iteamfox:
[DEMO LINK](https://iteamfox.github.io/react_todo-app-with-api/)


Zibi95 (changes requested):
I’ve already suggested changes in another PR that contains the exact same code. Once you apply the fixes there and that PR gets accepted, you should copy those changes here. This one should only be opened after the first one is merged, otherwise it doesn’t really make sense.

src/App.tsx
```diff
@@ -1,26 +1,299 @@
...
+
+  const loadTodos = useCallback(() => {
+    setLoading(true);
+    todoService
+      .getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
+        setShowError(true);
+        setTimeout(() => setShowError(false), 3000);
```
zibi-dev, line 36: showError state is duplicate for errorMessage. Look at places where you use them. They are both used in the same places. You can just derive state from errorMessage. 

`showError = !!errorMessage`

src/App.tsx
```diff
@@ -1,26 +1,299 @@
...
-  if (!USER_ID) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+  const [newTodoTitle, setTodoTitle] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [showError, setShowError] = useState(false);
+  const [loading, setLoading] = useState(true);
+  const [filteredBy, setFilteredBy] = useState(Status.All);
+  const [editingTodosIds, setEditingTodosIds] = useState<number[]>([]);
```
zibi-dev, line 20: Do you need that state here? I feel like it belongs to TodoItem and it sohuld be boolean, bcs only one todo Item at time should be editable.

src/App.tsx
```diff
@@ -1,26 +1,299 @@
...
+import { Header } from './components/Header/Header';
+import { Footer } from './components/Footer/Footer';
+import { Status } from './types/Status';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+  const [newTodoTitle, setTodoTitle] = useState('');
```
zibi-dev, line 15: I believe we dont need state for input. It can be uncontrolled input and we can extract value from input on submit.

zibi-dev: The same applies for TodoItem edit input state

src/App.tsx
```diff
@@ -1,26 +1,299 @@
...
+  const [errorMessage, setErrorMessage] = useState('');
+  const [showError, setShowError] = useState(false);
+  const [loading, setLoading] = useState(true);
+  const [filteredBy, setFilteredBy] = useState(Status.All);
```
zibi-dev, line 19: Let's use query params instead. instead of passing state around to components we can simply in the Footer apply query param for filter and here or in the TodoList just read that param and filter based on that.  WDYT?

Zibi95 (changes requested):
Good job. I have suggested some improvements and simplifications.

PR #1885, opened 2025-08-17
Solution P3

PaulKotsan:
[DEMO LINK](https://PaulKotsan.github.io/react_todo-app-with-api/)
Spent 3 days on it. Needed '''some''' **AI**ssistance, not much 💯

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,14 @@
 /* eslint-disable react-hooks/rules-of-hooks */
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import * as serverManager from './api/todos';
 import { Todo } from './types/Todo';
-import classNames from 'classnames';
+import { TodoHeader } from './components/Header';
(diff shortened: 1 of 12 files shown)


volodymyr-soltys97 (changes requested):
Hi, it is bad practice to write all the code in one component, you need to separate the logic into different components, for example `Header, Footer, TodoItem, TodoList`

src/App.tsx
```diff
@@ -1,26 +1,322 @@
...
+
+    if (filterType === 'active') {
+      return !statusForFiltering;
+    }
+
+    if (filterType === 'completed') {
+      return statusForFiltering;
+    }
+
+    return true;
```
etojeDenys: you can use a switch case here

src/App.tsx
```diff
@@ -1,26 +1,322 @@
...
+  const handleToggleTodo = (todo: Todo) => {
+    const newStatus = !todo.completed;
+
+    setTempStatuses(prev => ({ ...prev, [todo.id]: newStatus }));
+    updateTodo({ ...todo, completed: newStatus }, false);
+  };
+
+  const handleClearCompleted = () => {
+    todos.filter(t => t.completed).forEach(t => deleteTodo(t.id, false, ''));
+  };
```
etojeDenys: you should use Promise.all() here

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #1887, opened 2025-08-18
task solution

VladaProskura:
- [DEMO LINK](https://vladaproskura.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,256 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import {
+  deleteTodo,
+  getTodos,
+  postTodo,
+  updateTodo,
+  USER_ID,
+} from './api/todos';
+import { Todo } from './types/Todo';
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,26 +1,256 @@
...
+    }
+  };
+
+  const clearCompletedTodos = async () => {
+    const completedIds = todos
+      .filter(todo => todo.completed)
+      .map(todo => todo.id);
+
+    completedIds.forEach(id => handleDeleteTodo(id));
+  };
```
etojeDenys: it is better to use Promise.all here

etojeDenys (changes requested):
1. toggle all button should look different
your:
image

example:
image

2. after successfully adding a post, the input should become focused

3. the editing style is different
your: 
image

example:
image

4. do not forget to implement these steps
image

etojeDenys (changes requested):
1. after pressing the toggle buttons we see this error. however, when I try to change the status, it changes
image

2. position of the toggle button is incorrect
example: 
image

your: 
image

3. the same problem with editing style
image

Anton-Kuchmasov (changes requested):
Nice work you did!

Please fix the styles here:
image

(check out an initial markup - it might help you find correct className for the input)

etojeDenys (approved):
lgtm

PR #1897, opened 2025-08-26
todo with api solution

NicssLucik02:
[DEMOLINK](https://NicssLucik02.github.io/react_todo-app-with-api/)


O1ympus (approved):

src/Loadings.ts
```diff
@@ -0,0 +1,25 @@
...
+//   if (
+//     !isLoading.add &&
+//     !isLoading.todos &&
+//     !isLoading.deletedId &&
+//     inputRef.current
+//   ) {
+//     inputRef.current.focus();
+//   }
+
+// }, [isLoading.add, isLoading.todos, isLoading.deletedId]);
```
alexander-ignatow: Don't keep commented out code
And don't keep an empty file

src/types/Props.ts
alexander-ignatow, line 1: 1. mind the file naming - only components should be Capitalized
2. I am not sure if there is a point to keep all "props' types here. Normally you keep props near the omponent and even don't export them

src/components/TodoMain/TodoMain.tsx
```diff
@@ -0,0 +1,48 @@
+import { PropsMain } from '../../types/Props';
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from './TodoInfo/TodoInfo';
+
+export const TodoMain: React.FC<PropsMain> = ({
```
alexander-ignatow: Component name is unclear. Is it a list? Some area? A main part of the Todo item?

alexander-ignatow (changes requested):
Good, but lets improve your solution
What is the point of breaking types into multiple short files? IMO - if you get rid of Props.ts all other types can be placed inside one single file

Second problem - you have put _ALL_ your app's logic in a single hook.
This is an anti-pattern, that is called God Object. Please think and break your monster-hook into parts. For instance - why you handle DOM and logic together as a single mix? It is OK to have some single point to handle state and interact with the API, but all other things should be removed from there

NicssLucik02:
I hope it looks better now

src/hooks/useTodos.ts
```diff
@@ -0,0 +1,93 @@
...
+    loadTodos();
+  }, []);
+
+  return {
+    todos,
+    isLoadingTodos,
+    isLoadingAdd,
+    isLoadingDelete,
+    isLoadingUpdate,
+    isLoadingEdit,
```
ArturErmolenko: why we need so many loading state?
can we have just one?

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,29 @@
+import classNames from 'classnames';
+
+export type Props = {
+  currentError: string | null;
+  handleHideError: () => void;
+};
+
+export const ErrorNotification: React.FC<Props> = ({
```
ArturErmolenko, line 8: where is useErrorReset hook?
that is the best place i think you should call it

ArturErmolenko: i mean useEffect
state can be place in App

src/hooks/useTodos.ts
```diff
@@ -0,0 +1,93 @@
...
+    setCurrentError,
+    setTodos,
+    inputRef,
+  );
+  const { isLoadingUpdate, handleCheckTodo, handleToggleAll } = useUpdateTodos(
+    todos,
+    setTodos,
+    setCurrentError,
+    setIsLoadingTodos,
+  );
```
ArturErmolenko: to be honest it's hard to understand what happened here
wat is the diff gor exemple here
handleEditTodo,
    editTodoQuery,
    handleEditTodoQuery,
    defaultEditInputValue,
    handleActivateEdit,
    activeEdit,
    setActiveEdit,
    
    
    so many states is there a chance to get rid of smth?

src/hooks/useTodos.ts
ArturErmolenko: i believe we should have 1 loading state)

ArturErmolenko (changes requested):
Nice by let's make one more iteration
We should definitely try to decrease states
Try to handle things with just one state

ArturErmolenko (approved):
Well done 🔥

PR #1902, opened 2025-08-28
sol

O1ympus:
[DEMO LINK](https://O1ympus.github.io/react_todo-app-with-api/)


AARON-FOX (approved):
This is a good solution, but you need to split App.tsx into several components, remove comments about tasks, and move the user ID to a separate file. I will approve this pull request, but you will need to make the changes for mentor review.

DanilWeda (approved):
Good job. Approved.

PR #1906, opened 2025-08-31
add task soluiton

maxrori228:
- [DEMO LINK](https://maxrori228.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,49 +6,53 @@ import { deleteTodos, getTodos, postTodos, updateTodos } from './api/todos';
 import { Todo } from './types/Todo';
 import { Header } from './components/Header/Header';
 import { TodoList } from './components/TodoList/TodoList';
-import { FILTERS, ERRORS_MESSAGE } from './types/enums';
+import { Filters, ErrorsMessage } from './types/enums';
 import { USER_ID } from './consts/consts';
 import { Footer } from './components/Footer/Footer';
+// eslint-disable-next-line max-len
 import { ErrorNotification } from './components/ErrorNotifications/ErrorNotifications';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [error, setError] = useState('');
-  const [filter, setFilter] = useState(FILTERS.all);
+  const [filter, setFilter] = useState(Filters.all);
   const [title, setTitle] = useState('');
   const [isAdding, setIsAdding] = useState(false);
   const [addingTodoId, setAddingTodoId] = useState<number | null>(null);
   const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
   const [updatingTodoIds, setUpdatingTodoIds] = useState<number[]>([]);
   const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
-  const [editingTodoTitle, setEditingTodoTitile] = useState('');
+  const [editingTodoTitle, setEditingTodoTitle] = useState('');
   const editTitleRef = useRef<HTMLInputElement>(null);
+  const inputRef = useRef<HTMLInputElement>(null);
 
   useEffect(() => {
+    if (!USER_ID) {
+      return;
+    }
+
     const todosFromServer = async () => {
       try {
         const preparedTodos = await getTodos();
 
         setTodos(preparedTodos);
-      } catch (err) {
-        setError(ERRORS_MESSAGE.loadError);
+      } catch {
+        setError(ErrorsMessage.loadError);
       }
     };
 
     todosFromServer();
-  }, [ERRORS_MESSAGE.loadError]);
+  }, []);
 
   useEffect(() => {
-    if (error) {
-      const errorTimer = setTimeout(() => {
-        setError('');
-      }, 3000);
-
-      return () => clearTimeout(errorTimer);
+    if (!error) {
+      return;
     }
-  }, [error]);
 
-  const inputRef = useRef<HTMLInputElement>(null);
+    const timer = setTimeout(() => setError(''), 3000);
+
+    return () => clearTimeout(timer);
+  }, [error]);
 
   useEffect(() => {
     if (inputRef.current) {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,344 @@
...
+      setTodos(prev => prev.filter(todo => todo.id !== todoId));
+    } catch {
+      setError(ERRORS_MESSAGE.deleteError);
+    } finally {
+      setDeletingTodoIds(prev => prev.filter(id => id !== todoId));
+      focusTitleInput();
+    }
+  };
+
   if (!USER_ID) {
```
Aonarro: Move the conditional return above hooks to avoid calling hooks conditionally.

src/App.tsx
```diff
@@ -1,26 +1,344 @@
...
-
-const USER_ID = 0;
+import { deleteTodos, getTodos, postTodos, updateTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { FILTERS, ERRORS_MESSAGE } from './types/enums';
+import { USER_ID } from './consts/consts';
+import { Footer } from './components/Footer/Footer';
+import { ErrorNotification } from './components/ErrorNotifications/ErrorNotifications';
```
Aonarro, line 13: // eslint-disable-next-line max-len You can use this rule for linting long immports

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,108 @@
...
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {visibleTodos.map(todo => (
+        <div
+          data-cy="Todo"
+          className={classNames('todo', { completed: todo.completed })}
+          key={todo.id}
+        >
+          <label className="todo__status-label">
```
Aonarro, line 44: /* eslint-disable-next-line jsx-a11y/label-has-associated-control */ You can also can use this rule for better linting

src/types/enums.ts
```diff
@@ -0,0 +1,13 @@
+export enum FILTERS {
+  all = 'All',
+  completed = 'Completed',
+  active = 'Active',
+};
+
+export enum  ERRORS_MESSAGE {
```
Aonarro: For the enum name, you can use either all uppercase letters UPPER_SNAKE_CASE or PascalCase

Aonarro (approved):
Everything looks good!

src/types/enums.ts
dothearts: FIx this error

src/App.tsx
```diff
@@ -1,26 +1,344 @@
...
+        prev.filter(id => !todosToUpdate.some(t => t.id === id)),
+      );
+    }
+  };
+
+  const handleEditing = useCallback((todo: Todo) => {
+    setEditingTodoId(todo.id);
+    setEditingTodoTitile(todo.title);
+  }, []);
+
```
dothearts, line 230: You can disabled all this rule using eslint-disable (only for line)

DanilWeda (changes requested):
Good job. 
Please check all eslint errors (disabled it, or fix). 
Many thanks!

src/components/ErrorNotifications/ErrorNotifications.tsx
```diff
@@ -0,0 +1,29 @@
+import classNames from 'classnames';
+import React from 'react';
+
+type Props = {
+  error: string;
+  setError: React.Dispatch<React.SetStateAction<string>>;
```
alexander-ignatow, line 6: minor: for the future don't give "too wide" functionality to the component.
Simply having `clearError: () => void` would be better

alexander-ignatow: Minor: I would also put cleanup timer functionality inside this component

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,75 @@
+import classNames from 'classnames';
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { Filters } from '../../types/enums';
+
+type Props = {
+  todos: Todo[];
+  items: Todo[];
```
alexander-ignatow, line 8: Bad property naming: whats the difference between todos and items?

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,75 @@
...
+  filter: Filters;
+  setFilter: React.Dispatch<React.SetStateAction<Filters>>;
```
alexander-ignatow, line 10: minor: unnecessarily wide option type
just `setFilter: (value: Filter) => void` will work better here

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { forwardRef } from 'react';
+import { Todo } from '../../types/Todo';
+import classNames from 'classnames';
+
+type Props = {
+  todos: Todo[];
+  title: string;
+  setTitle: React.Dispatch<React.SetStateAction<string>>;
```
alexander-ignatow, line 8: why not put the entire title state inside this component? Looks like an internal matter of the header

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,50 @@
...
+  isAdding: boolean;
+  handleToggleAll: (event: React.MouseEvent<HTMLButtonElement>) => void;
+  handleAddTodo: (event: React.FormEvent<HTMLFormElement>) => void;
+};
+
+export const Header = React.memo(
+  forwardRef<HTMLInputElement, Props>(
```
alexander-ignatow, line 15: You use forward ref to pass the ref to the parent just to focus it
Why not do it inside the header?

src/App.tsx
```diff
@@ -1,26 +1,338 @@
...
+  useEffect(() => {
+    if (!error) {
+      return;
+    }
+
+    const timer = setTimeout(() => setError(''), 3000);
+
+    return () => clearTimeout(timer);
+  }, [error]);
```
alexander-ignatow, line 55: this logic would better fit ErrorNotification component

src/App.tsx
```diff
@@ -1,26 +1,338 @@
...
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, []);
```
alexander-ignatow, line 61: This would be better taken to the Header component

alexander-ignatow (approved):
Lets move forward to the other tasks, but please review all my comments about components, component scope and component properties

PR #1909, opened 2025-09-01
Develop

YuliaProkipchuk:
[DEMO LINK](https://YuliaProkipchuk.github.io/react_todo-app-with-api/)


DavidHurmanchuk (changes requested):
1. In FilterButton, you use <a href> with onClick but without preventDefault, which will trigger a page reload.
2. In TodoItem, linter warns because handleDelete is used before it is defined.
Overall, your codebase already shows a strong sense of structure and clean separation of concerns

DavidHurmanchuk (approved):
Well done!

DanilWeda (approved):
Good job. 
Approved!

PR #1913, opened 2025-09-01
Third stage

DidZhara:
[DEMO LINK](https://DidZhara.github.io/react_todo-app-with-api/)


src/components/TodoRow.tsx
```diff
@@ -0,0 +1,185 @@
...
+      setIsLoading(true);
+      try {
+        await onRename?.(trimmedTitle);
+      } finally {
+        setIsLoading(false);
+      }
+    }
+
+    setEdited(false);
+  };
```
derpedcatto, line 112: These functions do pretty much the same thing with only difference being in `event` prop, you could condense it into one function and make `event` optional similar to this example from my code:
```tsx
  const handleTodoTitleChange = async (event?: React.FormEvent) => {
    if (event) {
      event.preventDefault();
    }
    
    // ...
```

derpedcatto (approved):
looks good chief 🫡

DanilWeda (approved):
Good job. 
Approved. 
Many thanks!

PR #1924, opened 2025-09-04
add task solution

pbahrynovskyi:
[DEMO LINK](https://pbahrynovskyi.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,22 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 /* eslint-disable jsx-a11y/label-has-associated-control */
 import React, { useEffect, useRef, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
 import { addTodo, deleteTodo, getTodos, USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
 import { client } from './utils/fetchClient';
-import classNames from 'classnames';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+
+enum ErrorMessages {
+  LOAD_TODOS = 'Unable to load todos',
+  ADD_TODO = 'Unable to add a todo',
+  DELETE_TODO = 'Unable to delete a todo',
+  UPDATE_TODO = 'Unable to update a todo',
+  EMPTY_TITLE = 'Title should not be empty',
+}
 
 export const App: React.FC = () => {
   const [todosList, setTodosList] = useState<Todo[]>([]);
@@ -37,7 +48,7 @@ export const App: React.FC = () => {
 
         setTodosList(todos);
       } catch {
-        setErrorMsg('Unable to load todos');
+        setErrorMsg(ErrorMessages.LOAD_TODOS);
       }
     };
 
@@ -86,7 +97,7 @@ export const App: React.FC = () => {
   const handleAddTodo = async (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
     if (!newTitle.trim()) {
-      setErrorMsg('Title should not be empty');
+      setErrorMsg(ErrorMessages.EMPTY_TITLE);
 
       return;
     }
@@ -106,10 +117,9 @@ export const App: React.FC = () => {
       setTodosList(list => [...list, addedTodo]);
       setNewTitle('');
     } catch {
-      setErrorMsg('Unable to add a todo');
+      setErrorMsg(ErrorMessages.ADD_TODO);
     } finally {
       setPendingTodo(null);
-
       if (newTitleRef.current) {
         setTimeout(() => {
           newTitleRef.current?.focus();
@@ -124,10 +134,9 @@ export const App: React.FC = () => {
       await deleteTodo(id);
       setTodosList(list => list.filter(todo => todo.id !== id));
     } catch {
-      setErrorMsg('Unable to delete a todo');
+      setErrorMsg(ErrorMessages.DELETE_TODO);
     } finally {
       setDeletingTodoId(null);
-
       if (newTitleRef.current) {
         setTimeout(() => {
           newTitleRef.current?.focus();
@@ -136,36 +145,15 @@ export const App: React.FC = () => {
     }
   };
 
-  const handleDeleteCompleted = async () => {
+  const handleDeleteCompleted = () => {
     const completedTodos = todosList.filter(todo => todo.completed);
     const completedIds = completedTodos.map(todo => todo.id);
 
     setDeletingTodosIds(ids => [...ids, ...completedIds]);
 
-    try {
-      const results = await Promise.allSettled(completedIds.map(deleteTodo));
-      const successfulIds = completedTodos
-        .filter((_todo, i) => results[i].status === 'fulfilled')
-        .map(todo => todo.id);
-
-      if (results.some(r => r.status === 'rejected')) {
-        setErrorMsg('Unable to delete a todo');
-      }
-
-      if (successfulIds.length > 0) {
-        setTodosList(list =>
-          list.filter(todo => !successfulIds.includes(todo.id)),
-        );
-      }
-    } finally {
-      setDeletingTodosIds(ids => ids.filter(id => !completedIds.includes(id)));
-
-      if (newTitleRef.current) {
-        setTimeout(() => {
-          newTitleRef.current?.focus();
-        }, 0);
-      }
-    }
+    completedTodos.forEach(todo => {
+      handleDeleteTodo(todo.id);
+    });
   };
 
   const handleToggleTodo = async (todo: Todo) => {
@@ -176,7 +164,7 @@ export const App: React.FC = () => {
 
       setTodosList(list => list.map(t => (t.id === todo.id ? updated : t)));
     } catch {
-      setErrorMsg('Unable to update a todo');
+      setErrorMsg(ErrorMessages.UPDATE_TODO);
     } finally {
       setUpdatingIds(ids => ids.filter(id => id !== todo.id));
     }
@@ -208,7 +196,7 @@ export const App: React.FC = () => {
       }
 
       if (results.some(r => r.status === 'rejected')) {
-        setErrorMsg('Unable to update a todo');
+        setErrorMsg(ErrorMessages.UPDATE_TODO);
       }
     } finally {
       setUpdatingIds(ids =>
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,26 +1,426 @@
...
+      return;
+    }
+
+    const timer = setTimeout(() => setErrorMsg(''), 3000);
+
+    return () => clearTimeout(timer);
+  }, [errorMsg]);
+
+  const filterTodos = (todos: Todo[], filter: string) => {
+    if (filter === 'Active') {
```
DmitriyALEX, line 81: use enums instead strings

src/App.tsx
```diff
@@ -1,26 +1,426 @@
...
+  const visibleTodos = filterTodos(todosToDisplay, currentFilter);
+
+  const handleTitleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setNewTitle(e.target.value);
+  };
+
+  const handleAddTodo = async (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    if (!newTitle.trim()) {
+      setErrorMsg('Title should not be empty');
```
DmitriyALEX: use enums instead strings

src/App.tsx
```diff
@@ -1,26 +1,426 @@
...
+                    data-cy="TodoTitleField"
+                    value={editingTitle}
+                    onChange={e => setEditingTitle(e.target.value)}
+                    onBlur={() => {
+                      if (!updatingIds.includes(todo.id)) {
+                        handleEditSubmit(todo);
+                      }
+                    }}
+                    onKeyUp={e => {
+                      if (e.key === 'Escape') {
```
DmitriyALEX: try to write functions body and logic outside html

src/App.tsx
```diff
@@ -1,26 +1,426 @@
...
+            </div>
+          ))}
+        </section>
+
+        {todosList.length > 0 && (
+          <footer className="todoapp__footer" data-cy="Footer">
+            <span
+              className="todo-count"
+              data-cy="TodosCounter"
+            >{`${remainingItems.length} items left`}</span>
```
DmitriyALEX: jsx don't need interpolation in this case

src/App.tsx
```diff
@@ -1,26 +1,426 @@
...
+            <nav className="filter" data-cy="Filter">
+              {['All', 'Active', 'Completed'].map(option => (
```
DmitriyALEX: use constant/enums for this values instead of defining the array inline in jsx/html.

src/App.tsx
```diff
@@ -1,26 +1,426 @@
...
+  };
+
+  const handleToggleAll = async () => {
+    const allCompleted = todosList.every(t => t.completed);
+    const toUpdate = todosList.filter(t => t.completed === allCompleted);
+
+    setUpdatingIds(ids => [...ids, ...toUpdate.map(t => t.id)]);
+
+    try {
+      const results = await Promise.allSettled(
```
DmitriyALEX, line 182: allSettled seems overkill for this case, use forEach

DmitriyALEX (approved):
Good job! Consider splitting into separate components

src/App.tsx
pbahrynovskyi: thanks for advice

src/App.tsx
pbahrynovskyi: ok, thanks for advice

src/App.tsx
pbahrynovskyi: I disagree

src/App.tsx
pbahrynovskyi: Ok, thanks for advice, I will think about such solution

src/App.tsx
DmitriyALEX: <span >{remainingItems.length} items left</span > 

seems better instead of 

<span >{`${remainingItems.length} items left`}</span >

so why you are  disagree?

DmitriyALEX (approved):

src/App.tsx
```diff
@@ -1,26 +1,312 @@
...
+  EMPTY_TITLE = 'Title should not be empty',
+}
 
 export const App: React.FC = () => {
+  const [todosList, setTodosList] = useState<Todo[]>([]);
+  const [currentFilter, setCurrentFilter] = useState('All');
+  const [errorMsg, setErrorMsg] = useState('');
+
+  const [newTitle, setNewTitle] = useState('');
+  const newTitleRef = useRef<HTMLInputElement>(null);
```
alexander-ignatow, line 27: You create ref in the parent to pass it to the child. And use it for the focus only - not for some interactions
this is wrong

Move this ref to the child and focus it inside

src/components/Header.tsx
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+
+interface HeaderProps {
+  newTitle: string;
+  onTitleChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+  onAddTodo: (e: React.FormEvent<HTMLFormElement>) => void;
```
alexander-ignatow: Your typings are too close to the actual implementation, while they should reflect the logic

The "title" looks confusing here
You may use AI to pick variable names - just describe your current logic to the chat and ask it to suggest variable names (my CTO does so regularly)

```suggestion
  newTodoText: string;
  onNewTodoTextChange: (value: string) => void;
  onAddTodo: (newTodoText: string) => void;
```

src/components/Header.tsx
```diff
@@ -0,0 +1,51 @@
...
+  onToggleAll,
+}) => {
+  return (
+    <header className="todoapp__header">
+      {todosList.length > 0 && (
+        <button
+          type="button"
+          className={classNames('todoapp__toggle-all', {
+            active: todosList.length > 0 && todosList.every(t => t.completed),
+          })}
```
alexander-ignatow: This changes the look of the button - but it does not change the button behavior.
And that does not disable the button - so your intent here is not clear to me

src/components/Header.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <form onSubmit={onAddTodo}>
+        <input
+          data-cy="NewTodoField"
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          value={newTitle}
+          onChange={onTitleChange}
+          ref={newTitleRef}
+          disabled={!!pendingTodo}
```
alexander-ignatow: You pass the whole object just to !! it and to disable the input - that is bad

Please remove pendingTodo property and add something like 'newTodoFormEnabled', 'canAddTodo', etc of the boolean type

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+interface TodoItemProps {
+  todo: Todo;
+  onDelete: (id: number) => void;
+  onToggle: (todo: Todo) => void;
+  onEditSubmit: (todo: Todo) => void;
+  onStartEditing: (todo: Todo) => void;
+  onCancelEditing: () => void;
+  editingTodoId: number | null;
+  editingTitle: string;
+  setEditingTitle: (title: string) => void;
```
alexander-ignatow, line 16: Remove this 5 props completely and move all editing logic entirely inside the Todo item

Exception - only if you need to guarantee that just one todo is edited at the given moment - you can pass onStartEditing (todoId: number) => void and isEditing: bool

Otherwise hide it inside the component

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+      <div
+        data-cy="TodoLoader"
+        className={classNames('modal overlay', {
+          'is-active':
+            todo.id === 0 ||
+            todo.id === deletingTodoId ||
+            deletingTodosIds.includes(todo.id) ||
+            updatingIds.includes(todo.id),
+        })}
+      >
```
alexander-ignatow, line 104: This calculation should happen on the higher level
TodoItem does not need this logic

Just pass one property isLoading: boolean

alexander-ignatow (changes requested):
Good, but some changes are needed

Please don't share the information between the components that they dont need.
Try incapsulating the details inside the components

src/App.tsx
pbahrynovskyi: Sorry, but it ruined the logic of whole project.  I went to the Q&A with this question and was adviced not to change it.

src/components/Header.tsx
pbahrynovskyi: Thank you. Fixed

src/components/Header.tsx
pbahrynovskyi: Thank you. Fixed

src/components/Header.tsx
pbahrynovskyi: Thank you. Fixed

src/components/TodoItem.tsx
pbahrynovskyi: Thank you. Fixed

src/components/TodoItem.tsx
pbahrynovskyi: Sorry, I didn't change it, because it ruins my logic too much(

src/App.tsx
```diff
@@ -1,26 +1,314 @@
...
+      return;
+    }
+
+    const timer = setTimeout(() => setErrorMsg(''), 3000);
+
+    return () => clearTimeout(timer);
+  }, [errorMsg]);
+
+  const filterTodos = (todos: Todo[], filter: string) => {
+    if (filter === 'Active') {
```
ArturErmolenko, line 81: still using magic string - we should fix it - use enum values instead of magic string

src/App.tsx
```diff
@@ -1,26 +1,314 @@
...
+      return todos.filter(todo => !todo.completed);
+    }
+
+    if (filter === 'Completed') {
```
ArturErmolenko, line 85: switch case

src/App.tsx
```diff
@@ -1,26 +1,314 @@
...
+            remainingCount={remainingItems.length}
+            currentFilter={currentFilter}
+            onFilterChange={setCurrentFilter}
+            onClearCompleted={handleDeleteCompleted}
+            hasCompletedTodos={!todosList.every(todo => !todo.completed)}
+          />
+        )}
+      </div>
+
+      <div
```
ArturErmolenko, line 297: ErrorComponent?

ArturErmolenko (approved):
👍

PR #1925, opened 2025-09-04
solution

iastivka:
[DEMO LINK](https://iastivka.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,203 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useEffect, useMemo, useRef, useState } from 'react';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Todo } from './types/Todo';
+import * as todoMethods from './api/todo';
+import { ErrorNotification } from './components/Error';
+import { FilterStatus } from './types/FilterStatus';
+import getTodosFilter from './utils/getTodosFilter';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>(
+    FilterStatus.All,
+  );
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletingTodoIds, setDeletingTodoIds] = useState<number[] | null>(null);
+  const [updatingTodoIds, setUpdatingTodoIds] = useState<number[] | null>(null);
+  const [isInputDisabled, setInputDisabled] = useState(false);
+
+  const isTodosEmpty = todos.length === 0;
+  const todosActiveQuantity = todos.filter(todo => !todo.completed).length;
+  const todosComplitedQuantity = todos.filter(todo => todo.completed).length;
+  const allTodosIsComplited = todos.every(todo => todo.completed);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [isInputDisabled]);
+
+  useEffect(() => {
+    todoMethods
+      .getTodos()
+      .then(setTodos)
+      .catch(error => {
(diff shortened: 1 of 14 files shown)


Chips666 (changes requested):
It seems like test dont pass, try to fix it.
Check your add-todo logic and todo rendering.

src/App.tsx
```diff
@@ -1,26 +1,203 @@
...
+      inputRef.current.focus();
+    }
+  }, [isInputDisabled]);
+
+  useEffect(() => {
+    todoMethods
+      .getTodos()
+      .then(setTodos)
+      .catch(error => {
+        setErrorMessage('Unable to load todos');
```
ArturErmolenko: magic sting - use enum for error messages

src/components/Header.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const handleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (title.trim() === '') {
+      setErrorMessage('Title should not be empty');
```
ArturErmolenko: enum

src/App.tsx
```diff
@@ -1,26 +1,203 @@
...
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>(
+    FilterStatus.All,
+  );
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletingTodoIds, setDeletingTodoIds] = useState<number[] | null>(null);
+  const [updatingTodoIds, setUpdatingTodoIds] = useState<number[] | null>(null);
```
ArturErmolenko: what is the different between this states?
can we use just one?
like processingIds or smth like that

src/App.tsx
```diff
@@ -1,26 +1,203 @@
...
+  const [isInputDisabled, setInputDisabled] = useState(false);
```
ArturErmolenko, line 30:
```suggestion
```
we shouldn't handle it with extra state - let's find better way

ArturErmolenko (changes requested):
Almost done 👍 
Let's make final clean up

src/App.tsx
alexander-ignatow: Please address this issue

src/components/Header.tsx
```diff
@@ -0,0 +1,76 @@
+import classNames from 'classnames';
+import React, { useState } from 'react';
+import { ErrorMessage } from '../types/ErrorMessage';
+
+type Props = {
+  addTodo: (title: string) => Promise<void>;
+  setErrorMessage: (error: string) => void;
+  isInputDisabled: boolean;
+  inputRef: React.RefObject<HTMLInputElement>;
```
alexander-ignatow, line 9: Why not to call useRef hook inside the Header?

src/components/Header.tsx
```diff
@@ -0,0 +1,76 @@
...
+      });
+  };
+
+  return (
+    <header className="todoapp__header">
+      {!isTodosEmpty && (
+        <button
+          type="button"
+          className={classNames('todoapp__toggle-all', {
+            active: allTodosIsComplited,
```
alexander-ignatow, line 55: Why activate the button, that "completes" all todos, when all todos are completed?

src/components/Header.tsx
```diff
@@ -0,0 +1,76 @@
...
+        setErrorMessage('');
+      })
+      .catch(() => {
+        setErrorMessage(ErrorMessage.ADD);
+      });
+  };
+
+  return (
+    <header className="todoapp__header">
+      {!isTodosEmpty && (
```
alexander-ignatow: It is better and "softer" just to disable the button, not remove it from rendering
And you're not consistent - the button has no effect either on empty todos list of on "all todos completed' - but you act differentely

I would replace allTodosIsComplited and  isTodosEmpty with single prop, ex enableTodosMassCompletion" or something like this

src/components/Header.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <button
+          type="button"
+          className={classNames('todoapp__toggle-all', {
+            active: allTodosIsComplited,
+          })}
+          data-cy="ToggleAllButton"
+          onClick={() => toggleTodos()}
```
alexander-ignatow:
```suggestion
          onClick={toggleTodos}
```

alexander-ignatow (changes requested):
Better, but some more effort is needed

ArturErmolenko (approved):

PR #1927, opened 2025-09-05
Develop

bebelekele:
[DEMO LINK](https://bebelekele.github.io/react_todo-app-with-api/)


yanaovcharyk (approved):

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { SortTodos } from './types/SortTodos';
+import { ErrorMessage } from './types/ErrorMessage';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortTodos, setSortTodos] = useState<SortTodos>('All');
```
ArturErmolenko, line 17: use enum
why we need this string?

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [processTodoIds, setProcessTodoIds] = useState<number[]>([]);
+
+  const [input, setInput] = useState('');
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [disableInput, setDisable] = useState(false);
```
ArturErmolenko, line 23: do we realy need this state? i believe we should get rid of it

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+
+  const [error, setError] = useState<ErrorMessage>(null);
```
ArturErmolenko, line 25: why null? can we have default enum value like ''?

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+  const [isLoading, setLoader] = useState(false);
```
ArturErmolenko, line 26:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```
you should use same name

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+
+  const [isEditing, setIsEditing] = useState(false);
+  const [changingValue, setChangingValue] = useState('');
```
ArturErmolenko, line 29:
```suggestion
  const [changingValue, setChangingValue] = useState('');
```
?? let's get rid of this extra state

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+
+  function sorterTodos(sortStatus: SortTodos) {
+    switch (sortStatus) {
+      case 'Active':
```
ArturErmolenko, line 33: enum

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+        return todos.filter((t: Todo) => !t.completed);
```
ArturErmolenko, line 34: t?

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+
+  function loadTodos() {
+    todoService
+      .getTodos()
+      .then(data => {
+        setTodos(data);
+        setTimeout(() => inputRef.current?.focus(), 0);
+      })
+      .catch(() => {
+        setError('LOAD');
```
ArturErmolenko, line 50: enum

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+      });
+  }
+
+  function creationOfTodo(title: string) {
```
ArturErmolenko, line 54: use arrow function

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+    setDisable(true);
+    setLoader(true);
```
ArturErmolenko, line 56: looks like you repeat yourself
DRY

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+    setTempTodo(temp);
+
+    todoService
+      .createTodo({ title })
+      .then(createdTodo => {
+        setTodos(prev => [...prev, createdTodo]);
+        setInput('');
+      })
+      .catch(() => {
+        setError('ADD');
```
ArturErmolenko, line 74: enum

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setError(null);
+    }, 3000);
+  }, [error]);
```
ArturErmolenko, line 187: we should handle timer in ErrorComponent

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+          input={input}
+          disableInput={disableInput}
+          inputRef={inputRef}
+          setInput={value => setInput(value)}
+          makeAllTodoComplete={makeAllTodoComplete}
+          handleSubmit={handleSubmit}
+        />
+
+        <section className="todoapp__main" data-cy="TodoList">
+          {sorterTodos(sortTodos).map(todo => {
```
ArturErmolenko, line 209: move to const

ArturErmolenko: don't call functions in html

src/App.tsx
```diff
@@ -1,26 +1,250 @@
...
+            return (
+              <TodoRow
```
ArturErmolenko, line 211: is it row? i don't see table

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,50 @@
...
+          Unable to delete a todo
+          <br />
+        </>
+      )}
+      {error === 'UPDATE' && (
+        <>
+          Unable to update a todo
+          <br />
+        </>
+      )}
```
ArturErmolenko, line 47: thts because errorMessage should be enum
```suggestion
      {errorMessage}
      )}
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+          Active
+        </a>
+
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: sortTodos === 'Completed',
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setSortTodos('Completed')}
```
ArturErmolenko, line 54: use enum values instead of magic strings

src/components/TodoRow.tsx
```diff
@@ -0,0 +1,123 @@
...
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+          onChange={() => {
+            makeTodoComplete(todo);
+          }}
+        />
+      </label>
+
+      {isEditing && chosenTodoIds.includes(todo.id) ? (
```
ArturErmolenko, line 63: const

src/types/ErrorMessage.ts
```diff
@@ -0,0 +1,7 @@
+export type ErrorMessage =
```
ArturErmolenko, line 1: why we need type?

src/types/SortTodos.ts
```diff
@@ -0,0 +1 @@
+export type SortTodos = 'All' | 'Active' | 'Completed';
```
ArturErmolenko, line 1: why not enum?)

ArturErmolenko (changes requested):
GJ 👍 
Let's improve a bit

PR #1932, opened 2025-09-11
Implemented Todo app

nikitachupahin:
[DEMO LINK](https://nikitachupahin.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,282 +1,277 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useCallback, useEffect, useRef, useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import {
-  addTodos,
   getTodos,
+  USER_ID,
+  addTodo,
   deleteTodo,
   updateTodo,
-  USER_ID,
 } from './api/todos';
 import { Todo } from './types/Todo';
-import { FilterType } from './types/FilterType';
-import { ErrorType } from './types/ErrorType';
+import { TodoList } from './components/TodoList';
+import { Notification } from './components/ErrorNotification';
+import { Footer } from './components/Footer';
+import { NewTodo } from './components/NewTodo';
 import classNames from 'classnames';
+import { ErrorMessages } from './constants/ErrorMessages';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [isLoading, setIsLoading] = useState(false);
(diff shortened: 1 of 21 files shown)


src/App.tsx
```diff
@@ -1,26 +1,498 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<ErrorType>(ErrorType.DEFAULT_VALUE);
+  const [filter, setFilter] = useState<FilterType>(FilterType.ALL);
+  const [title, setTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
+  const [updatingTodoIds, setUpdatingTodoIds] = useState<number[]>([]);
```
vitalii-bondarenko-ua:
```suggestion
  const [processingTodoIds, setProcessingTodoIds] = useState<number[]>([]);
```
it can be one state.

src/App.tsx
```diff
@@ -1,26 +1,498 @@
...
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
+  const [editingTitle, setEditingTitle] = useState('');
```
vitalii-bondarenko-ua: App component should be splitting into separate smaller components (Header, Footer, TodoList, TodoItem, ErrorNotifcation etc.)

This states can live in TodoItem component. App shouldn't know about editingTitle of some component.
And instead editingTodoId can be isEditign boolean state in TodoItem.

src/App.tsx
```diff
@@ -1,26 +1,498 @@
...
+      >
+        <button
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={hideError}
+        />
+        {error}
+      </div>
+    </div>
```
vitalii-bondarenko-ua, line 328: Split App component into smaller separate components.
Think about logic which you can move from App into separate smaller components.

vitalii-bondarenko-ua (changes requested):
Good job.
Please check comments.
And you should pass all tests without skip them in page.spec.js.
page.spec.js shouldn't have and changes in PR.
Screenshot 2025-09-13 at 09 43 51

etojeDenys (changes requested):
after changing the status of several todos, we only see the loader on the first one. the same problem occurs when deleting

nikitachupahin:
Thank you for your feedback. If I understand you correctly, I have already implemented this. When the status changes, a loader should appear on each todo according to the task. Below are screenshots showing that it works. Is this what you meant? 
Знімок екрана 2025-09-13 о 22 07 15
image

Anton-Kuchmasov (approved):
Well done!

PR #1942, opened 2025-09-17
Added solution

Yurii-Kupchynskyi:
Demo Link : https://yurii-kupchynskyi.github.io/react_todo-app-with-api/


src/App.tsx
```diff
@@ -1,26 +1,443 @@
...
+  };
+
+  const handleRenameTodo = async (todo: Todo) => {
+    const trimmed = editingTitle.trim();
+
+    if (!trimmed) {
+      handleDeleteTodo(todo.id);
+
+      return;
+    }
```
alexander-ignatow, line 221: Please explain
Why do you delete TODO when renaming it to an empty title? These are two completely different ops

alexander-ignatow (changes requested):
PLease fix the issues in the code

Also break App.tsx into the components
- <header> functionality can be moved to a separate component
- editing and displaying todo also should be moved to a separate component

PR #1958, opened 2025-10-02
add task solution

Anna-0805:
https://Anna-0805.github.io/react_todo-app-with-api


etojeDenys (changes requested):
1. after trying to change the status of several todos, the loader only shows up on the last one, but it should show up on every todo. the same problem with deleting

2. after pressing "toggle all" and "clear completed" a loader should appear on the todos that will change status or be deleted

vadiimvooo (approved):
Well done

PR #1969, opened 2025-10-10
add solution, it took me 3 days

Aiiyuu:
[DEMO LINK](https://aiiyuu.github.io/react_todo-app-with-api/)


src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,32 @@
...
+type Props = {
+  currentMethod: TodoFilterMethod;
+  onSelect: (method: TodoFilterMethod) => void;
+};
+
+const Filter: React.FC<Props> = ({ currentMethod, onSelect }: Props) => {
+  return (
+    <nav className="filter" data-cy="Filter">
+      {Object.entries(TodoFilterMethod)
+        .filter(([key]) => key !== 'Default')
```
781calmstate, line 14: i'd recommend to do this kind of computations before return in the component body for more readability


```suggestion
      const filteredMethods = Object.entries(TodoFilterMethod)
        .filter(([key]) => key !== 'Default');
        
      return (... 
      {filteredMethods.map(...)
      )
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,45 @@
...
+  todos,
+  showLoadingFor,
+}: Props) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos.map(todo => {
+        const isLoading =
+          (todo.completed && showLoadingFor === 'completed') ||
+          (!todo.completed && showLoadingFor === 'uncompleted');
+
```
781calmstate, line 28: why can't you make the same computation but inside of TodoItem? this will make your map much cleaner

781calmstate (approved):
I suppose you just copied the code from previous task, because I can see some mistakes I mentioned in your previous PR I reviewed.

Consider fixing all the problems before moving to the next task.

Pay more attention to commit naming, `add solution, it took me 3 days)` isn't a good naming :)
More about commits naming you can read [here](https://www.conventionalcommits.org/en/v1.0.0/)

Aiiyuu:
> I suppose you just copied the code from previous task, because I can see some mistakes I mentioned in your previous PR I reviewed.

That’s the whole point of this series of tasks — you need to implement a to-do app, and for each part, you should copy your previous code and add new features.

PR #1970, opened 2025-10-11
Task Solution

Vovanchu:
[DEMO LINK](https://Vovanchu.github.io/react_todo-app-with-api/)


src/types/TempTodo.ts
```diff
@@ -0,0 +1,7 @@
+export interface TempTodo {
```
Anton-Kuchmasov, line 1: You can only `extend` Todo interface to use inheritance
image

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todos: Todo[];
+  todosLength: number;
+  filterType: string;
+  setFilterType: (filter: 'all' | 'active' | 'completed') => void;
```
Anton-Kuchmasov: use created `FilterType` here

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todos: Todo[];
+  todosLength: number;
```
Anton-Kuchmasov: looks overloaded.
Would you mind avoid this prop, using instead `todos.length` inside a `TodoFooter` component?

Anton-Kuchmasov (changes requested):
Well done!

Check out comments below and make minor changes

etojeDenys (changes requested):
1. after deleting several todos in a short period of time, the loader only shows up on the last todo pressed.  the same problem applies to changing status

2. after pressing the "toggle all" button, the loader should appear on every todo that will change status

vadiimvooo (approved):
Well done

PR #1973, opened 2025-10-15
add loader and toggle

Moddderi:
[DEMO LINK](https://Moddderi.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -193,7 +193,11 @@ export const App: React.FC = () => {
     }
   };
 
-  const handleUpdateTodo = async (todoId: number, title: string) => {
+  const handleUpdateTodo = async (
+    todoId: number,
+    title: string,
+    setEditing: React.Dispatch<React.SetStateAction<boolean>>,
+  ) => {
     setUpdatingTodoIds(prev => [...prev, todoId]);
     setError(null);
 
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,26 +1,261 @@
...
+    StatusFilter.All,
+  );
+  const [, setLoading] = useState(false);
+  const [, setIsActionLoading] = useState(false);
+  const [error, setError] = useState<ErrorType | null>(null);
+  const [newTitle, setNewTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
+  const [updatingTodoIds, setUpdatingTodoIds] = useState<number[]>([]);
```
vitalii-bondarenko-ua, line 38: it can be one state instead two (processingTodoIds for example)

src/App.tsx
```diff
@@ -1,26 +1,261 @@
...
+  Completed = 'completed',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [statusFilter, setStatusFilter] = useState<StatusFilter>(
+    StatusFilter.All,
+  );
+  const [, setLoading] = useState(false);
+  const [, setIsActionLoading] = useState(false);
```
vitalii-bondarenko-ua, line 32: looks like redundant states. why do you need this states, if you do not use them?

vitalii-bondarenko-ua (approved):
Looks good

PR #1987, opened 2025-10-30
Task switching and renaming has been implemented

DmytroLymar:
[DEMO LINK](https://DmytroLymar.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/hooks/useProcessing.ts
+++ b/src/hooks/useProcessing.ts
@@ -1,0 +1,39 @@
+import { useState } from 'react';
+
+export const useProcessing = () => {
+  const [processingIds, setProcessingIds] = useState<Set<number>>(new Set());
+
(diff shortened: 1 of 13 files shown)


Volodymyr-Bauzi (approved):
Good job!

src/hooks/useProcessing.ts
```diff
@@ -0,0 +1,39 @@
...
+
+      return n;
+    });
+
+  return [
+    processingIds,
+    addProcessing,
+    removeProcessing,
+    addManyProcessing,
+  ] as const;
```
781calmstate, line 38: it make sense to return array from custom hooks if you have 2 or 3 items returned. in case if there are more I'd suggest to return object

781calmstate (approved):

PR #1991, opened 2025-10-31
initial commit

pshy4ok:
[DEMO LINK](https://pshy4ok.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,412 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
+import cn from 'classnames';
+import {
+  createTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+  USER_ID,
+} from './api/todos';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { FILTERS, FilterType } from './constants/filters';
+import { NewTodo } from './components/NewTodo';
+import { TodoList } from './components/TodoList';
+import { Filter } from './components/Filter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState('');
+  const [notificationVisible, setNotificationVisible] = useState(false);
+  const [filter, setFilter] = useState<FilterType>(FILTERS.all);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [nextId, setNextId] = useState<number>(1);
+  const [adding, setAdding] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletingIds, setDeletingIds] = useState<Set<number>>(new Set());
+  const [updatingIds, setUpdatingIds] = useState<Set<number>>(new Set());
+  const [focusTick, setFocusTick] = useState(0);
+
+  const timerRef = useRef<NodeJS.Timeout | null>(null);
+
+  const showError = (message: string) => {
+    setError(message);
+    setNotificationVisible(true);
+
+    if (timerRef.current) {
+      clearTimeout(timerRef.current);
+    }
+
+    timerRef.current = setTimeout(() => {
+      setNotificationVisible(false);
+    }, 3000);
+  };
+
+  useEffect(() => {
+    const fetchTodos = async () => {
+      setLoading(true);
+
+      if (timerRef.current) {
+        clearTimeout(timerRef.current);
+      }
+
+      setNotificationVisible(false);
+      setError('');
+
+      try {
+        const data = await getTodos();
+
+        setTodos(data);
+        const maxId = data.reduce(
+          (max: number, t: { id: number }) => Math.max(max, t.id),
+          0,
+        );
+
+        setNextId(maxId + 1);
+      } catch (e) {
+        if (e instanceof Error && e.message.trim() !== '') {
+          showError(e.message);
+        } else {
+          showError('Unable to load todos');
+        }
+      } finally {
+        setLoading(false);
+      }
+    };
+
+    fetchTodos();
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-add-and-delete#react-todo-app-add-and-delete">
-          React Todo App - Add and Delete
-        </a>
-      </p>
+  const visibleTodos = todos.filter(todo => {
+    switch (filter) {
+      case FILTERS.active:
+        return !todo.completed;
+      case FILTERS.completed:
+        return todo.completed;
+      default:
+        return true;
+    }
+  });
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+  const allCompleted = todos.length > 0 && todos.every(t => t.completed);
+
+  const handleToggle = async (id: number, nextCompleted: boolean) => {
+    setUpdatingIds(prev => new Set(prev).add(id));
+
+    try {
+      const updated = await updateTodo(id, { completed: nextCompleted });
+
+      setTodos(prev =>
+        prev.map(t =>
+          t.id === id ? { ...t, completed: updated.completed } : t,
+        ),
+      );
+    } catch (e) {
+      showError('Unable to update a todo');
+    } finally {
+      setUpdatingIds(prev => {
+        const next = new Set(prev);
+
+        next.delete(id);
+
+        return next;
+      });
+    }
+  };
+
+  const handleDelete = async (id: number) => {
+    setDeletingIds(prev => new Set(prev).add(id));
+
+    try {
+      await deleteTodo(id);
+      setTodos(prev => prev.filter(t => t.id !== id));
+    } catch (e) {
+      showError('Unable to delete a todo');
+    } finally {
+      setDeletingIds(prev => {
+        const next = new Set(prev);
+
+        next.delete(id);
+
+        return next;
+      });
+      setFocusTick(t => t + 1);
+    }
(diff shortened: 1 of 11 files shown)


MariaAlefirenko (approved):
Good job! (＾▽＾)／
But you still need to add a placeholder (as a warning for better UX) in the input field when the user deletes the entire title while editing a todo.

src/components/Filter.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import cn from 'classnames';
+import { FILTERS, FilterType } from '../constants/filters';
+
+type Props = {
+  current: FilterType;
```
781calmstate: current what?)

src/components/NewTodo.tsx
```diff
@@ -0,0 +1,48 @@
...
+  focusTick = 0,
+}) => {
+  const [title, setTitle] = useState('');
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (!disabled) {
+      inputRef.current?.focus();
+    }
+  }, [disabled, focusTick]);
```
781calmstate, line 21: why do you need to depend on `focusTick` here?

src/components/NewTodo.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const handleSubmit = async (e: React.FormEvent) => {
+    e.preventDefault();
+
+    const success = await onSubmit(title);
+
+    if (success) {
+      setTitle('');
+    }
```
781calmstate: success can be true or false which means that the variable name is unclear. Consider renaming it to something more suitable for boolean value, e.g. `isSuccessful`

src/components/Todo.tsx
```diff
@@ -0,0 +1,135 @@
...
+    setIsEditing(true);
+  };
+
+  const cancelEditing = () => {
+    setTitle(todo.title);
+    setIsEditing(false);
+  };
+
+  const commitEditing = async () => {
+    const trimmed = title.trim();
```
781calmstate: trimmed what?:)

src/components/Todo.tsx
```diff
@@ -0,0 +1,135 @@
...
+        >
+          {todo.title}
+        </span>
+      )}
+
+      {isEditing && (
+        <form
+          onSubmit={e => {
+            e.preventDefault();
+            void commitEditing();
```
781calmstate: why do you need `void` here?

src/components/Todo.tsx
```diff
@@ -0,0 +1,135 @@
...
+          }}
+        >
+          <input
+            ref={inputRef}
+            data-cy="TodoTitleField"
+            type="text"
+            className="todo__title-field"
+            value={title}
+            onChange={e => setTitle(e.target.value)}
+            onBlur={() => void commitEditing()}
```
781calmstate: why do you need `void` here?

src/components/Todo.tsx
```diff
@@ -0,0 +1,135 @@
...
+    setTitle(todo.title);
+    setIsEditing(true);
+  };
+
+  const cancelEditing = () => {
+    setTitle(todo.title);
+    setIsEditing(false);
+  };
+
+  const commitEditing = async () => {
```
781calmstate: event handlers should have `handle` prefix before the name

src/components/Todo.tsx
```diff
@@ -0,0 +1,135 @@
...
+      return;
+    }
+
+    if (trimmed === todo.title) {
+      setIsEditing(false);
+
+      return;
+    }
+
+    const ok = await (onUpdateTitle?.(todo.id, trimmed) ?? Promise.resolve(false));
```
781calmstate: `ok` is a bad name for variable :)

src/components/Todo.tsx
```diff
@@ -0,0 +1,135 @@
...
+            type="text"
+            className="todo__title-field"
+            value={title}
+            onChange={e => setTitle(e.target.value)}
+            onBlur={() => void commitEditing()}
+            onKeyUp={e => {
+              if (e.key === 'Escape') {
+                cancelEditing();
+              }
+            }}
```
781calmstate: inline callbacks that takes more than one line usually should be moved to handler

src/App.tsx
```diff
@@ -1,26 +1,412 @@
...
+  const visibleTodos = todos.filter(todo => {
+    switch (filter) {
+      case FILTERS.active:
+        return !todo.completed;
+      case FILTERS.completed:
+        return todo.completed;
+      default:
+        return true;
+    }
+  });
```
781calmstate: may and should be moved into helper

781calmstate: wouldn't it be simpler to read your component when you see `const visibleTodos = getFilteredTodos(...)` instead of that?

src/App.tsx
```diff
@@ -1,26 +1,412 @@
...
+
+  const allCompleted = todos.length > 0 && todos.every(t => t.completed);
+
+  const handleToggle = async (id: number, nextCompleted: boolean) => {
+    setUpdatingIds(prev => new Set(prev).add(id));
+
+    try {
+      const updated = await updateTodo(id, { completed: nextCompleted });
```
781calmstate: updated what?

src/App.tsx
```diff
@@ -1,26 +1,412 @@
...
+      setTodos(prev =>
+        prev.map(t =>
+          t.id === id ? { ...t, completed: updated.completed } : t,
+        ),
+      );
+    } catch (e) {
+      showError('Unable to update a todo');
+    } finally {
+      setUpdatingIds(prev => {
+        const next = new Set(prev);
```
781calmstate: next what?)

src/App.tsx
```diff
@@ -1,26 +1,412 @@
...
+
+  const handleToggleAll = async () => {
+    if (todos.length === 0) {
+      return;
+    }
+
+    const nextCompleted = !allCompleted;
+    const toChange = todos
+      .filter(t => t.completed !== nextCompleted)
+      .map(t => t.id);
```
781calmstate: when I read to `toChange` for the first time I am really struggling to understand what can it possibly mean.

781calmstate (changes requested):
This is good solution, but you really should work on your naming.  I don't have to have a need going and starting to analyze your code to understand the purpose of a variable. As your teammate I wouldn't be happy with that. The variables/functions/props etc. should be named as clear and descriptive as possible, so your teammates will have to analyze the code only when they need it instead of when they just read the code and trying to figure out what the name of this variable means.

src/components/NewTodo.tsx
pshy4ok: This focusTick is needed if you want to force the field to refocus even when disabled doesn't change. I had to add this to make all the tests pass.

src/components/Todo.tsx
pshy4ok: void is used here to ignore the Promise returned by commitEditing() (because it is async).
Without void ESLint shows the error: "Promise returned in function argument was not handled".

src/api/todos.ts
```diff
@@ -0,0 +1,38 @@
...
+export const getFilteredTodos = (todos: Todo[], filter: FilterType) => {
+  switch (filter) {
+    case FILTERS.active:
+      return todos.filter(todo => !todo.completed);
+    case FILTERS.completed:
+      return todos.filter(todo => todo.completed);
+    default:
+      return todos;
+  }
+};
```
781calmstate: I am not sure the helper function should be placed in the same file as api functions

src/api/todos.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const createTodo = (title: string) => {
+  const trimmedTitle = title.trim();
```
781calmstate: I don't think API function should care about trimming anything, I think you should trim the title before passing to API function

src/components/Filter.tsx
```diff
@@ -0,0 +1,46 @@
...
+import cn from 'classnames';
+import { FILTERS, FilterType } from '../constants/filters';
+
+type Props = {
+  currentFilter: FilterType;
+  onChange: (next: FilterType) => void;
+};
+
+export const Filter: React.FC<Props> = ({
+  currentFilter: current,
```
781calmstate: still current :(

src/App.tsx
```diff
@@ -1,26 +1,404 @@
...
-
-const USER_ID = 0;
+import { FILTERS, FilterType } from './constants/filters';
+import { NewTodo } from './components/NewTodo';
+import { TodoList } from './components/TodoList';
+import { Filter } from './components/Filter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
```
781calmstate: boolean variables/props should have specific prefixes - `hasError`, `isVisible`, `areEqual`, `shouldGenerate` etc

src/App.tsx
```diff
@@ -1,26 +1,404 @@
...
+  const [error, setError] = useState('');
+  const [notificationVisible, setNotificationVisible] = useState(false);
+  const [filter, setFilter] = useState<FilterType>(FILTERS.all);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [nextId, setNextId] = useState<number>(1);
+  const [adding, setAdding] = useState(false);
```
781calmstate: boolean variables/props should have specific prefixes - `hasError`, `isVisible`, `areEqual`, `shouldGenerate` etc

src/App.tsx
```diff
@@ -1,26 +1,404 @@
...
+import { FILTERS, FilterType } from './constants/filters';
+import { NewTodo } from './components/NewTodo';
+import { TodoList } from './components/TodoList';
+import { Filter } from './components/Filter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState('');
+  const [notificationVisible, setNotificationVisible] = useState(false);
```
781calmstate: boolean variables/props should have specific prefixes - `hasError`, `isVisible`, `areEqual`, `shouldGenerate` etc

781calmstate (changes requested):
Your naming is much better now, but there's still room to grow! 👍

DanilWeda (approved):
Many thanks for changes! 
Approved!

781calmstate (approved):

PR #1994, opened 2025-10-31
third part of todo app

MarinaKramarchuk:
[DEMO LINK](https://MarinaKramarchuk.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import cn from 'classnames';
 import { UserWarning } from './UserWarning';
 import {
@@ -34,7 +34,7 @@ export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [error, setError] = useState<string>('');
   const [filter, setFilter] = useState<Filters>(Filters.All);
-  const [loading, setLoading] = useState(true);
+  const [isLoading, setIsLoading] = useState(true);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
   const [editingTodoId, setEditingTodoId] = useState<Todo['id'] | null>(null);
@@ -48,7 +48,7 @@ export const App: React.FC = () => {
       .catch(() => {
         setError('Unable to load todos');
       })
-      .finally(() => setLoading(false));
+      .finally(() => setIsLoading(false));
   }, []);
 
   useEffect(() => {
@@ -63,13 +63,10 @@ export const App: React.FC = () => {
     return () => clearTimeout(timer);
   }, [error]);
 
-  useEffect(() => {
-    const input =
-      document.querySelector<HTMLInputElement>('.todoapp__new-todo');
+  const inputRef = useRef<HTMLInputElement>(null);
 
-    if (input) {
-      input.focus();
-    }
+  useEffect(() => {
+    inputRef.current?.focus();
   }, [todos]);
 
   if (!USER_ID) {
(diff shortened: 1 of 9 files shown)


src/components/TodoComp.tsx
```diff
@@ -0,0 +1,99 @@
...
+        >
+          <input
+            data-cy="TodoTitleField"
+            type="text"
+            className="todo__title-field"
+            placeholder="Empty todo will be deleted"
+            value={query}
+            onChange={event => onSetQuery(event.target.value.trimStart())}
+            onBlur={() => onUpdateTodo(todo.id, query)}
+            onKeyUp={event => {
```
MaksimShe: for better readeble (in my opinion) u can create func like handleExitFromField() an in this func can be that "if (eve..." btw, i think this variant is ok

MaksimShe: and also to separate UI and logic as much as possible

MaksimShe (approved):
nice wrk!

src/components/TodoComp.tsx
781calmstate: that is a great idea. whenever your inline callback takes more than one line - it means that it's probably a good idea to move it into separated handler function

src/components/Filter.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import cn from 'classnames';
+import { Filters } from '../constants/filter';
+
+type Props = {
+  filter: Filters;
+  onSetFilter: (filter: Filters) => void;
```
781calmstate: this name works just fine, but imho `onFilterChange` would be a little bit better

src/App.tsx
```diff
@@ -1,26 +1,302 @@
...
+    default:
+      return newTodos;
+  }
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string>('');
+  const [filter, setFilter] = useState<Filters>(Filters.All);
+  const [loading, setLoading] = useState(true);
```
781calmstate:
```suggestion
  const [isLoading, setIsLoading] = useState(true);
```

[Tips on naming boolean variables](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig)

src/App.tsx
```diff
@@ -1,26 +1,302 @@
...
+  }, [error]);
+
+  useEffect(() => {
+    const input =
+      document.querySelector<HTMLInputElement>('.todoapp__new-todo');
+
+    if (input) {
+      input.focus();
+    }
+  }, [todos]);
```
781calmstate: could be done with the `ref` by creating it here and passing as prop to your input. it would look much more like React solution. Usually you want to avoid using document.querySelector in React as much as possible and utilize the useRef instead

src/App.tsx
```diff
@@ -1,26 +1,302 @@
...
+    deleteTodo(id)
+      .then(() => {
+        setTodos(prev => prev.filter(todo => todo.id !== id));
+      })
+      .catch(() => {
+        setError('Unable to delete a todo');
+      })
+      .finally(() =>
+        setLoadingTodoIds(prev => prev.filter(loadingId => loadingId !== id)),
+      );
```
781calmstate: here you use `.then.catch` approach to handle promise, but above you use `try...catch` with `async...await`.
I suggest picking one of these approaches and stick to it. I'd recommend `async await` with `try catch` - much easier to read

781calmstate (changes requested):
Overall, well done!

src/components/TodoComp.tsx
MarinaKramarchuk: Fixed, moved the handler to a separate function.

src/components/Filter.tsx
MarinaKramarchuk: сhanged the name

src/App.tsx
MarinaKramarchuk: changed the variable name

src/App.tsx
MarinaKramarchuk: changed the logic, used ref and passed props to the component newTodo

src/App.tsx
MarinaKramarchuk: applied the same syntax to all promises

781calmstate (approved):
Well done!

PR #1996, opened 2025-10-31
Dev

MaksimShe:
[DEMO LINK](https://MaksimShe.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,8 +14,8 @@ import { filterTodos } from './utils/fiterTodos';
 import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
-  const [isDisabledInput, setDisableInput] = useState<boolean>(false);
-  const [idTodoLoading, setIdTodoLoading] = useState<number[]>([]);
+  const [isDisabledInput, setIsDisableInput] = useState<boolean>(false);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
   const [preparedTodos, setPreparedTodos] = useState<Todo[]>([]);
   const [activeFilterStatus, setActiveFilterStatus] = useState<FilterStatus>(
     FilterStatus.All,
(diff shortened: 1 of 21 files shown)


MaksimShe:
i think this test like - random() > 0.5 ? approve : reject. Because on localhost my last tests -on img, but before (i change nothing) only this test fail.


image

Bohdan-Diatliuk (approved):
like

src/App.tsx
```diff
@@ -1,26 +1,102 @@
...
+import { FilterStatus, ErrorMessages } from './types';
+import { Todo, USER_ID } from './types';
+import { TodoHeader } from './components/TodoHeader';
+import { TodoFooter } from './components/TodoFooter';
+import { TodoList } from './components/TodoList';
+import { filterTodos } from './utils/fiterTodos';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
+  const [isDisabledInput, setDisableInput] = useState<boolean>(false);
```
781calmstate:
```suggestion
  const [isDisabledInput, setIsDisabledInput] = useState<boolean>(false);
```

set function should have same name as variable but with `set`

src/App.tsx
```diff
@@ -1,26 +1,102 @@
...
+  const [idTodoLoading, setIdTodoLoading] = useState<number[]>([]);
```
781calmstate:
```suggestion
  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
```

781calmstate (changes requested):
Overall great solution. But I want you to understand the next:
## Why Use Callbacks Instead of Direct setState Props
### 1. Abstraction and Encapsulation

When you pass a callback like onError={(error) => setError(error)} instead of setError directly, you:
- Hide implementation details from child components
- Create a clear contract about what the child can do
- Maintain control over how state changes happen

### 2. Flexibility and Future-Proofing

Callbacks allow you to add logic without breaking child components:
```
// With callback - easy to extend
const handleError = (error: string) => {
  setError(error);
  logError(error); // Added later without breaking children
  trackAnalytics('error', error); // More logic added easily
};

// vs. passing setState directly - harder to extend
<Child setError={setError} /> // Child controls everything
```

### 3. Better Component Boundaries

-  Child components don't need to know about React's state management
-  Reduces coupling between parent and child
-  Makes components more reusable and testable

781calmstate (approved):
GJ!

PR #1997, opened 2025-10-31
solution

livadin:
[DEMO LINK](https://livadin.github.io/react_todo-app-with-api/)


vtmnkmarta (approved):
Good job!👍

src/components/NewTodoForm/NewTodoForm.tsx
```diff
@@ -0,0 +1,41 @@
...
+  value,
+  onChange,
+  onSubmit,
+  disabled,
+}) => {
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    inputRef.current?.focus();
+  }, [disabled]);
```
781calmstate: does it make sense to focus input only when it's not disabled?

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    try {
+      if (!trimmedTitle) {
+        await onDelete?.(todo.id);
+      } else {
+        await onUpdate?.(todo.id, { title: trimmedTitle });
+      }
+
+      setIsEditing(false);
+    } catch {}
```
781calmstate: you probably should handle the error somehow, because otherwise if the action fail no one will know the reason, neither user nor you

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+  const allCompleted = hasTodos && itemsLeft === 0;
+
+  const filteredTodos = useMemo(() => {
+    switch (filter) {
+      case Filters.ACTIVE:
+        return todos.filter(t => !t.completed);
+      case Filters.COMPLETED:
+        return todos.filter(t => t.completed);
+      case Filters.ALL:
+        return todos;
```
781calmstate, line 77: would look cleaner if this functionality was moved to helper

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+      setNewTitle('');
+    } catch {
+      setErrorMessage(ERROR_TEXT.add);
+    } finally {
+      setTempTodo(null);
+      setIsAdding(false);
+
+      document
+        .querySelector<HTMLInputElement>('[data-cy="NewTodoField"]')
+        ?.focus();
```
781calmstate: it's bad practice to use document.querySelector in react, try using useRef instead

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+
+    try {
+      await deleteTodo(id);
+      setTodos(prev => prev.filter(todo => todo.id !== id));
+      document
+        .querySelector<HTMLInputElement>('[data-cy="NewTodoField"]')
+        ?.focus();
+    } catch (e) {
+      setErrorMessage(ERROR_TEXT.delete);
+      throw e;
```
781calmstate: make sure you handle this error you throwing, because if it's unhandled, it might crash your web-app

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+    hideError();
+    setUpdatingIds(prev => new Set(prev).add(id));
+
+    try {
+      const updatedTodo = await updateTodo(id, data);
+
+      setTodos(prev => prev.map(todo => (todo.id === id ? updatedTodo : todo)));
+    } catch (e) {
+      setErrorMessage(ERROR_TEXT.update);
+      throw e;
```
781calmstate: same here

781calmstate (changes requested):

src/hooks/useProcessingIds.ts
```diff
@@ -0,0 +1,59 @@
+import { useState, useMemo } from 'react';
+
+export type UseProcessingReturn = {
+  processingIds: Set<number>;
+  addProcessing: (id: number) => void;
+  removeProcessing: (id: number) => void;
+  addManyProcessing: (ids: number[]) => void;
+  isProcessing: (id: number) => boolean;
+};
```
alexander-ignatow, line 9: Why you need this type?
And why you export it?

src/hooks/useProcessingIds.ts
```diff
@@ -0,0 +1,59 @@
...
+      if (ids.length === 0) {
+        return prev;
+      }
+
+      const newSet = new Set(prev);
+
+      ids.forEach(id => newSet.add(id));
+
+      return newSet;
+    });
```
alexander-ignatow, line 43: You can unite two functions into one

```suggestion
  const addProcessing = (ids: number | number[]) =>
    const idsToProcess = Array.isArray(ids) ? ids : [ids];
    
    setProcessingIds(prev => {
      if (idsToProcess.length === 0) {
        return prev;
      }

      const newSet = new Set(prev);

      idsToProcess.forEach(id => newSet.add(id));

      return newSet;
    });
```

alexander-ignatow (approved):
Looks good!

You can check the issues to futher improve the code - but nothing critical there

PR #2005, opened 2025-11-03
add task solution

Julia-Hertseva:
[DEMO LINK](https://Julia-Hertseva.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,12 +10,12 @@ import {
   deleteTodo,
   updateTodo,
 } from './api/todos';
-import { Todo } from './types/Todo';
+import { Todo } from './types/todo';
 import { useState } from 'react';
 // import { client } from './utils/fetchClient';
 import cn from 'classnames';
-import { TodoFilter } from './types/Filters';
-import { ErrorTypes } from './types/ErrorTypes';
+import { TodoFilter } from './types/filters';
+import { ErrorTypes } from './types/errorTypes';
 import { TodoList } from './components/TodoList';
 import { Footer } from './components/Footer/Footer';
 import { NewTodo } from './components/NewTodo';
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,17 +18,18 @@ import { ErrorTypes } from './types/errorTypes';
 import { TodoList } from './components/TodoList';
 import { Footer } from './components/Footer/Footer';
 import { NewTodo } from './components/NewTodo';
+import { getFilteredTodos } from './utils/getFilteredTodos';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [loading, setLoading] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
   const [error, setError] = useState<string | null>(null);
   const [filter, setFilter] = useState<TodoFilter>(TodoFilter.All);
   const [newTodo, setNewTodo] = useState('');
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [isAdding, setIsAdding] = useState(false);
-  const [deletingIds, setDeletingIds] = useState<number[]>([]);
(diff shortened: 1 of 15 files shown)


ViktoriiaPitsan (approved):

781calmstate:
Fix lint issues. Git case insensitive which means if you commited the file that was named `Something` and then renamed it to `something` git wouldn't recognize any changes. To file name where you changed the case you need delete and create a new file

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,141 @@
...
+import '../../styles/todo.scss';
+
+type Props = {
+  todo: TodoType;
+  onDeleteTodo: (id: number) => void;
+  deletingIds: number[];
+  onUpdateTodo: (id: number, updatedTodo: Partial<TodoType>) => void;
+  updatingIds?: number[];
+  editingId: number | null;
+  setEditingId: (id: number | null) => void;
```
781calmstate: the props names shouldn't mix up with `useState` names which means you don't want to name the props similar to how you name setState function.

I suggest `onEditingIdChange` or something similar

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+type Props = {
+  visibleTodos: TodoType[];
+  tempTodo: TodoType | null;
+  onDeleteTodo: (id: number) => void;
+  deletingIds: number[];
+  updatingIds: number[];
+  onUpdateTodo: (id: number, updatedTodo: Partial<TodoType>) => void;
+  editingId: number | null;
+  setEditingId: (id: number | null) => void;
```
781calmstate: same here

src/App.tsx
```diff
@@ -1,26 +1,318 @@
...
+import cn from 'classnames';
+import { TodoFilter } from './types/filters';
+import { ErrorTypes } from './types/errorTypes';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { NewTodo } from './components/NewTodo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
```
781calmstate: boolean variables/props should have specific prefixes - `hasError`, `isVisible`, `areEqual`, `shouldGenerate` etc

src/App.tsx
```diff
@@ -1,26 +1,318 @@
...
+        setTodos(receivedTodos);
+      })
+      .catch(() => {
+        setError(ErrorTypes.LoadTodos);
+        // const timeoutId = setTimeout(() => {
+        //   setError(null);
+        // }, 3000);
+
+        // return () => clearTimeout(timeoutId);
+      })
```
781calmstate: do you need it?

src/App.tsx
```diff
@@ -1,26 +1,318 @@
...
+  const visibleTodos = todos.filter((todo: Todo) => {
+    switch (filter) {
+      case TodoFilter.Active:
+        return !todo.completed;
+      case TodoFilter.Completed:
+        return todo.completed;
+      default:
+        return true;
+    }
+  });
```
781calmstate: should be moved to the helper function:

`const visibleTodos = getFilteredTodos(...)`

781calmstate (changes requested):
You are mixing different methods of working with Promises. In one place you use `async await` in another you use `.then().catch()...`. Mixing different approaches in programming usually not the best idea, you should stick to one approach in the most of situations.

I suggest to prioritize `async await` since it's newer and better way to handle promises

Julia-Hertseva:
> Fix lint issues. Git case insensitive which means if you commited the file that was named `Something` and then renamed it to `something` git wouldn't recognize any changes. To file name where you changed the case you need delete and create a new file

I did that, but there were still mistakes... 
and if I add the extension as it asks (.ts), then vsc highlights an error.

781calmstate (approved):

PR #2009, opened 2025-11-04
Develop

ViktoriiaPitsan:
[DEMO LINK](https://ViktoriiaPitsan.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,12 +10,13 @@ import {
 } from './api/todos';
 import cn from 'classnames';
 import { TodoItem } from './components/TodoItem';
-import { TODO_STATUS_FILTER_OPTIONS, Status } from './types/TodoStatusFilter';
+import { Status } from './types/TodoStatusFilter';
 import { getFilteredTodos, Todo } from './types/Todo';
 import { useError } from './hooks/useError';
 import { TodoCreate } from './types/TodoCreate';
-import { TodoCreateForm } from './components/TodoCreateForm';
 import { TodoList } from './components/TodoList';
+import { Header } from './components/TodoHeader';
+import { Footer } from './components/TodoFooter';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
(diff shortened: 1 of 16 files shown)


DmytroLymar (approved):

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,144 @@
...
+};
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  isLoading = false,
+  onDelete,
+  onToggleStatus,
+  onSaveTitle,
+}) => {
+  const [editing, setEditing] = useState(false);
```
781calmstate: boolean variables/props should have specific prefixes - `hasError`, `isVisible`, `areEqual`, `shouldGenerate` etc

src/types/Todo.ts
```diff
@@ -0,0 +1,26 @@
...
+        case Status.ACTIVE:
+          return !todo.completed;
+        default:
+          throw new Error('Missing case in getFilteredTodos');
+      }
+    });
+  }
+
+  return filteredTodos;
+};
```
781calmstate: I am against helper functions inside of type's files because in my opinion it doesn't make any sense - why helper function should be located by `src/types/MyType` path?

The choice is yours to leave it like this or move it, I won't force it

src/utils/filterOptions.ts
```diff
@@ -0,0 +1,29 @@
...
+    href: '/completed',
+    testId: 'FilterLinkCompleted',
+    text: 'Completed',
+  },
+  [Status.ACTIVE]: {
+    href: '/active',
+    testId: 'FilterLinkActive',
+    text: 'Active',
+  },
+};
```
781calmstate, line 19: similar as above but for constant

src/utils/filterOptions.ts
781calmstate, line 1: why this file has tsx extension?

src/App.tsx
```diff
@@ -1,26 +1,276 @@
...
+
+  const handleRemoveTodoFromLoading = (todoId: Todo['id']) => {
+    setLoadingTodoIds(currentLoading =>
+      currentLoading.filter(id => id !== todoId),
+    );
+  };
+
+  const getIsTodoLoading = (todoId: Todo['id']) => {
+    return loadingTodoIds.includes(todoId);
+  };
```
781calmstate: just make it a value? why is it a function?

781calmstate (changes requested):
Remember as we discussed with you about mixing two approaches to work with promises. Choose one and stick to it, either `async await` or `.then()...`. I suggest `async await` because it's newer and modern way to deal with promises.

src/hooks/useError.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (!error) {
+      return;
+    }
+
+    const timer = setTimeout(handleRemoveError, 3000);
+
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [error, handleRemoveError]);
```
alexander-ignatow, line 23: Bug: when two non-empty errors arrive one after another in less than 3 sec, the second will be shown for less than 3 sec, because you don't clear the timeout after the first error

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,60 @@
...
+import cn from 'classnames';
+import { Status } from '../types/TodoStatusFilter';
+import { TODO_STATUS_FILTER_OPTIONS } from '../utils/filterOptions';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todos: Todo[];
+  selectedStatus: Status;
+  setSelectedStatus: (status: Status) => void;
+  handleDeleteCompleted: () => void;
```
alexander-ignatow, line 11:
```suggestion
  onDeleteCompleted: () => void;
```

alexander-ignatow (approved):
Please look through some minor issues and fix them when polishing the project for your portfolio

PR #2011, opened 2025-11-05
Part3

Vlshedevr:
I need help because I'm a little confused and can't solve the problem where when I change the Todo title, if it sees, for example, a network error, my focus is lost, and the written text is not saved (moreover, the input essentially disappears).
[DEMO LINK](https://Vlshedevr.github.io/react_todo-app-with-api/)


brespect (changes requested):
First of all, I have the ability to edit the title of the todo correctly in your expanded link, so if you have any questions or to clarify, you can go to the fe_chat, also don't forget to add the attribute data-cy="TodoTitleField", this is important for tests.

Anton-Kuchmasov (approved):
Great work you did!

PR #2015, opened 2025-11-09
solution

ghost:
https://denysherych.github.io/react_todo-app-with-api/


etojeDenys (changes requested):
1. after changing the status of multiple todos, a problem arises:
The loading indicator appears on all items and then disappears as soon as the first todo finishes loading, even though the other todos haven’t finished updating their status

etojeDenys (approved):
good job

PR #2017, opened 2025-11-10
add task solution

webqarbon:
https://webqarbon.github.io/react_todo-app-with-api/


src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,42 @@
...
+    if (!error) {
+      return;
+    }
+
+    const timerId = setTimeout(() => {
+      onClose();
+    }, 3000);
+
+    return () => clearTimeout(timerId);
+  }, [error, onClose]);
```
alexander-ignatow, line 20: When two consecutive non-empty errors arrive the timeout is not cleared, so the second error will be shown for less then 3 seconds

src/App.tsx
```diff
@@ -1,26 +1,272 @@
...
+      .catch(() => {
+        setError('Unable to load todos');
+      });
+  }, []);
+
+  useEffect(() => {
+    if (!isAdding && !tempTodo) {
+      const inputElement = document.querySelector<HTMLInputElement>(
+        '[data-cy="NewTodoField"]',
+      );
```
alexander-ignatow, line 44: Don't query the document directly - use refs instead

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  isLoading = false,
+  onDelete,
+  onToggle,
+  onUpdate,
+}) => {
+  const [isEditing, setIsEditing] = React.useState(false);
+  const [editTitle, setEditTitle] = React.useState(todo.title);
```
alexander-ignatow, line 22: Minor - but the style is inconcistent

```suggestion
  const [isEditing, setIsEditing] = useState(false);
  const [editTitle, setEditTitle] = useState(todo.title);
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,152 @@
...
+      setEditTitle(todo.title);
+      setIsEditing(false);
+    }
+  };
+
+  React.useEffect(() => {
+    if (isEditing) {
+      const input = document.querySelector<HTMLInputElement>(
+        `[data-cy="TodoTitleField"][data-id="${todo.id}"]`,
+      );
```
alexander-ignatow, line 87: ref should be used

alexander-ignatow (changes requested):
Good, but some issues are =needed to be fixed

PR #2020, opened 2025-11-12
implemented: toggling a todo status, renaming a todo and animations

Maevalar:
[DEMO LINK](https://maevalar.github.io/react_todo-app-with-api/)


nazkov (approved):

781calmstate (approved):
Overall good job!
But I'd recommend to fix the disappearing error message problem. For now, the message disappears first and only after the error box starting to fade-out which causes the error box to "jump" in height. My suggestion is to find a solution to make the error and the error box disappear simultaneously.

PR #2026, opened 2025-11-21
react_todo-app-with-api

DiegoBalkeKuhn:


pedro-ruas (changes requested):
Your code is missing the implementation for the components.
You made the API, but you haven't used it or made the app work

IanLuan (approved):
Good job! A tip: You almost passed all tests, the only missing thing is keeping the new todo input focused when page load or after adding a todo.

PR #2036, opened 2025-12-17
Solution

zaurmamedov:
[DEMO LINK](https://zaurmamedov.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,26 +1,289 @@
...
+      .updateTodo({ id, title: todoTitle, completed })
+      .then(updatedTodo => {
+        setTodos(currentTodos => {
+          return currentTodos.map(todo =>
+            todo.id === updatedTodo.id ? updatedTodo : todo,
+          );
+        });
+      })
+      .catch(error => {
+        setErrorMessage('Unable to update a todo');
```
Anton-Kuchmasov, line 159: Consider creating custom type / enum / [object as const ](https://stackoverflow.com/questions/66993264/what-does-the-as-const-mean-in-typescript-and-what-is-its-use-case)in separate file (e.g. ErrorType.ts) and re-use it to avoid hardcoded typo mistakes

src/App.tsx
```diff
@@ -1,26 +1,289 @@
...
+
+        <TodoList
+          todos={visibleTodos}
+          loadingTodoIds={loadingTodoIds}
+          onChange={onComplete}
+          onRemove={deleteTodo}
+          onUpdate={updateTodo}
+        />
+
+        {todos.length !== 0 && (
```
Anton-Kuchmasov, line 271:
```suggestion
        {!!todos.length && (
```

src/api/todos.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const deleteTodo = (id: number) => {
+  return client.delete(`/todos/${id}`);
+};
+
+export const updateTodo = ({
+  id,
+  title,
+  completed,
+}: Pick<Todo, 'id' | 'title' | 'completed'>) => {
```
Anton-Kuchmasov, line 26: Great solution, but it can be better with declaring a new type and re-use it

Anton-Kuchmasov (approved):
Great job you did!

Check out comments below - they might be helpful

PR #2058, opened 2026-01-17
add task solution

devsync25:


etojeDenys (approved):
good job

to improve:
 you should split the app into several components (for example, Header, Footer, TodoList and Error Notification)

PR #2060, opened 2026-01-21
Part 3 Solution 1

SerMamchur:
https://SerMamchur.github.io/react_todo-app-with-api/


etojeDenys (changes requested):
almost done

1. after pressing 'Clear Completed', the loader only shows up on one todo instead of on every todo that will be deleted. please fix it

etojeDenys (approved):
good job

PR #2065, opened 2026-01-23
react_todo-app-with-api

igortinyayev:
[DEMO LINK](https://igortinyayev.github.io/react_todo-app-with-api/)


etojeDenys (changes requested):
almost done

1. after pressing 'toggle all' button, the loader only appears on the last todo instead of on every todo that will change status

2. after pressing 'clear completed' button, the loader does not appear at all, but it should

brespect (changes requested):
Make sure you`ve correctly deployed your App, because comments from previous review were not fixed.

etojeDenys (approved):
lgtm

PR #2071, opened 2026-01-27
solution

eg598:


brespect (changes requested):
Good work, let`s check next:
1. Add DEMO Link to the description.
2. Check logic of load todos error

PR #2072, opened 2026-01-27
solution

Nabeyko:
- [DEMO LINK](https://Nabeyko.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -49,15 +49,17 @@ export const App: React.FC = () => {
   }, []);
 
   useEffect(() => {
+    let timer: ReturnType<typeof setTimeout>;
+
     if (isError) {
-      const timer = setTimeout(() => {
+      timer = setTimeout(() => {
         setIsError(null);
       }, 3000);
-
-      return () => clearTimeout(timer);
     }
 
-    return undefined;
+    return () => {
+      clearTimeout(timer);
+    };
   }, [isError]);
 
   useEffect(() => {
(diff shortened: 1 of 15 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,89 @@
...
+      href: '#/completed',
+      dataCy: 'FilterLinkCompleted',
+      buttonTitle: 'Completed',
+      status: FilterStatus.Completed,
+    },
+  ];
+
+  const notCompletedTodo = () => {
+    return allTodos.filter(todo => !todo.completed).length;
+  };
```
Denys-Kravchuk9988: I'd recommend using a constant instead of a function here.

src/App.tsx
```diff
@@ -1,26 +1,272 @@
...
+  useEffect(() => {
+    if (isError) {
+      const timer = setTimeout(() => {
+        setIsError(null);
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+
+    return undefined;
```
Denys-Kravchuk9988: `return undefined` is redundant, so it can be removed

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:

etojeDenys (changes requested):
you should trim the text for editing purposes
Image

if I add a space here, the todo will not be empty and instead of deleting it, we will get a todo with a space

Image

brespect (approved):
Lgtm 🚀

PR #2087, opened 2026-02-02
solution

PODDN:
[DEMO LINK](https://PODDN.github.io/react_todo-app-with-api/)


src/components/ErrorNotif/ErrorNotif.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
```
Anton-Kuchmasov: Consider use normal names and avoid contractions in components naming (`ErrorNotification.tsx`)

src/App.tsx
```diff
@@ -1,26 +1,306 @@
...
 
-const USER_ID = 0;
+export type TodosError =
+  | 'Unable to load todos'
+  | 'Title should not be empty'
+  | 'Unable to add a todo'
+  | 'Unable to delete a todo'
+  | 'Unable to update a todo'
+  | null;
+
```
Anton-Kuchmasov: Please move `TodosError` & `Filter` (it's better rename it e.g. `TodoStatus` or `TodoFilter` at least) in separate files (f.e. inside already existing `types` dir)

Anton-Kuchmasov (changes requested):
Well done!

Please make minor changes before get approved

etojeDenys (changes requested):
almost done

there should not be a footer if there are no todos in the list
Image

etojeDenys (approved):
good job

PR #2120, opened 2026-02-20
Develop

Wiolip:
[DEMO LINK](https://wiolip.github.io/react_todo-app-with-api/)


danon321 (changes requested):
Tez 404 z linku do demo

Zibi95 (changes requested):
the same here, let's firstly finish second part before going into this, so we can only focus on problems that this task introduces

Zibi95 (commented):
let's comeback to this after part 2 will be approved

Zibi95 (changes requested):

danon321 (approved):
Wygląda super. Dobra robota!

PR #2130, opened 2026-02-26
Solution

HiBlurryface:
https://HiBlurryface.github.io/react_todo-app-with-api/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,14 +14,13 @@ import { Footer } from './components/Footer/Footer';
 import { TodoContext } from './store/TodoContext';
 import { ErrorContext } from './store/ErrorContext';
 
-import { useError } from './hooks/useError';
 import { getTodos, USER_ID } from './api/todos';
 import { UserWarning } from './UserWarning';
+import { useContext } from 'react';
 
 export const App: React.FC = () => {
-  const { error, showError, closeError } = useError();
-  const [todos, setTodos] = useState<Todo[]>([]);
-  const [loadingIds, setLoadingIds] = useState<number[]>([]);
+  const { isError, errorMessage, showError, closeError } = useContext(ErrorContext);
+  const { todos, setTodos } = useContext(TodoContext)
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [filter, setFilter] = useState<Filter>('All');
 
@@ -64,18 +63,12 @@ export const App: React.FC = () => {
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
-        <TodoContext.Provider
-          value={{ todos, setTodos, loadingIds, setLoadingIds }}
-        >
-          <ErrorContext.Provider value={{ ...error, showError, closeError }}>
-            <Header setTempTodo={setTempTodo} />
-            <TodoList todos={filteredTodos} />
-            {tempTodo !== null && <TodoItem todo={tempTodo} />}
-            {todos.length > 0 && (
-              <Footer filter={filter} setFilter={setFilter} />
-            )}
-          </ErrorContext.Provider>
-        </TodoContext.Provider>
+        <Header setTempTodo={setTempTodo} />
+        <TodoList todos={filteredTodos} />
+        {tempTodo !== null && <TodoItem todo={tempTodo} />}
+        {todos.length > 0 && (
+          <Footer filter={filter} setFilter={setFilter} />
+        )}
       </div>
 
       <div
@@ -83,7 +76,7 @@ export const App: React.FC = () => {
         className={classNames(
           'notification is-danger is-light has-text-weight-normal',
           {
-            hidden: !error.isError,
+            hidden: !isError,
           },
         )}
       >
@@ -93,7 +86,7 @@ export const App: React.FC = () => {
           className="delete"
           onClick={closeError}
         />
-        {error.errorMessage}
+        {errorMessage}
       </div>
     </div>
   );
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,21 +1,22 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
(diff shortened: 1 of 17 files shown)


src/App.tsx
```diff
@@ -1,26 +1,138 @@
...
+      } catch (err) {
+        showError('Unable to load todos');
+      }
+    };
+
+    loadTodos();
+  }, []);
+
+  const filteredTodos = () => {
+    let copyTodos = [...todos];
```
zibi-dev: Array.filter() już zwraca nową tablicę. Spread [...todos] jest niepotrzebny.

src/App.tsx
```diff
@@ -1,26 +1,138 @@
...
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <TodoContext.Provider
+          value={{ todos, setTodos, loadingIds, setLoadingIds }}
+        >
+          <ErrorContext.Provider value={{ ...error, showError, closeError }}>
```
zibi-dev: Here ytou have some good tips for context

https://kentcdodds.com/blog/how-to-use-react-context-effectively

src/App.tsx
```diff
@@ -1,26 +1,138 @@
...
+        break;
+      case 'Completed':
+        copyTodos = copyTodos.filter(item => item.completed);
+        break;
+      default:
+        break;
+    }
+
+    return copyTodos;
+  };
```
zibi-dev: use useMemo for this

src/App.tsx
zibi-dev: Both contexts create new object on each render

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,170 @@
+import classNames from 'classnames';
+import { useContext } from 'react';
```
zibi-dev: Why are you impolrting in each line from the same module it can be squashed into one line  

`import { useRef, useEffect, useContext, useState } from 'react';`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,170 @@
...
+      className={classNames('todo', { completed: todo.completed })}
+    >
+      {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+      <label className="todo__status-label">
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+          onClick={todoCompleteButton}
```
zibi-dev: Here it should listen for onChange event. onClick z checked will create React warning about controlled component

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,170 @@
...
+      setIsEditing(false);
+
+      return;
+    }
+
+    setLoadingIds(prev => [...prev, todo.id]);
+    try {
+      if (trimmed.length === 0) {
+        inputRef.current?.focus();
+        await deleteTodoFromServer();
```
zibi-dev: In editTodo, you add todo.id to loadingIds and then (if the title is empty) call deleteTodoFromServer(), which also adds the same todo.id to loadingIds. So the same ID gets added twice. Then editTodo removes the ID in its finally, but deleteTodoFromServer also tries to remove it in its own finally- this creates a race condition where one finally can clear the loading state while the other operation is still in progress.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+        return;
+      }
+
+      await toggleTodo(todo.id, { title: trimmed });
```
zibi-dev: Better name fro this could be patchTodo

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+export const Footer: React.FC<Props> = ({ filter, setFilter }) => {
+  const { todos, setTodos, setLoadingIds } = useContext(TodoContext);
+  const { showError } = useContext(ErrorContext);
+
+  const filters: Filter[] = ['All', 'Active', 'Completed'];
+  const completed = todos.filter(todo => todo.completed);
+  const completedCount = completed.length;
+  const active = todos.filter(todo => !todo.completed);
+  const activeCount = active.length;
```
zibi-dev:
```suggestion
const activeCount = todos.filter(t => !t.completed).length;
const completedCount = todos.length - activeCount;
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    results.forEach((result, index) => {
+      if (result.status === 'fulfilled' && result.value === 1) {
+        const todoId = idsToDelete[index];
+
+        setTodos(prev => prev.filter(todo => todo.id !== todoId));
+      } else {
+        showError('Unable to delete a todo');
+      }
+    });
```
zibi-dev: Each setTodos call inside the loop is a separate state update. In React 18, batching usually covers updates in event handlers, but in async code (after an await) it isn’t guaranteed. It’s better to collect the results first and then call setTodos once:

```
const deletedIds = results
  .map((r, i) => (r.status === "fulfilled" ? idsToDelete[i] : null))
  .filter(Boolean);

setTodos((prev) => prev.filter((t) => !deletedIds.includes(t.id)));
```

Zibi95 (changes requested):

Zibi95 (changes requested):
Thanks for the update - most of my previous comments are addressed well.  
I still see two issues to fix before approval:

1. src/components/TodoItem/TodoItem.tsx  
In deleteTodoFromServer finally, you append todo.id to loadingIds instead of removing it.  
This can leave the loader active permanently for that item.

2. src/components/Footer/Footer.tsx  
activeCount and completedCount are both calculated via filter.  
Please simplify to one filter:
```
const activeCount = todos.filter(t => !t.completed).length;
const completedCount = todos.length - activeCount;
```
Please fix these and I’ll approve.

Zibi95 (approved):
Nice. Good job!

PR #2149, opened 2026-03-15
add solution

0-nira-0:
[DEMO LINK](https://0-nira-0.github.io/react_todo-app-with-api/)


etojeDenys (changes requested):
almost done

1. there is a problem when you change the title of a todo to just a space (you should use the trim())

Image

brespect (approved):
Lgtm 🚀

PR #2167, opened 2026-03-30
Develop

AnastasiiaHombalevska:
[DEMO LINK](https://AnastasiiaHombalevska.github.io/react_todo-app-with-api/)

### DIFF
--- a/cypress/integration/page.spec.js
+++ b/cypress/integration/page.spec.js
@@ -561,7 +561,7 @@ describe('', () => {
           page.flushJSTimers();
 
           todos.assertCount(7);
-          // todos.assertNotLoading(6);
+          todos.assertNotLoading(6);
           todos.assertNotCompleted(6);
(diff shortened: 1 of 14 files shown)


Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. There is no loading state on todos when I press on `Clear completed`
Image

2. It's better to remove the text that appears when I toggle some task
Image

3. By adding a new task the placeholder text doesn't have vertical paddings as already loaded tasks
Image

brespect (changes requested):
Good progress, but you need to pass all tests before requesting the review.

etojeDenys (changes requested):
1. after trying to change the status of several todos in a short period of time, the loader only appears on the last todo that was pressed

2. after pressing the 'toggle all' button, a loader should appear on the todos that will change status

Image

brespect (approved):
Well done!

PR #2175, opened 2026-04-14
add solution

dimon2202:
[DEMO LINK](https://dimon2202.github.io/react_todo-app-with-api/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,172 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useState, useEffect, useCallback } from 'react';
+import * as todoServise from './api/todos';
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { ErrorNotification } from './components/ErrorNotification';
+import { TodoList } from './components/TodoList';
+import { ErrorMessage } from './types/ErrorMessage';
 
-const USER_ID = 0;
+function getPrepearedTodos(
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,26 +1,172 @@
...
+      .catch(error => {
+        setErrorMessage(ErrorMessage.Unable_to_delete_a_todo);
+        throw error;
+      })
+      .finally(() => setLoadingTodoId(null));
+  };
+
+  const deleteCompletedTodos = () => {
+    complitedTodos.map(todo => deleteTodo(todo.id));
+  };
```
etojeDenys, line 123: it would be better to use Promise.all here

src/App.tsx
```diff
@@ -1,26 +1,172 @@
...
+      .finally(() => setLoadingTodoId(null));
+  }, []);
+
+  const toggleAll = () => {
+    if (activeTodos.length > 0) {
+      activeTodos.map(todo => updateTodo({ ...todo, completed: true }));
+    } else {
+      todos.map(todo => updateTodo({ ...todo, completed: false }));
+    }
+  };
```
etojeDenys, line 102: use Promise.all here as well

etojeDenys (changes requested):
1. after trying to change the status of several todos, the loader only shows up on the last one that was pressed

the same problem with deleting, toggling all and clearing completed

src/App.tsx
```diff
@@ -1,26 +1,174 @@
...
-    return <UserWarning />;
+  if (status) {
+    switch (status) {
+      case Status.Active:
+        preparedTodos = preparedTodos.filter(todo => !todo.completed);
+        break;
+      case Status.Completed:
+        preparedTodos = preparedTodos.filter(todo => todo.completed);
+        break;
+    }
```
Anton-Kuchmasov, line 31: better add `default` case to all `switch-case` statements

Anton-Kuchmasov (approved):
Well done!

PR #2179, opened 2026-04-21
add solution

RuslanV23:
https://ruslanv23.github.io/react_todo-app-with-api/


brespect (changes requested):
For now when I add something I get infinite loader:
Image

RuslanV23:
> For now when I add something I get infinite loader: Image



How so? ) 

Everything work fine for me. I can't reproduce the error. ((

RuslanV23:
I don't understand, everything works fine for me. Could you please check again?

brespect (approved):
Strange, for now its good, well done!

PR #2186, opened 2026-05-02
Develop

Bohdan259:
[DEMO LINK](https://Bohdan259.github.io/react_todo-app-with-api/)


etojeDenys (changes requested):
almost done
after trying to change the status of several todos, the loader only appears on the last one that was pressed. it should appear on every todo that will change state. the same problem with deleting

src/components/TodoAppHeader/TodoAppHeader.tsx
```diff
@@ -0,0 +1,111 @@
...
+            });
+          }
+        }
+      },
+      [title],
+    );
+
+    return (
+      <header className="todoapp__header">
+        {todos.length > 0 && (
```
Anton-Kuchmasov, line 79:
```suggestion
        {!!todos.length && (
```

Anton-Kuchmasov (approved):
Well done!


Before preparing to the Final Review tasks, please get rid of all inline comments in your code

PR #2204, opened 2026-05-17
implement solution

whomngmnt:
[DEMO LINK]  https://whomngmnt.github.io/react_todo-app-with-api/


dothearts (approved):
The same changes. 
Can add a single interface/type/enum for Filter (status). 
Also you can move error messages in constants/enum.

PR #2228, opened 2026-05-25
add task solution

rlysunets:
[DEMO LINK](https://rlysunets.github.io/react_todo-app-with-api/)


src/components/Footer.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <span className="todo-count" data-cy="TodosCounter">
+        {activeTodosCount} items left
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={`filter__link ${filter === 'all' ? 'selected' : ''}`}
+          data-cy="FilterLinkAll"
+          onClick={() => onFilterChange('all')}
```
dothearts, line 29: you can use enum in this case

src/types/Filter.ts
```diff
@@ -0,0 +1 @@
+export type Filter = 'all' | 'active' | 'completed';
```
dothearts, line 1: Improvements:  you can update this type using ENUM 

In this case you ref all types 'all', 'active', etc.  to one place and can easy change it in future. 
Don't need changes inside components.

dothearts (approved):
Left some comments for changes. 

Approved! 

Many thanks!

PR #2231, opened 2026-05-26
solution

joselangoni:
[DEMO LINK](https://joselangoni.github.io/react_todo-app-with-api/)


IanLuan (approved):
Great job! I'm accepting your task, but just one tip: next time, try to split your components. Currently, all the code is inside `App.tsx`, which makes it harder to maintain. It's a good practice to break it into smaller components.

PR #2232, opened 2026-05-26
add solution

MaxShaggy:
https://MaxShaggy.github.io/react_todo-app-with-api/


Palfinger2002 (approved):
Good job. But it's better for you to make a components for: Footer, TodoList, TodoItem, Header, etc and bring out of the logic. Because it's more readability and understand. Also you add files from master fork maybe you need to change on develop.

dothearts (approved):

PR #2235, opened 2026-05-29
add task solution

anastasiiadns:
[DEMO LINK](https://anastasiiadns.github.io/react_todo-app-with-api/)


etojeDenys (changes requested):
good job, but there is one thing to fix
1. right now, the loader only appears when a task is being added. after other manipulations, the loader does not show up, but it should

brespect (approved):
Well done!

PR #2240, opened 2026-06-02
Solution

lilp3ka:
https://lilp3ka.github.io/react_todo-app-with-api/


src/App.tsx
```diff
@@ -1,26 +1,469 @@
...
+  USER_ID,
+  getTodos,
+  addTodo,
+  deleteTodo,
+  updateTodo,
+} from './api/todos';
+
+import { Todo } from './types/Todo';
+
+type Filter = 'all' | 'active' | 'completed';
```
dothearts, line 18: You can made this as ENUM

src/App.tsx
```diff
@@ -1,26 +1,469 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [filter, setFilter] = useState<Filter>('all');
+  const [title, setTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingIds, setLoadingIds] = useState<number[]>([]);
+  const [isAdding, setIsAdding] = useState(false);
+  const [editingId, setEditingId] = useState<number | null>(null);
+  const [editedTitle, setEditedTitle] = useState('');
```
dothearts, line 29: You can perform state optimization in this case. 
You really need all this property as separate state? 
Can you unit some of them using object for example? 
Try to think and upgrade this solution

src/App.tsx
```diff
@@ -1,26 +1,469 @@
...
+    inputRef.current?.focus();
+  }, []);
+
+  useEffect(() => {
+    setError('');
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setError('Unable to load todos');
```
dothearts, line 49: You can move all errors in ENUM / constants

src/App.tsx
```diff
@@ -1,26 +1,469 @@
...
+    return () => clearTimeout(timer);
+  }, [error]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const visibleTodos = todos.filter(todo => {
+    switch (filter) {
+      case 'active':
```
dothearts, line 71: You can use Enum in this case for better redability and changes in future

src/App.tsx
```diff
@@ -1,26 +1,469 @@
...
+      const createdTodo = await addTodo({
+        userId: USER_ID,
+        title: trimmedTitle,
+        completed: false,
+      });
+
+      setTodos(current => [...current, createdTodo]);
+      setTitle('');
+    } catch {
+      setError('Unable to add a todo');
```
dothearts, line 119: Create ENUM / constatns for all errors

src/App.tsx
```diff
@@ -1,26 +1,469 @@
...
+    }, 3000);
+
+    return () => clearTimeout(timer);
+  }, [error]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const visibleTodos = todos.filter(todo => {
```
dothearts, line 69: You can made it as pure function and move in separate file

dothearts (changes requested):
Left comments to improve your solution. 
Also check you pipelines status (currently it **x** - not passed) 

In general, you need to make some changes for better code readability, scalability, and future changes. Try to make it better using my comments! 

Many thanks!

PR #2245, opened 2026-06-04
Solution

YuriiHoliuk:
Test submission for in-API AI review finalization (staging).


src/App.tsx
```diff
@@ -1,26 +1,496 @@
...
+        current.filter(todo => !successfulIds.includes(todo.id)),
+      );
+
+      setProcessingIds(current =>
+        current.filter(pid => !completedIds.includes(pid)),
+      );
+    });
+  };
+
   return (
```
mate-academy-dev, line 269: This violates task checklist item `App.js code should be split into several components` and `Code should be split into small, reusable components if it possible (Filter, TodoList, Todo, NewTodo)` because all rendering and logic live in `App` instead of dedicated components.

src/App.tsx
```diff
@@ -1,26 +1,496 @@
...
+import { USER_ID } from '../src/api/todos';
+import * as postService from '../src/api/todos';
+import classNames from 'classnames';
 
-const USER_ID = 0;
+enum FilterStatus {
+  All = '',
+  Active = 'active',
+  Completed = 'completed',
+}
```
mate-academy-dev, line 15: Checklist item `Do not rely on the unknown string, make constants for this. const FILTERS = { all: 'all', completed: 'completed', active: 'active' };` is not followed. You use `enum FilterStatus` instead of the required `FILTERS` constant, which may break tests expecting that shape.

src/App.tsx
```diff
@@ -1,26 +1,496 @@
...
+        inputRef.current?.focus();
+      })
+      .catch(() => {
+        setErrorMessage(ErrorMessage.Add);
+      })
+      .finally(() => {
+        setTempTodo(null);
+        setIsAdding(false);
+      });
+  };
```
mate-academy-dev, line 125: Checklist item `ID for new todos should be unique, you can use an internal ID counter for this, and increment it.` is not implemented; you always use the server-created todo and never maintain an internal incremental ID counter for new todos.

src/App.tsx
```diff
@@ -1,26 +1,496 @@
...
+                      data-cy="TodoLoader"
+                      className={classNames('modal', 'overlay', {
+                        'is-active': isProcessing,
+                      })}
+                    >
+                      <div className="modal-background has-background-white-ter" />
+                      <div className="loader" />
+                    </div>
+                  </div>
+                );
```
mate-academy-dev, line 387: Task checklist item `Callbacks that work with the main state should take prepared data instead of the whole child's state` is not really addressed because all logic is inside `App`. Once you split into `Todo`, `NewTodo`, `Filter`, ensure props callbacks receive primitive/prepared values (like `id`, `newTitle`, `completed`) instead of passing whole internal child state objects.

src/styles/filter.css
```diff
@@ -0,0 +1,17 @@
...
+  padding: 3px 7px;
+  color: inherit;
+  text-decoration: none;
+  border: 1px solid transparent;
+  border-radius: 3px;
+}
+.filter__link:hover {
+  border-color: rgba(175, 47, 47, 0.1);
+}
+.filter__link.selected {
```
mate-academy-dev, line 15: According to **Task checklist #10**: `"Do not rely on the unknown string, make constants for this."` the required constant shape is `const FILTERS = { all: 'all', completed: 'completed', active: 'active' }`. Here you use an enum `FilterStatus` with different keys (`All`, `Active`, `Completed`). While it avoids raw strings, it doesn’t match the exact constant structure required in the checklist; consider introducing the `FILTERS` object as specified and basing your enum or logic on it.

src/styles/index.css
```diff
@@ -0,0 +1,286 @@
...
+body {
+  min-width: 230px;
+  max-width: 550px;
+  margin: 0 auto;
+}
+
+.notification {
+  transition-property: opacity, min-height;
+  transition-duration: 1s;
+  min-height: 36px;
```
mate-academy-dev, line 15: Using `enum FilterStatus` instead of the specified constants violates **Task checklist item #10**: `"Do not rely on the unknown string, make constants for this. const FILTERS = { all: 'all', completed: 'completed', active: 'active' };"`. Replace this enum with the required `FILTERS` object and use its values where you compare or set the filter.

src/untils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+  return new Promise(resolve => {
+    setTimeout(resolve, delay);
+  });
+}
+
+// To have autocompletion and avoid mistypes
+type RequestMethod = 'GET' | 'POST' | 'PATCH' | 'DELETE';
+
+function request<T>(
+  url: string,
```
mate-academy-dev, line 15: Task checklist item #10 requires a specific constant: `const FILTERS = { all: 'all', completed: 'completed', active: 'active' };`. Here you define `enum FilterStatus` instead, which doesn’t match the required `FILTERS` object structure. Replace or supplement this enum with the exact `FILTERS` constant and use its values in your filter logic.

mate-academy-dev (changes requested):
### Review Summary

I am requesting changes because several explicit task checklist requirements are not yet met, even though the core functionality works well. The most important blocker is that all UI and logic are still in `App.tsx` instead of being split into reusable components like `NewTodo`, `TodoList`, `Todo`, and `Filter` as required [CHECKLIST ITEM #1, #3]; this also means callbacks with prepared data (e.g. `onToggle(id, completed)`, `onSave(id, newTitle)`) are not implemented as specified [CHECKLIST ITEM #2]. The task explicitly requires a `FILTERS` constant with shape `const FILTERS = { all: 'all', completed: 'completed', active: 'active' }`, but you use an enum `FilterStatus` instead, which can break tests expecting the constant [CHECKLIST ITEM #10]. Another blocking issue is the missing internal incremental ID counter for new todos; you currently rely only on server IDs, which contradicts the requirement to maintain your own unique ID generation [CHECKLIST ITEM #4]. Finally, when `todos.length === 0` but a `tempTodo` exists during the first add, the main section/footer still render, while the checklist says that only the NewTodo form should be visible when the todos array is empty [CHECKLIST ITEM #11], and PropTypes are not defined for components that will receive object/array props once you split `App` [Basic React CHECKLIST].

To move forward, refactor `App` into smaller components (`NewTodo`, `TodoList`, `Todo`, `Filter`), adjust their props so callbacks take primitive/prepared values rather than whole todo objects, introduce the exact `FILTERS` constant and use it in your filter logic, add an internal ID counter for new todos, and ensure the footer/main section are hidden whenever there are no persisted todos. Once these structural and checklist-aligned changes are in place, your already solid business logic will fully satisfy the task requirements.

***

✨ Please rate this review in the Platform's notification center[^1]. I will be grateful for your feedback! ✨

[^1]: [Rate AI review example](https://snipboard.io/gD6XKF.jpg)

> [!IMPORTANT]
> Fix the issues, commit the changes, and then re-request my review using the **`Re-request review` button**[^2] in the Reviewers tab.

[^2]: [Re-request AI review example](https://snipboard.io/nOXxHN.jpg)

> [!TIP]
> If you have any comments about this review, please leave a message in the [Add a comment](#new_comment_field) section below and **mention me with the `@mateacademy-ai-mentor` tag**[^3].
> I will read your message during my next review of your pull request. Please do not reply directly to my comments, as that feature is not supported yet.

[^3]: [Add AI review comments example](https://snipboard.io/hVlOey.jpg)

> [!CAUTION]
> Do not close the pull request until the review is complete. Closing it early can lead to unexpected behavior and disrupt the review process.

YuriiHoliuk:
Closing — staging AI-review finalization verification.

PR #2248, opened 2026-06-06
solution

staner1:
[DEMO LINK](https://staner1.github.io/react_todo-app-with-api/)


etojeDenys (changes requested):
right now, users can change the 'todo' to just a space, which leads to this problem(you should use trim() here as well)
image

vadiimvooo (approved):
Well done

PR #2261, opened 2026-06-26
add task solution

bodyafell:
[DEMO LINK](https://bodyafell.github.io/react_todo-app-with-api/)

I just did the last task, can you tell me how the previous ones were completed????


src/App.tsx
```diff
@@ -1,26 +1,382 @@
...
+  };
+
+  // Load todos on component mount
+  useEffect(() => {
+    getTodos()
+      .then(data => {
+        setTodos(data);
+      })
+      .catch(() => {
+        showError('Unable to load todos');
```
etojeDenys, line 54: create an enum value for errors

etojeDenys (changes requested):
1. after trying to delete several todos in a short amount of time, the loader only appears on the last todo pressed. (the same with changing status)

2. loader does not appear after pressing the 'Toggle All' and 'Clear Completed' buttons

3. split the app into several components (for example: header, main, footer errorNotification)

etojeDenys (approved):
great job

PR #2265, opened 2026-07-03
add task solution

impo10nt:
DEMO LINK https://impo10nt.github.io/react_todo-app-with-api/


src/components/Todo/TodoItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  loading,
+  onDelete,
+  onChange,
+}) => {
+  const [edit, setEdit] = useState(false);
+
+  function handleChangeCompleted() {
+    onChange(todo.id, { completed: !todo.completed }).then(() => {});
```
dothearts, line 22: Why we need then in this case? 
And also, why we don't catch error?

src/components/Todo/TodoItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+      onDelete(todo.id);
+
+      return;
+    }
+
+    onChange(todo.id, { title })
+      .then(() => {
+        setEdit(false);
+      })
+      .catch(() => {});
```
dothearts, line 36: We catch an error, but the input is still in edit mode

src/hooks/useTodo.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    setLoadingTodoIds(prev => prev.filter(id => !ids.includes(id)));
+  }
+
+  const getTodoList = useCallback(async () => {
+    return getTodos()
+      .then(response => setTodoList(response))
+      .catch(() => {
+        throw new Error('Unable to load todos');
```
dothearts, line 66: You can move all errors in ENUM or constants

src/hooks/useTodo.ts
```diff
@@ -0,0 +1,188 @@
...
+  const addTodo = useCallback(async (title: Todo['title']) => {
+    return createTodo(title)
+      .then(response => setTodoList(prevState => [...prevState, response]))
+      .catch(() => {
+        throw new Error('Unable to add a todo');
+      });
+  }, []);
+
+  const deleteTodo = useCallback(async (id: Todo['id']) => {
+    loadingAddTodos([id]);
```
dothearts, line 79: Need fix naming, by my view: load todo which we want delete?

src/hooks/useTodo.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    return apiDeleteTodo(id)
+      .then(() =>
+        setTodoList(prevState => prevState.filter(todo => todo.id !== id)),
+      )
+      .catch(() => {
+        throw new Error('Unable to delete a todo');
+      })
+      .finally(() => loadingRemoveTodos([id]));
```
dothearts, line 88: Also don't understand clearly what do this function by name

src/hooks/useTodo.ts
```diff
@@ -0,0 +1,188 @@
...
+          prevState.map(todo => (todo.id === id ? response : todo)),
+        ),
+      )
+      .catch(() => {
+        throw new Error('Unable to update a todo');
+      })
+      .finally(() => loadingRemoveTodos([id]));
+  }, []);
+
+  const updateTodoCompletedField = useCallback(
```
dothearts, line 131: _updateAllCompletedTodos_? 
Or something like that, it's update several todos as i see not one todo

dothearts (changes requested):
In general all work fine. 
Please, try use better naming, add right catch strategy and remove code duplication. 

I left some comments, many thanks for fixes!

PR #2271, opened 2026-07-06
 task completed

droopy-bit:
[DEMO LINK](https://droopy-bit.github.io/react_todo-app-with-api/)


src/App.tsx
```diff
@@ -1,26 +1,402 @@
...
+    const allCompleted = todos.every(todo => todo.completed);
+    const newStatus = !allCompleted;
+    const todosToUpdate = todos.filter(todo => todo.completed !== newStatus);
+
+    todosToUpdate.forEach(todo => handleToggleTodo(todo));
+  };
+
+  const handleClearCompleted = () => {
+    completedTodos.forEach(todo => handleDelete(todo.id));
+  };
```
etojeDenys: you should use Promise.all() here

src/App.tsx
```diff
@@ -1,26 +1,402 @@
...
+        setTodos(prev => prev.map(t => (t.id === updated.id ? updated : t)));
+      })
+      .catch(() => {
+        setErrorMessage('Unable to update a todo');
+        setTimeout(() => setErrorMessage(''), 3000);
+      })
+      .finally(() => {
+        setLoadingIds(current => current.filter(id => id !== todo.id));
+      });
+  };
```
etojeDenys, line 135: use Promise.all() here as well

etojeDenys (changes requested):
it's the same as in the previous task. it looks good, but fix the problems from the first task(and new comments)

brespect (approved):
Well done!

PR #2302, opened 2026-08-08
React-todo-withFullApi 1.0

FRANCISCOPELEGRINI:


raulriato (changes requested):
You should not have all your code inside the `App.tsx` for this project.

The main feature in React is `componentization`. Make use of it.

Also, on your `DEMO LINK`, when the user clicks on `Completed` filter, if there are no completed tasks, it breaks the code. You should fix it.

DanielCamposDevX (approved):

PR #2308, opened 2026-08-15
react todo app with api

vovagorbovuy:


src/App.tsx
```diff
@@ -1,26 +1,247 @@
...
+        }, 0);
+      });
+  }
+
+  function handleClearCompleted() {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    if (completedTodos.length === 0) {
+      return;
+    }
```
etojeDenys, line 122: you should use promise.all() here

src/App.tsx
```diff
@@ -1,26 +1,247 @@
...
+
+    const isAllCompleted = todos.every(todo => todo.completed);
+    const targetStatus = !isAllCompleted;
+    const todoUpdate = todos.filter(todo => todo.completed !== targetStatus);
+
+    todoUpdate.forEach(todo => {
+      handleUpdateTodo({ ...todo, completed: targetStatus });
+    });
+  }
+
```
etojeDenys, line 185: use promise.all here as well

etojeDenys (changes requested):
almost done

roman-mirzoian (approved):
Well done 👍

